PR #599, opened 2026-02-27
fix(platform): improve chat streaming stability and UX

Israeltheminer:
## Summary
- Handle `aborted` stream status as terminal in `useChatLoadingState` and `useMessageProcessing` to prevent stuck loading indicators after stream cancellation
- Move `abortController` initialization earlier in `generate_response.ts` to ensure availability across all error paths
- Add early return for cancelled generation (`finishReason === 'cancelled'`) in agent action to skip validation
- Improve chat input with proper send `Button` component, `Tooltip` for attach, and `aria-label` accessibility
- Fix sidebar rename to fall back to "Untitled" on empty input instead of showing error toast
- Improve sidebar chat item click target area and add `aria-label`

## Test plan
- [ ] Send a message and verify streaming completes with loading indicator clearing correctly
- [ ] Verify chat input send button is visible and accessible (keyboard navigation, screen reader)
- [ ] Rename a chat to empty string — should save as "Untitled" instead of showing error
- [ ] Verify sidebar chat items are clickable across the full row area
- [ ] Verify attach button tooltip appears on hover

## Release Notes

* **New Features**
  * Added send button with arrow icon to chat input
  * Added agent selector to chat header
  * New tooltip for attachment action

* **Improvements**
  * Chat history rename now uses double-click to edit (single-click navigates)
  * Improved rename validation with auto-default for empty names
  * Send button intelligently requires text or attachments before enabling
  * Better handling of cancelled message states

### DIFF
--- a/services/platform/app/features/chat/hooks/__tests__/use-chat-loading-state.test.ts
+++ b/services/platform/app/features/chat/hooks/__tests__/use-chat-loading-state.test.ts
@@ -222,6 +222,61 @@ describe('useChatLoadingState', () => {
       expect(result.current.isLoading).toBe(true);
     });
 
+    it('returns false when last assistant message is aborted', () => {
+      const { result } = renderHook(() =>
+        useChatLoadingState({
+          isPending: false,
+          setIsPending,
+          uiMessages: [
+            createUIMessage({
+              id: 'msg-1',
+              order: 0,
+              role: 'assistant',
+              // oxlint-disable-next-line typescript/no-unsafe-type-assertion -- runtime value exists but SDK types lack it
+              status: 'aborted' as UIMessage['status'],
+            }),
+          ],
+          threadId: THREAD_A,
+          pendingThreadId: null,
+        }),
+      );
+
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    it('returns false when aborted mid-tool-call (aborted overrides unfinished tool turn)', () => {
+      const { result } = renderHook(() =>
+        useChatLoadingState({
+          isPending: false,
+          setIsPending,
+          uiMessages: [
+            createUIMessage({
+              id: 'msg-1',
+              order: 0,
+              role: 'assistant',
+              // oxlint-disable-next-line typescript/no-unsafe-type-assertion -- runtime value exists but SDK types lack it
+              status: 'aborted' as UIMessage['status'],
+              text: 'Let me look that up.',
+              parts: [
+                { type: 'text', text: 'Let me look that up.' },
+                { type: 'step-start' },
+                {
+                  type: 'tool-rag_search',
+                  toolCallId: 'call-1',
+                  input: { query: 'test' },
+                  state: 'input-available',
+                },
+              ],
+            }),
+          ],
+          threadId: THREAD_A,
+          pendingThreadId: null,
+        }),
+      );
+
+      expect(result.current.isLoading).toBe(false);
+    });
+
     it('returns true when last message is user (waiting for AI response)', () => {
       const { result } = renderHook(() =>
         useChatLoadingState({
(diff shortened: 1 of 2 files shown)


larryro:
## Post-merge review: streaming stability changes

Good intent — fixing stuck loading spinners on abort is important. The UI improvements (send button, tooltip, sidebar hit target) are solid. However, **the core abort-handling logic is built on incorrect assumptions about the Convex Agent SDK**, meaning the key streaming fixes are dead code and the original bug remains.

---

### Core issue: `'aborted'` never reaches UIMessage

The Convex Agent SDK's `statusFromStreamStatus()` in `deltas.ts` maps stream states to UIMessage statuses:

```typescript
case "streaming": return "streaming";
case "finished":  return "success";
case "aborted":   return "failed";   // ← abort becomes "failed"
```

`UIMessage.status` is typed as `"streaming" | "pending" | "success" | "failed"` — there is **no `'aborted'` value**. The SDK stores `{ kind: "aborted", reason: "..." }` in the stream DB document, but this gets mapped to `"failed"` by the time it reaches the frontend.

This means three changes in this PR are dead code:

| File | Line | Code | Why it's dead |
|---|---|---|---|
| `use-chat-loading-state.ts` | 64 | `if (status === 'aborted') return false` | UIMessage status is never `'aborted'`, always `'failed'` |
| `use-message-processing.ts` | 207 | `m.status === 'aborted'` | Same — `'failed'` on line 206 already covers it |
| `internal_actions.ts` | 277 | `if (result.finishReason === 'cancelled')` | `'cancelled'` is not a valid `FinishReason` in AI SDK or Convex Agent (`"stop" \| "length" \| "content-filter" \| "tool-calls" \| "error" \| "other" \| "unknown"`). When `abortController.abort()` fires, the AI SDK **throws** an `AbortError` — it doesn't return a result. |

The `const status: string | undefined` type widening on line 61 masks this — it bypasses TypeScript's type checking to allow comparing against a value that isn't in the union. The test's `status: 'aborted' as UIMessage['status']` type cast is also a signal: the type system is correctly saying this value doesn't exist.

### The real bug (unfixed)

The actual bug is: when `status === 'failed'` AND `isUnfinishedToolTurn()` returns `true`, the loading state gets **stuck forever**.

The old logic:
```typescript
return !(
  lastMessage.role === 'assistant' &&
  !isUnfinishedToolTurn(lastMessage) &&       // false when tool turn incomplete
  (lastMessage.status === 'success' || lastMessage.status === 'failed')
);
// → !(true && false && true) → !(false) → true → loading forever
```

A failed/aborted generation mid-tool-call has unfinished tool parts, so `isUnfinishedToolTurn` returns `true`, and the whole condition never resolves. But a failed message will **never** complete its tool turn — no more steps will come.

### Recommended fix

Make `'failed'` unconditionally terminal, bypassing the tool-turn check:

```typescript
if (lastMessage.role !== 'assistant') return true;
if (lastMessage.status === 'failed') return false;  // always terminal, even mid-tool-call
return !(!isUnfinishedToolTurn(lastMessage) && lastMessage.status === 'success');
```

Then remove the dead `'aborted'` checks in `use-message-processing.ts` and the `'cancelled'` check in `internal_actions.ts`. Update tests to cover `status: 'failed'` + unfinished tool parts (the actual bug scenario) instead of the fictional `'aborted'` status.

### What's correct in this PR

- **`generate_response.ts`**: Moving `abortController` outside `try` is a real, necessary fix — ensures it's accessible in error/cleanup paths
- **`chat-input.tsx`**: Send button with proper disable logic and a11y is good
- **`chat-history-sidebar.tsx`**: Full-surface click target with `absolute inset-0` is a nice UX improvement
- **`messages/en.json`**: Translation addition is correct

### Minor UI notes

- Sidebar input `min-h-[1.5rem]` (24px) vs display span `min-h-[20px]` (20px) — slight visual jump when entering/exiting edit mode
- Rename empty fallback to "Untitled" is fine UX-wise, but `history.toast.titleEmpty` may now be dead i18n

PR #603, opened 2026-02-28
fix(platform): make failed status unconditionally terminal and remove dead code

Israeltheminer:
## Summary

Addresses review feedback from @larryro on #599. The Convex Agent SDK maps stream abort to `failed` (not `aborted`) at the UIMessage level via `statusFromStreamStatus()`, and `finishReason` never contains `cancelled` (valid values: stop, length, content-filter, tool-calls, error, other, unknown).

- **Fix stuck loading on mid-tool-call failure**: `failed` is now unconditionally terminal — when generation fails during a tool turn, `isLoading` correctly resolves to `false` instead of staying stuck forever
- **Remove dead `aborted` status checks**: SDK never surfaces `aborted` as a UIMessage status (maps to `failed`), so these checks in `use-chat-loading-state.ts` and `use-message-processing.ts` were unreachable
- **Remove dead `finishReason === 'cancelled'` check**: Not a valid `vFinishReason` value in the SDK, so this early-return in `internal_actions.ts` was unreachable
- **Remove dead `history.toast.titleEmpty` i18n key**: No longer referenced anywhere after sidebar rename UX changes
- **Fix `min-h` mismatch in sidebar**: Edit input used `min-h-[1.5rem]` (24px) while display span used `min-h-[20px]` — aligned both to `1.5rem`

## Test plan

- [x] Updated `use-chat-loading-state.test.ts` — replaced tool-turn-aware tests with `failed` unconditionally terminal tests
- [x] Added test: `failed` mid-tool-call clears `isPending`
- [x] Lint passes (0 warnings, 0 errors)
- [x] Format passes
- [ ] CI checks pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed loading state indicator to properly clear when tool calls fail, ensuring accurate feedback during chat operations.
  * Improved handling of failed operations during message processing.

* **Style**
  * Refined chat title area vertical spacing for better UI consistency.

### DIFF
--- a/services/crawler/pyproject.toml
+++ b/services/crawler/pyproject.toml
@@ -3,7 +3,7 @@ name = "tale-crawler"
 version = "0.1.0"
 requires-python = ">=3.11"
 dependencies = [
-    "fastapi==0.133.1",
+    "fastapi==0.134.0",
     "uvicorn[standard]==0.41.0",
     "pydantic==2.12.5",
     "pydantic-settings==2.13.1",
--- a/services/crawler/uv.lock
+++ b/services/crawler/uv.lock
@@ -751,7 +751,7 @@ wheels = [
 
 [[package]]
 name = "fastapi"
-version = "0.133.1"
+version = "0.134.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "annotated-doc" },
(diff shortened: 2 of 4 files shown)


larryro (approved):
lgtm

PR #958, opened 2026-04-03
feat: optimize Docker containers — 60% total image reduction (~10 GB → ~4 GB) + container e2e tests

saifaliunity:
## Summary

Optimizes all Docker container images using multi-stage builds with `FROM scratch` squash, reducing total image size by **60% (~10.1 GB → ~4.0 GB)** while maintaining full runtime functionality. Adds comprehensive container e2e test suite.

## Image size comparison

| Service | Before | After | Reduction |
|---------|--------|-------|-----------|
| DB (ParadeDB) | 2.44 GB | 718 MB | **71%** |
| Platform | 2.98 GB | 1.75 GB | **41%** |
| Crawler | 2.05 GB | 1.31 GB | **36%** |
| RAG | 539 MB | 357 MB | **34%** |
| Proxy | 87 MB | 62 MB | **29%** |
| **Total** | **~10.1 GB** | **~4.0 GB** | **~60%** |

## Key techniques

- **`FROM scratch` squash** as the final build stage — flattens all layers so file deletions actually reclaim disk space
- **Targeted cleanup** — strip debug symbols (`.dbg`), LLVM/Clang libraries, unused PostGIS extensions, test dirs, `__pycache__`
- **Multi-stage builds** — separate builder (compile deps) from runtime (only runtime libs)
- **Architecture-aware cleanup** — uses `dpkg --print-architecture` for ARM64 compatibility
- **JIT disabled** — `jit = off` in postgresql.conf after removing LLVM libraries
- **`STOPSIGNAL SIGINT`** for PostgreSQL fast shutdown (prevents SIGKILL on `docker stop`)
- **Chrome → headless_shell** — replaces full Chromium binary (257 MB) with a symlink to headless_shell, verified locally with Playwright + crawl4ai

## Key design decisions

- Platform entrypoint runs as root initially, then drops to `app` user via `gosu` (needed for Docker named volume ownership)
- `FROM scratch` drops all Docker metadata (ENV, USER, STOPSIGNAL) — all re-declared in the squash stage
- Volume mountpoints pre-created in runtime stage before squash
- Compatible with the JSON provider config from [Clean up legacy env var config and add SOPS encryption #961](https://github.com/tale-project/tale/pull/961)
- Kept `.dist-info` directories to preserve `importlib.metadata` compatibility
- Full Chrome binary replaced with symlink to headless_shell — Playwright follows the symlink transparently, no application code changes needed

## Container test suite

| Test | Script | Command |
|------|--------|---------|
| Smoke test | `tests/container-smoke-test.sh` | `bun run docker:test` |
| Image validation | `tests/container-image-test.sh` | `bun run docker:test:image` |
| Vulnerability scan | `tests/container-vulnerability-scan.sh` | `bun run docker:test:vulnerability` |

## CI/CD improvements

- **Trivy action pinned** to `@v0.35.0` (was `@master` — supply chain risk)
- **Release cache scopes** include architecture suffix (`-amd64`/`-arm64`) to prevent cross-arch cache collisions
- **Vulnerability scan blocking mode** now works — `VULNERABILITY_EXIT_CODE=1` actually fails the script
- **Trivy scan error capture** — `set +e`/`set -e` wrapper so scan failures are captured instead of aborting silently

## Documentation updated

- `docs/container-architecture.md` — full architecture reference
- `docs/contributing-docker.md` — local development guide
- `docs/production-deployment.md` — production deployment with size budgets
- `docs/quickstart.md` — test commands
- `docs/troubleshooting.md` — common issues

## Review feedback addressed

All review comments from @yannickmonney and @larryro have been addressed:

### @yannickmonney
- ✅ Vendor label → "Tale" (all Dockerfiles)
- ✅ No "vuln" abbreviations — renamed to "vulnerability" everywhere
- ✅ `github.server_url` variable in CI workflows
- ✅ `delivery_report.html` excluded via `.gitignore`

### @larryro
- ✅ **Full Chrome replaced with headless_shell** — symlink from `chromium-*/chrome` to `chromium_headless_shell-*/chrome-headless-shell` saves ~257 MB. Playwright follows the symlink transparently. Verified locally with both Playwright and crawl4ai.
- ✅ Architecture-aware LLVM cleanup paths for ARM64 (crawler, platform, db)
- ✅ `STOPSIGNAL SIGINT` added to db Dockerfile (PostgreSQL fast shutdown)
- ✅ `jit = off` in postgresql.conf (comment updated to reflect this)
- ✅ `TALE_PLATFORM_SHARED_CONFIG_DIR` ENV in RAG squash stage
- ✅ Removed redundant hardcoded-amd64 SOPS download in RAG runtime
- ✅ `distutils-precedence.pth` cleanup in both crawler and rag
- ✅ Kept `.dist-info` for `importlib.metadata` compatibility
- ✅ `set -euo pipefail` in all test scripts
- ✅ Trivy action pinned to `@v0.35.0` (was `@master`)
- ✅ Release cache scope includes arch suffix (prevents cross-arch cache collision)
- ✅ Vulnerability scan: fixed dead code (`scan_exit` unreachable under `set -e`)
- ✅ Vulnerability scan: `VULNERABILITY_EXIT_CODE` wired into exit logic (blocking mode works)

### Smoke test fix
- ✅ Fixed Docker race condition where crawler + rag containers both try to `mkdir providers` inside the shared `platform-data` volume concurrently
- ✅ Moved test fixtures to `tests/fixtures/platform-config/` and mount the whole directory instead of sub-mounting into a named volume


saifaliunity:
Hey @yannickmonney  , thanks for the feedback! We've investigated distroless-python and distroless-bun thoroughly. Here's what we found:

What we can do:
- RAG: Switch runtime to `gcr.io/distroless/python3-debian12:nonroot` (~109 MB saved, plus better security — no shell, no package manager)
- Crawler: Switch runtime to distroless-python (~100-200 MB saved), though Playwright's Chromium headless_shell (~1.4 GB) has to stay regardless
- Platform: Deeper node_modules pruning (~144 MB saved)

Why <5 GB total isn't achievable:
- Platform is built on `ghcr.io/get-convex/convex-backend` (2.98 GB) — this is the Convex backend binary + runtime, we can't swap it for distroless-bun since the Convex backend isn't a Bun app, it's a pre-compiled Rust binary
- DB is `paradedb/paradedb` (2.44 GB) — controlled upstream, no changes possible
- Platform + DB alone = 5.42 GB, already over the 5 GB target

Realistic outcome with distroless migration:
- Current total: 8.1 GB
- After distroless (RAG + Crawler) + deeper pruning: ~7.5-7.6 GB
- Savings: ~450-500 MB additional

Services we control (Crawler + RAG + Proxy):
- Current: 2.68 GB → After distroless: ~2.37 GB (well under 5 GB individually)

We're happy to proceed with the distroless migration for RAG and Crawler — the security benefits alone are worth it (no shell = smaller attack surface, no package manager in production). But wanted to be transparent that the <5 GB total target is constrained by the Convex and ParadeDB upstream base images.

Let us know how you'd like to proceed

saifaliunity:
Hey — updated PR with significantly improved results. Here's a quick summary:

**Image sizes: ~10.1 GB → ~6.1 GB total (40% reduction)**

| Service | Before | After | Saved |
|---------|--------|-------|-------|
| DB | 2.44 GB | 1.06 GB | 57% |
| Platform | 2.98 GB | 2.58 GB | 13% |
| Crawler | 2.05 GB | 1.85 GB | 10% |
| RAG | 539 MB | 515 MB | 4% |

The main technique is a `FROM scratch` squash as the final build stage — this flattens all layers so file deletions actually reclaim space. The DB was the biggest find — ParadeDB ships with ~1 GB of debug symbols and LLVM that have no use in production.

Key things to be aware of:
- Platform entrypoint runs as root initially, then drops to `app` user via `gosu` (needed for Docker named volume ownership)
- `FROM scratch` drops all ENV vars from upstream stages — they're re-declared in the squash stage
- Compatible with the new JSON provider config from #961
- Volume mountpoints pre-created in runtime stage before squash

All 22 CI checks are green ✅ — builds, smoke test, image validation, vulnerability scans all pass. All 5 services verified healthy locally. Documentation updated across 5 files (container-architecture, production-deployment, operations, contributing-docker, troubleshooting).

Happy to walk through any of these changes.

tests/container-vuln-scan.sh
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env bash
```
yannickmonney: rename do not use vuln abbr, write it out.

tests/container-vuln-scan.sh
```diff
@@ -0,0 +1,167 @@
...
+    trivy convert --format sarif --output "${OUTPUT_DIR}/${svc}-report.sarif" "${vuln_json_file}" 2>/dev/null || true
+
+    # Count vulnerabilities by severity
+    critical=$(grep -o '"Severity": "CRITICAL"' "${vuln_json_file}" 2>/dev/null | wc -l || echo "0")
+    high=$(grep -o '"Severity": "HIGH"' "${vuln_json_file}" 2>/dev/null | wc -l || echo "0")
+    medium=$(grep -o '"Severity": "MEDIUM"' "${vuln_json_file}" 2>/dev/null | wc -l || echo "0")
+    low=$(grep -o '"Severity": "LOW"' "${vuln_json_file}" 2>/dev/null | wc -l || echo "0")
+
+    svc_total=$((critical + high))
+    TOTAL_VULNS=$((TOTAL_VULNS + svc_total))
```
yannickmonney: no abbr

services/crawler/Dockerfile
```diff
@@ -1,29 +1,110 @@
...
 ARG VERSION=dev
+# Set to 'true' to include CJK (Chinese/Japanese/Korean) font support (~100 MB)
+# Most crawl targets are Latin/Cyrillic — enable only if crawling CJK sites
+ARG INSTALL_CJK_FONTS=false
 
-FROM python:3.11-slim
+# =============================================================================
+# Stage 1: BUILDER — install dependencies, download browsers
+# =============================================================================
+FROM python:3.11-slim AS builder
```
yannickmonney: Try cgr.dev/chainguard/python should be better

services/crawler/Dockerfile
```diff
@@ -1,29 +1,110 @@
...
+    find /usr/local/lib/python3.11/site-packages -type d -name "testing" -exec rm -rf {} + 2>/dev/null || true; \
+    # 7. Strip debug symbols from compiled .so files (saves ~20-40 MB)
+    find /usr/local/lib/python3.11/site-packages -name "*.so" -exec strip --strip-debug {} \; 2>/dev/null || true; \
+    # 8. Remove .dist-info directories (package metadata not needed at runtime, saves ~5 MB)
+    find /usr/local/lib/python3.11/site-packages -type d -name "*.dist-info" -exec rm -rf {} + 2>/dev/null || true
+
+# =============================================================================
+# Stage 2: RUNTIME — minimal final image
+# =============================================================================
+FROM python:3.11-slim AS runtime
```
yannickmonney: Try cgr.dev/chainguard/python should be better

yannickmonney (changes requested):

services/db/Dockerfile
```diff
@@ -49,38 +79,30 @@ RUN chmod +x /usr/local/bin/docker-entrypoint-wrapper.sh \
...
 EXPOSE 5432
 
-# Health check — waits for both PostgreSQL readiness and migration completion
 HEALTHCHECK --interval=10s --timeout=5s --start-period=90s --retries=10 \
     CMD pg_isready -U ${DB_USER:-tale} -d ${DB_NAME:-tale} && test -f /tmp/.db_ready || exit 1
 
-# Use custom entrypoint wrapper that handles DB_ env vars
 ENTRYPOINT ["/usr/local/bin/docker-entrypoint-wrapper.sh"]
 CMD ["postgres"]
-
```
larryro: The `FROM scratch` on line 45 drops all upstream metadata from the ParadeDB/PostgreSQL base image, including `STOPSIGNAL SIGINT`. I can see you re-declared `ENV`, `ENTRYPOINT`, `CMD`, `EXPOSE`, `HEALTHCHECK` — but `STOPSIGNAL` is missing.

PostgreSQL uses SIGINT for fast shutdown. Without it Docker sends SIGTERM (its default), which PG treats as "smart shutdown" — it waits for all clients to disconnect first. If any connection hangs, PG won't stop within Docker's 10s timeout and gets SIGKILL'd.

Can you add `STOPSIGNAL SIGINT` somewhere before ENTRYPOINT? One-liner fix.

services/crawler/Dockerfile
```diff
@@ -45,39 +123,43 @@ RUN apt-get update && apt-get install -y \
...
+# Copy cleaned Python site-packages from builder
+COPY --from=builder /usr/local/lib/python3.11/site-packages /usr/local/lib/python3.11/site-packages
+COPY --from=builder /usr/local/bin /usr/local/bin
 
-# Install Playwright browsers (Chromium)
-# Note: We already installed dependencies manually above, so we skip `playwright install-deps`
-# which has compatibility issues with newer Debian versions (missing ttf-unifont, ttf-ubuntu-font-family)
-RUN playwright install chromium
+# Copy ONLY the headless shell browser binary from builder (not full Chrome, not FFmpeg)
+COPY --from=builder /root/.cache/ms-playwright /root/.cache/ms-playwright
```
larryro: The comment on line 11 says "Remove full Chrome UI (keep headless_shell which Playwright defaults to, saves ~364 MB)" and this line says "Copy ONLY the headless shell browser binary" — but the actual COPY grabs the entire `/root/.cache/ms-playwright` directory, which still includes full Chrome.

I don't see any `rm -rf` targeting the full Chrome binary before this COPY. The FFmpeg removal is there (line 62) but Chrome itself is untouched. So the ~364 MB savings claimed in the header aren't actually realized.

Either implement the extraction (remove the chrome binary, keep only headless_shell) or update the comments to reflect what's actually happening.

.github/workflows/release.yml
```diff
@@ -125,11 +131,77 @@ jobs:
...
           build-args: |
             VERSION=${{ needs.prepare.outputs.version_number }}
+          labels: |
+            org.opencontainers.image.created=${{ steps.meta.outputs.created }}
+            org.opencontainers.image.revision=${{ steps.meta.outputs.revision }}
+            org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}
           provenance: false
+          cache-from: type=gha,scope=${{ matrix.service.name }}-release
+          cache-to: type=gha,mode=max,scope=${{ matrix.service.name }}-release
+
```
larryro: The cache scope here is `${{ matrix.service.name }}-release` but the build matrix includes both amd64 and arm64. Both architectures write to the same scope, so whichever finishes last overwrites the other's cache. Next release, the "losing" arch gets a cache full of layers from the wrong platform — total miss.

Should be something like:
```yaml
cache-from: type=gha,scope=${{ matrix.service.name }}-release-${{ matrix.arch.name }}
cache-to: type=gha,mode=max,scope=${{ matrix.service.name }}-release-${{ matrix.arch.name }}
```

.github/workflows/build.yml
```diff
@@ -88,17 +110,178 @@ jobs:
...
+          file: services/${{ matrix.service }}/Dockerfile
+          platforms: linux/amd64
+          push: false
+          load: true
+          tags: tale-${{ matrix.service }}:scan
+          cache-from: type=gha,scope=${{ matrix.service }}
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@master
+        with:
```
larryro: This is the only action in the entire workflow pinned to a branch (`@master`) rather than a version tag. Every other action uses `@v6`, `@v4`, etc. Tracking master means any upstream commit — including a bad one — runs in our CI immediately.

Could you pin this to a release tag like `@0.28.0` or whichever version you tested with?

tests/container-vulnerability-scan.sh
```diff
@@ -0,0 +1,167 @@
...
+    # Single scan producing JSON (derive all outputs from this)
+    vulnerability_json_file="${OUTPUT_DIR}/${svc}-report.json"
+    trivy image \
+        --severity "CRITICAL,HIGH,MEDIUM,LOW" \
+        --format json \
+        --output "${vulnerability_json_file}" \
+        ${IGNOREFILE_ARG:-} \
+        "$img" 2>/dev/null
+    scan_exit=$?
+
```
larryro: This line is dead code. `set -euo pipefail` is active (line 22), so if `trivy image` exits non-zero, the script terminates before reaching `scan_exit=$?`. And with `2>/dev/null` on the trivy call, there's zero diagnostic output — the script just vanishes.

Also, `VULNERABILITY_EXIT_CODE` (line 27) is parsed and printed in the summary but never actually used in any exit path — line 168 exits with `$SCAN_FAILED`, ignoring `$EXIT_CODE` entirely. So the "blocking mode" advertised in the header doesn't work.

Fix for the first issue: wrap the trivy call with `set +e` / `set -e`, or use `|| true` and check the json output. For the second: wire `EXIT_CODE` into the actual exit logic when vulnerabilities are found.

saifaliunity: Fixed.

yannickmonney (approved):

PR #1395, opened 2026-04-11
fix(platform): clear isPending on successful message send

### LINKED ISSUE #1339: Chat Input Field Becomes Non-Editable After Sending Message
Description

After sending a message in the chat, the input field becomes non-editable, preventing the user from correcting or modifying their message.

Steps to Reproduce
Open chat page
Type a message
Send the message
Try to edit or modify the sent message
Actual Result
Sent message cannot be edited
No edit option available

yannickmonney:
## Summary

- Fix chat input becoming stuck/disabled after sending a message by adding explicit `setIsPending(false)` calls in the success paths of `useSendMessage`
- The catch block already cleared pending state via `clearChatState()`, but the success paths relied solely on the reactive handoff in `useChatLoadingState`, which could miss on slow networks or when React batched state updates
- Add tests verifying `isPending` is cleared on successful standard send, new thread creation send, and arena mode send

## Test plan

- [x] Lint passes (`bun run --filter @tale/platform lint`)
- [x] Typecheck passes (`bun run --filter @tale/platform typecheck`)
- [x] Server tests pass (156 files, 1955 tests)
- [x] Client tests pass (245 files, 1855 tests)
- [ ] Manual: send a message and verify the input field re-enables after the AI response begins
- [ ] Manual: send a message on a slow network and verify the input does not stay stuck
- [ ] Manual: send a message in arena mode and verify both panels load and input re-enables

Closes #1339

### DIFF
--- a/services/platform/app/features/chat/hooks/__tests__/use-send-message.test.ts
+++ b/services/platform/app/features/chat/hooks/__tests__/use-send-message.test.ts
@@ -132,6 +132,58 @@ describe('useSendMessage — error handling', () => {
     expect(params.setIsPending).not.toHaveBeenCalled();
   });
 
+  it('clears isPending on successful send', async () => {
+    const params = createParams();
+    const { result } = renderHook(() => useSendMessage(params));
+
+    await act(async () => {
+      await result.current.sendMessage('Hello');
+    });
+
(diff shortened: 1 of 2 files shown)


yannickmonney:
@coderabbitai review

PR #1675, opened 2026-05-06
fix(platform,rag): authenticate RAG service + add SSRF guard

larryro:
## Summary

Independently-deployable security hotfix. RAG service shipped with **zero authentication** on all 13 endpoints (including DELETE on `/api/v1/documents/{file_id}`); host-published on port 8001 in `compose.yml`. Any caller reachable on the network could read or destroy any deployment's indexed corpus.

This PR closes that gap with a Bearer-token contract between the platform and RAG containers + a small SSRF guard on the platform-side `RAG_URL` so cloud-metadata pivot attacks are refused.

- **RAG (Python)**: global `verify_internal_token` FastAPI dependency on documents/search/llm_cache routers; `/health` and `/` exempt for liveness probes. Default `tale-rag-dev-only` baked into the Dockerfile so a fresh `docker compose up` works with no setup. Loud `[SECURITY]` warning every restart while default is in use. `RAG_REQUIRE_CUSTOM_INTERNAL_TOKEN=true` opts into strict prod posture.
- **Platform (TS/Convex)**: `getRagConfig()` rewritten — sync, V8-runtime-safe, validates `RAG_URL` against `169.254.0.0/16` (link-local / IMDS) + `0.0.0.0/8` ("this" network) CIDR blocklist using `ipaddr.js` (encoding bypasses auto-normalize). New `ragFetch()` helper sets the Authorization header automatically. **All 14 callers migrated** + 5 helpers refactored to drop the manual `serviceUrl` parameter.

## Deployment note

Operators must upgrade BOTH the platform and RAG containers in the same release: the new RAG image 401s un-authed callers; the old platform image doesn't send the Authorization header. The Docker-baked default ensures a coordinated upgrade works without env changes; production operators override `RAG_INTERNAL_TOKEN` (matching on both services) once both images are running.

## Compatibility note

DNS-rebinding pinning (the original plan's undici-Agent design) is deferred — undici / `node:dns` are unavailable in Convex's V8 runtime and 5 RAG callers run V8. The URL-string SSRF check covers the realistic threat model; DNS rebinding requires the operator to set `RAG_URL` to an attacker-controlled hostname (a deeper compromise).

## Test plan

- [ ] `bun run check` (format / lint / typecheck / tests) passes locally
- [ ] Fresh `docker compose up` works without setting `RAG_INTERNAL_TOKEN` (warn-log appears in `docker logs`)
- [ ] Setting `RAG_INTERNAL_TOKEN=secret-foo` on both RAG and platform containers, `docker compose restart` — chat continues to work; warning gone
- [ ] Setting only one side starts auth-mismatch — RAG returns 401; platform logs the failure
- [ ] `RAG_URL=http://169.254.169.254/metadata` refused at platform startup with link-local CIDR error
- [ ] `RAG_URL=http://0xa9fea9fe/metadata` (octal IMDS) refused (`ipaddr.js` normalizes)
- [ ] Existing tests in `services/rag/tests/` and `services/platform/convex/governance/__tests__/` continue to pass

## Release Notes

* **Infrastructure & Security**
  * Implemented internal token authentication for RAG service endpoints
  * Enhanced RAG configuration with centralized token and URL management
  * Added SSRF protection utilities to validate service URLs

* **Refactor**
  * Consolidated RAG API integration by centralizing URL handling, removing the need to pass service URLs throughout the codebase

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -105,6 +105,12 @@
         "lucide-react": "1.8.0",
         "tailwind-merge": "3.5.0",
       },
+      "peerDependencies": {
+        "vitest": "*",
+      },
+      "optionalPeers": [
+        "vitest",
+      ],
     },
     "services/crawler": {
       "name": "@tale/crawler",
@@ -182,6 +188,7 @@
         "i18next": "26.0.4",
         "i18next-icu": "2.4.3",
         "intl-messageformat": "11.2.0",
+        "ipaddr.js": "^2.4.0",
         "jexl": "2.3.0",
         "jose": "6.2.2",
         "json-diff-kit": "1.0.35",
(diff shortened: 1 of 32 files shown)


larryro:
Superseded by #1676 — Phase 4 (RAG auth + SSRF guard) was merged into the unified data-protection-retention branch. All 14 phases ship as one PR.

PR #1678, opened 2026-05-06
refactor(platform): toast positioning, automations created column, UI polish

Israeltheminer:
## Summary

- Add `position` prop to toast (`top-right` default, `top-center` for auth flows: log in, sign up, 2FA, forgot-password hint)
- Replace automations table `version` column with a localized `created` date — derived from the earliest history snapshot epoch, falling back to file birthtime/mtime when no history exists yet
- Wire the automations table through `DataTable`'s built-in loading / empty / infinite-scroll props instead of bespoke branches; introduce an `icon-text` skeleton variant
- Replace the forgot-password helper text with a click-to-toast hint that explains password reset is admin-managed
- Reorder integrations panel actions: **Test connection** and **Delete** live in the body, **Disconnect** in the footer
- Migrate raw `<select>` and `<input>` to shared `Select` / `Input` / `SearchableSelect` components in providers, agents, prompts
- Drop `size="lg"` overrides on auth pages, brighten the default toast info icon
- Add `searchPlaceholder` keys for products / customers / vendors / websites, plus `automations.entityLabel`; translate the new forgot-password toast in `de` and `fr`

## Pre-PR checklist

- [x] Ran `bun run check` (format, lint, typecheck, all tests). Lint + typecheck pass. Test suite has 5 pre-existing unrelated failures (PII regex ReDoS perf budget + a `html-encoding-sniffer` ESM/CJS pool issue blocking 1 dashboard layout test). The touched `automations-table.test.tsx` could not run locally due to the same pool issue — CI runs in a clean environment.
- [x] Updated `services/platform/messages/{en,de,fr}.json` for new keys (`forgotPasswordToast`, `entityLabel`, `searchPlaceholder` × 4).
- [N/A] No public-facing config or API changes that require docs updates.
- [N/A] `bun run --filter @tale/docs lint` — no docs changed.
- [N/A] No README impact.

## Test plan

- [ ] Log in / sign up / 2FA flows — toasts appear at top-center
- [ ] Forgot-password link on log-in page opens the explanatory toast
- [ ] Automations list shows "Created" date, sorted/empty states behave like other tables, skeletons render an icon-text shape, infinite-scroll counter shows total
- [ ] Provider settings: image-generation mode select renders via `Select`
- [ ] Integrations details panel: Test connection and Delete are in the body, Disconnect is in the footer; behavior unchanged
- [ ] German + French translations render correctly on each touched screen

## Summary by CodeRabbit

* **New Features**
  * Toast notifications now support top-center positioning
  * Search placeholders and entity labels added for automations, products, customers, vendors, and websites
  * Workflow creation dates now tracked and displayed in the automations table
  * Image generation mode selector added to provider settings

* **Improvements**
  * Automations table redesigned with enhanced loading states and intuitive layout
  * Login and sign-up pages refined for better usability
  * Integration panel reorganized with clearer test, disconnect, and delete actions
  * Enhanced loading skeleton states in data tables for improved visual feedback


Israeltheminer:
Addressing CodeRabbit feedback in 4e27f4f28:

**Applied:**
- `use-automations-table-config.tsx:65` — switched to `ms !== undefined` so a `0` epoch still renders.
- `providerName.tsx:1057` — replaced the `as` assertion with an explicit value-equality guard, removing the `oxlint-disable` and complying with the project rule against type assertions.

**Skipped (with reason):**
- `data-table-filters.tsx:167` — the `sm:max-w-[16rem]` cap is the intended new width for the filter row; the skeleton's `sm:w-[18.75rem]` is shared by other tables, so I'd rather not flip the skeleton for one consumer. Width jump on this specific table only happens for ~one frame on first paint and was pre-existing.
- `integration-panel.tsx:268` — the disconnect button markup was relocated, not introduced by this PR; adding `role="status"`/`aria-live` is worth doing but should land in a focused a11y pass alongside the other auth/integration spinners.
- `providerName.tsx:1063-1066` — `images-api` and `chat-multimodal` are API enum values from the provider config schema, not user-facing copy. Localizing them would mistranslate identifiers users may need to copy/paste from docs.
- `file_actions.ts:106` — the docstring already calls out that this is best-effort. Persisting a stable "first-write" timestamp would mean schema migration on every existing workflow file, out of scope for this PR. Tracking properly would warrant a follow-up.

PR #1699, opened 2026-05-11
feat(platform): public /status page with per-component health

larryro:
## Summary

Adds an unauthenticated `/status` page showing overall up/down plus a per-component breakdown using neutral product vocabulary (no stack names leaked).

- Server-rendered HTML, no JS, no auto-refresh
- 5s single-flight probe cache bounds upstream load on a public surface
- Probes Convex `/version` (status-only — body is plain text), RAG/Crawler `/health` (lenient 2xx)
- Response bodies discarded — a misbehaving upstream can't push bytes into the public response
- `en` / `de` / `fr` copy via `Accept-Language` prefix
- Component labels are nouns (Application / Knowledge base / Web & document services) so each label covers all failure modes of that subsystem without exposing Convex / RAG / Crawler publicly
- A11y: status conveyed by text + colour (status dots are `aria-hidden`)

## How it routes

- Caddy: `/status` falls through to the default `reverse_proxy platform:3000` — no Caddyfile change needed (avoids the `/api/*` → Convex collision)
- Production: `app.get('/status', ...)` in `server.ts`, before the SPA catch-all
- Dev: matching `vite-plugins/serve-status.ts` mirrors the Hono route (the same pattern used by `serve-canvas-preview.ts` / `serve-branding-images.ts`)

Defaults align with the platform's localhost convention (`rag_config.ts`, `get_crawler_service_url.ts`, `vite.config.ts`); Docker compose env vars override to in-network DNS.

## Test plan

- [ ] `bun run check` — green
- [ ] `bun run dev` → open `http://localhost:3000/status` → shows the page (likely "Partial degradation" since RAG/Crawler aren't running locally by default)
- [ ] `curl -s -H 'Accept-Language: de-DE' http://localhost:3000/status | grep '<title>'` → German title
- [ ] `curl -s -H 'Accept-Language: fr-FR' http://localhost:3000/status | grep '<title>'` → French title
- [ ] `curl -I http://localhost:3000/status` → `Cache-Control: public, max-age=5`, `Content-Type: text/html`
- [ ] Full Docker stack (`tale start`) → `$SITE_ORIGIN/status` works through Caddy
- [ ] Stop one backend service → banner flips to "Partial degradation" within 5s
- [ ] Stop all three → banner flips to "Service outage"

## Out of scope (filed for follow-up)

- Per-user usage metric labelling — `requestCount` is model invocations, not user prompts
- Org-level "aggregate-only" usage mode (Betriebsrat / § 87 BetrVG)

## Summary by CodeRabbit

* **New Features**
  * Published a public `/status` endpoint that monitors platform health in real-time, displaying the status of core services. The status page offers multilingual support (English, German, French) and includes intelligent caching to optimize performance and reduce backend load.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tale-project/tale/pull/1699)

### DIFF
--- a/docs/de-CH/develop/api-reference.md
+++ b/docs/de-CH/develop/api-reference.md
@@ -293,3 +293,37 @@ Liste der verfügbaren Agents (Modelle).
   ]
 }
 ```
+
+## Status-Endpoints
+
+Zwei öffentliche, nicht authentifizierte Endpoints geben den Gesamt-Up/Down-Status der Plattform aus. Sie teilen sich denselben Probe (5-Sekunden-In-Memory-Cache) und unterscheiden sich nur in der Darstellung:
+
+| Endpoint       | Verwendung                                                                                                                  |
+| -------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `/status`      | Lesbare HTML-Statusseite. Wählt Englisch, Deutsch oder Französisch anhand von `Accept-Language`.                            |
+| `/status.json` | Maschinenlesbarer Feed für externe Monitore (BetterStack, UptimeRobot, Atlassian Statuspage, Datadog Synthetics und andere) |
+
+Beide Endpoints antworten immer mit `200 OK` und `Cache-Control: public, max-age=5`. Die Plattform selbst ist die Quelle der Wahrheit — wenn dein Monitor `/status.json` überhaupt nicht erreicht, ist der Plattformprozess nicht erreichbar, und das Timeout des Monitors ist das Signal.
+
+### Wire-Format (`/status.json`)
+
+```json
+{
+  "status": "operational",
+  "checkedAt": "2026-05-11T13:45:07.123Z",
+  "components": [
+    { "id": "convex", "status": "operational" },
+    { "id": "rag", "status": "operational" },
+    { "id": "crawler", "status": "outage" }
+  ]
+}
+```
+
+| Feld                  | Typ    | Werte                                                                                |
+| --------------------- | ------ | ------------------------------------------------------------------------------------ |
+| `status`              | string | `operational`, `degraded` (einige Komponenten ausgefallen) oder `outage` (alle aus). |
+| `checkedAt`           | string | ISO-8601-Zeitstempel der letzten Probe-Runde.                                        |
+| `components[].id`     | string | Stabile Komponenten-Kennung: `convex`, `rag` oder `crawler`.                         |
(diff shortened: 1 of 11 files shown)


yannickmonney:
@larryro How does it deal with blue-green deployments? Also you'd need to store the status over time and show current/past incidents, like for example: https://status.claude.com/

larryro:
> @larryro How does it deal with blue-green deployments? Also you'd need to store the status over time and show current/past incidents, like for example: https://status.claude.com/

@yannickmonney I'd argue we shouldn't build that into the service itself. Storing status history and incidents inside the same service we're monitoring has a shared-fate problem — if Convex goes down, the status page goes with it, which is the worst possible time for it to be unavailable.

The status.claude.com link is actually a good argument for this: it's running on Atlassian Statuspage (see the footer), deliberately hosted off Anthropic's own infrastructure. If we want history and incident tracking, we should adopt a similar external service rather than reimplement it.

larryro:
> @larryro How does it deal with blue-green deployments? Also you'd need to store the status over time and show current/past incidents, like for example: https://status.claude.com/

@yannickmonney pushed 64c74ade6 — adds `/status.json` as a sibling to `/status`, projecting the same canonical `StatusFeed` model the HTML renders from (single source of truth; the human view and the machine view can't drift). Always 200 + body; keyword monitors can match `"status":"operational"` / `"status":"outage"`.

That operationalises the external-service path. Self-hosters wire `/status.json` into BetterStack / UptimeRobot / Atlassian Statuspage / Datadog Synthetics / OpenStatus, and that service owns the timeline + incident pages — same shape status.claude.com uses (Statuspage running off Anthropic's own infra).

On blue-green — platform redeploys are transparent to the monitor. Caddy stays in front, health-checks `/api/health` every 2s, and routes to whichever color is healthy via the `platform` DNS alias (`services/proxy/Caddyfile`). The only sliver is the brief transition window when the old container has set its shutdown marker (`/api/health` → 503) and the new color isn't yet ready; Caddy intercepts the 5xx and serves `/var/www/maintenance.html` per its `handle_errors` block — a keyword-on-`"status":"operational"` monitor reads that as not-operational, which is accurate.

Stateful services (Caddy itself, Postgres, Convex) are a different story: there is no blue-green for those, so during a restart the whole stack is unreachable. The external monitor reports "down" for that window — that is the truth, and the expected behaviour for a self-hosted single-instance stack.

Docs: "Status endpoints" in `docs/{en,de,fr,de-CH}/develop/api-reference.md`.

Issue #1803, opened 2026-06-01
Epic: Security & compliance baseline for SOC 2 / ISO 27001 audit


yannickmonney:
## Goal

Close the control gaps a **SOC 2 Type II / ISO 27001** audit will test, by the audit deadline. Each child issue maps to one or more Trust Services Criteria / Annex A controls. This epic is the single traceability anchor for the auditor (control → issue → PR → test/doc).

## Control mapping (auditor evidence index)

| Child | Control area | SOC 2 TSC | ISO 27001:2022 Annex A |
|-------|--------------|-----------|------------------------|
| #1498 | Vulnerability mgmt / patch SLA | CC7.1, CC8.1 | A.8.8 technical vuln mgmt |
| #1502 | Session management / idle timeout | CC6.1 | A.8.5 secure authentication |
| #1505 | Audit logging integrity & retention | CC7.2, CC7.3 | A.8.15 logging, A.8.16 |
| #1506 | Federated identity / SSO | CC6.1, CC6.2 | A.5.16 identity mgmt |
| #1508 | Strong authentication (phishing-resistant 2FA) | CC6.1 | A.8.5 |
| #1509 | Data-at-rest protection (PII posture) | CC6.7 | A.8.24 cryptography |
| #1566 | Brute-force / lockout (auth UX) | CC6.1 | A.8.5 |

## Prerequisite (already merged — evidence only)

- #1507 — TOTP two-factor authentication. Provides the second-factor primitive #1508 extends and the `two_factor_policy` governance type #1502 sits beside. Link the PR as audit evidence.

## Child issues (sequenced)

**Wave 0 — immediately actionable (parallel):**
- [ ] #1509 — PII at-rest policy. **Decision: Option B** (document the Convex at-rest + TLS + RLS posture; field-encryption reserved for secrets). Docs/ADR only.
- [ ] #1505 — Audit-log hardening: scheduled integrity-check cron (alert on failure) + append-only test. (Hash chain, manual verify, retention, signed checkpoints already exist.)
- [ ] #1498 — Ratchet the CVE gate CRITICAL→HIGH once the pre-existing HIGH backlog clears; resolve the `xlsx` (SheetJS) parse-path advisory (replace parser or scoped-suppress + compensating control).

**Wave 1 — core access-control gaps:**
- [ ] #1502 — Org-admin **session idle timeout**: governance policy + server-side enforcement + client watchdog.
[...]

Issue #1506, opened 2026-04-13
Feature: Generic OIDC identity provider (Keycloak, Auth0, Okta, Google)


larryro:
## Context

Compliance requires integration with recognized identity management systems (AD, Keycloak) and support for OpenID Connect, OAuth 2.0, and SAML.

Current state:
- Entra ID (Azure AD) OAuth2/OIDC is implemented in [services/platform/convex/sso_providers/entra_id/](services/platform/convex/sso_providers/entra_id/).
- Generic trusted-headers auth covers Keycloak behind a reverse proxy: [services/platform/convex/betterAuth/trusted_headers/](services/platform/convex/betterAuth/trusted_headers/).
- No generic OIDC adapter — any new IdP requires either trusted-headers or a bespoke adapter.

## Proposed change

Generalize the Entra ID code into a generic OIDC provider adapter supporting:

- Discovery URL (`.well-known/openid-configuration`)
- Client ID / secret
- Scope configuration
- Claim mapping for role/team/name/email
- Multiple instances per org (e.g. one Keycloak + one Google)

Reuse the existing `sso_providers` registry pattern. Keep Entra ID as a pre-configured shortcut.

## Acceptance criteria

- [ ] Admin can configure a generic OIDC provider with discovery URL, client credentials, and claim mappings.
- [ ] Login via Keycloak works end-to-end in a test environment.
- [ ] Claim-to-role mapping is configurable.
- [ ] Existing Entra ID configs continue to work.

Part of the security compliance baseline. (SAML is tracked separately if there is demand.)

yannickmonney:
Triaged — this is a valid, high-value feature and we'll keep it open. Today only the Microsoft Entra adapter ships (`convex/sso_providers/entra_id/adapter.ts`); there's no generic OIDC adapter (no discovery-URL / `.well-known` handling), so Keycloak/Auth0/Okta/Google currently only work via trusted-headers mode. The `sso_providers` registry + `validate_sso_config.ts` + `upsert_sso_provider.ts` give a clean extension seam to add a generic OIDC provider (Authorization Code + PKCE, configurable claim mapping). Assigned to me.

PR #1809, opened 2026-06-01
feat(platform): add generic OIDC identity provider adapter

yannickmonney:
## What

Adds a discovery-based **generic OpenID Connect** SSO adapter alongside the existing Microsoft Entra ID provider, so any OIDC issuer (Keycloak, Auth0, Okta, Google, …) can be configured. Part of the security & compliance baseline epic.

## Backend

- **`oidc_discovery.ts`** — resolves the authorization / token / userinfo endpoints from the issuer's `.well-known/openid-configuration` **per auth flow**. No endpoint is persisted, so there is **no schema migration**.
- **`generic_oidc/adapter.ts`** — implements `SsoProviderAdapter`: forces the `openid` scope, maps `sub → externalId` / `email` / `name` (with given/family/preferred_username fallback) / `groups`, and reuses the provider-agnostic role matcher. Registered as `'generic-oidc'`.
- **Interface change** — `buildAuthorizeUrl` is now async-capable and `getUserInfo` / `getGroups` / `getAppRoles` take the provider `config` so discovery-driven adapters can resolve endpoints. Entra signatures and all call sites (authorize handler, callback handler, team sync, login) updated.
- **`validate_sso_config.ts`** — now dispatches to the selected adapter (Entra Graph client-credentials probe vs. OIDC discovery) instead of hardcoding a Microsoft Graph scope.

## UI

- `sso-config-dialog.tsx` gains a **provider-type selector**. Entra-only controls (OneDrive/SharePoint, Graph team sync, exclude groups) are gated to Entra; the generic path sends standard OIDC scopes and persists no provider features. Issuer/client help text and the connected-state copy adapt per type.
- `en` / `de` / `fr` strings added (de-CH inherits from de).

## Tests

- `generic_oidc/adapter.test.ts` — discovery, code→token exchange, userinfo claim mapping (incl. name fallbacks), and `validateConfig` success / unreachable-issuer / missing-userinfo paths (7 cases).
- `use-sso-config-form.test.ts` — provider-type branching: Entra sends Graph scopes + features, generic sends standard scopes + no features, the test path targets the correct adapter, and the type is derived from a loaded config (4 cases).

## Verification

`bun run --filter @tale/platform typecheck` · `lint` · `test` (server + ui) all green. CodeRabbit review applied (scope-selection helper deduped; load-mapping test added).

Refs #1506. Tracked under the security baseline epic #1803.

## Release Notes

* **New Features**
  * Added Generic OpenID Connect (OIDC) as an SSO provider option alongside Entra ID.
  * SSO configuration UI now displays provider-specific icons, descriptions, and help text based on selected provider type.
  * Generic OIDC automatically discovers endpoints from issuer configuration.

* **Improvements**
  * Enhanced SSO configuration validation to support multiple provider types.
  * Added multi-language support (English, French, German) for Generic OIDC provider.

### DIFF
--- a/services/platform/app/features/settings/integrations/components/sso-config-dialog.tsx
+++ b/services/platform/app/features/settings/integrations/components/sso-config-dialog.tsx
@@ -7,7 +7,7 @@ import { IconButton } from '@tale/ui/icon-button';
 import { Center, HStack, Stack } from '@tale/ui/layout';
 import { StatusIndicator } from '@tale/ui/status-indicator';
 import { Text } from '@tale/ui/text';
-import { CheckCircle, Loader2, X, XCircle } from 'lucide-react';
+import { CheckCircle, KeyRound, Loader2, X, XCircle } from 'lucide-react';
 
 import { MicrosoftIcon } from '@/app/components/icons/microsoft-icon';
 import { FormSection } from '@/app/components/ui/forms/form-section';
@@ -21,7 +21,10 @@ import type {
 } from '@/lib/shared/schemas/sso_providers';
 import { narrowStringUnion } from '@/lib/utils/type-guards';
 
-import { useSsoConfigForm } from '../hooks/use-sso-config-form';
+import {
+  type SsoProviderType,
+  useSsoConfigForm,
+} from '../hooks/use-sso-config-form';
 import { RoleMappingSection } from './sso-config/role-mapping-section';
 
 interface SSOConfigDialogProps {
@@ -41,6 +44,8 @@ export function SSOConfigDialog({
     t,
     tCommon,
     platformRoles,
+    providerType,
+    setProviderType,
     issuer,
     setIssuer,
     clientId,
@@ -75,6 +80,8 @@ export function SSOConfigDialog({
     organizationId,
     existingProvider,
   });
+
+  const isGeneric = providerType === 'generic-oidc';
 
   const isFormValid = isConnected
     ? !!issuer?.trim() && !!clientId?.trim()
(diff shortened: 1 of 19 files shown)


yannickmonney:
Addressed the CodeRabbit review (pushed):

- **Missing `sub` claim** — `getUserInfo` now fails fast with a clear error instead of producing `externalId: "undefined"` (which would collide every such user onto one account). Accepts string or numeric `sub`. Test added.
- **Discovery caching** — added a short-lived (5 min) in-memory cache keyed by issuer in `oidc_discovery.ts`, so a single login callback (token exchange + userinfo + group sync) folds into one network call. Best-effort per isolate, never persisted. Test added asserting one `.well-known` fetch across calls. This also resolves the "multiple discovery fetches per login" note on the callback handler.
- **Fetch timeout** — discovery, token, and userinfo fetches now use `AbortSignal.timeout(10_000)` (matching the `oauth2.ts` convention) so a stuck IdP socket fails fast instead of hanging the flow.
- **Test type-assertion** — replaced `init?.body as URLSearchParams` with an `instanceof URLSearchParams` narrow.

Skipped (with reason):
- **i18n the `your-client-id` / `All-Employees, Domain-Users` placeholders** — every placeholder in this dialog (the tenant-GUID example, the issuer URL, the groups example) is a literal example, and CodeRabbit's own guidance is to leave format/example literals as-is. Routing one through i18n while the others stay literal would be inconsistent, so I've left the established pattern intact.
- **Entra adapter fetch timeouts** — pre-existing code outside this PR's scope; worth a separate hardening pass.

Issue #1566, opened 2026-04-18
Bug: Login rate limiting locks account silently — user receives no progressive warning during failed attempts


huzaifaWD20:
### What happened?

When a user enters incorrect passwords repeatedly, the account is correctly locked after the threshold is reached. However, during attempts 1 through 5, the user sees only the generic "Wrong email or password" message with no indication that they are approaching a lockout. The account locks silently from the user's perspective. Only the admin receives a notification about the lockout — the user themselves gets no proactive warning.

Actual behavior
Attempts 1-5 show only "Wrong email or password." Attempt 6 shows "Account temporarily locked. Try again in 3 minutes, or contact an administrator." No progressive warning at any point. Admin receives a notification but the affected user receives none until after lockout.

### Steps to reproduce

Steps to reproduce

1. Log out of the platform
2. Enter incorrect password 5 times in succession on the login page
3. Observe the error messages shown during each attempt
4. Attempt login a 6th time

### Expected behavior

Expected behavior
After 2-3 failed attempts, the user should see a progressive warning such as: "2 more failed attempts will temporarily lock your account." This is standard security UX practice and ensures the user is not surprised by a lockout they had no warning about.

### Affected service

Platform

### Version or commit

v0.2.46

### Relevant log output

```shell

```

### Before submitting

- [x] I searched existing issues and this hasn't been reported yet.

yannickmonney:
Triaged — keeping open as a security-UX discussion item (re-labeled from bug). The "silent until locked" behavior is partly intentional: the login path uses per-account exponential lockout + per-IP rate limiting with deliberate timing fuzzing, and disclosing "N attempts remaining" is generally considered an info-leak that aids brute-forcing. So a progressive warning is a deliberate trade-off rather than a clear win — worth discussing what (if anything) to surface to the legitimate user without helping an attacker. Assigned to me.

yannickmonney:
Closing as **won't-implement (by design)** for the emailed alert, with the core UX already shipped.

**What shipped** (#1811): on a failed login the user now sees a lockout-risk advisory, and on lockout the org admins get an in-app + Slack security notification. Uniform error messages are kept deliberately (no attempts-remaining counter) to preserve the anti-enumeration posture.

[...]

PR #1811, opened 2026-06-01
feat(platform): advise of lockout risk on a failed login

yannickmonney:
## What

Addresses #1566. A failed sign-in gave the user no hint that repeated failures lock the account, so the lockout felt like it came out of nowhere — only the admin was notified.

Now a failed **credential** attempt shows a calm, muted advisory beneath the error:

> Repeated failed attempts will temporarily lock the account.

It is not shown once the account is already locked (the lockout countdown message stands on its own).

## Why not an "N attempts remaining" counter

The issue requested a progressive counter. In an audited deployment that is a security regression, so it is deliberately **not** implemented:

- A per-attempt counter only appears for real accounts, breaking the uniform "wrong email or password" response and reintroducing **account enumeration** (the very thing the existing jitter + uniform-404 mitigations remove).
- It also hands an attacker an exact **brute-force budget** per account.

The standing advisory conveys the consequence (so the user isn't surprised) without leaking either signal. An emailed owner alert was considered but the platform has no transactional-email infrastructure (`requireEmailVerification: false`, password reset is admin-managed), so it's out of scope here.

## Changes

- `log-in.tsx`: a `showLockoutHint` state, set on a failed credential attempt and cleared on lockout / new submit; muted advisory line under the error.
- `en`/`de`/`fr`: `auth.login.lockoutAdvisory`.

## Verification

`typecheck` · `lint` · i18n parity/usage green.

Refs #1566.

### DIFF
--- a/services/platform/app/routes/_auth/log-in.tsx
+++ b/services/platform/app/routes/_auth/log-in.tsx
@@ -107,6 +107,11 @@ export function LogInPage() {
   });
 
   const [loginError, setLoginError] = useState<string | null>(null);
+  // Standing, count-free advisory shown on a failed credential attempt. We
+  // deliberately never reveal attempts-remaining: a live counter would break
+  // the uniform "wrong email or password" response and hand an attacker both
+  // account-enumeration and a brute-force budget (#1566).
+  const [showLockoutHint, setShowLockoutHint] = useState(false);
 
   const { isSubmitting, isValid } = form.formState;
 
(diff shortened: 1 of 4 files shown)


yannickmonney:
Pre-emptively ran the CodeRabbit CLI and applied both findings: the post-await fallback now tracks an `authErrorHandled` flag instead of reading the stale `loginError` closure value (so a 429 lockout message can't be overwritten), and the advisory `<p>` got `aria-live="polite"` so screen readers announce it. `text-muted-foreground` matches the app's existing secondary/helper text token.

Issue #1502, opened 2026-04-13
Feature: Session idle timeout (server-side enforcement + client watchdog)


larryro:
## Context

Compliance requires session termination after a defined period of inactivity (e.g. 30 min). Currently sessions have a 24h absolute expiry that is refreshed on activity in [services/platform/convex/betterAuth/trusted_headers/create_session_for_trusted_user.ts](services/platform/convex/betterAuth/trusted_headers/create_session_for_trusted_user.ts) — that's absolute timeout, not idle timeout.

## Proposed change

1. Add configurable `sessionIdleTimeoutMs` (default 30 min) on the platform security settings.
2. On each authenticated request, compare `now - session.updatedAt`; invalidate if exceeded. Better Auth supports this via `session.expiresIn` + `updateAge` — confirm usage in [services/platform/convex/auth.ts](services/platform/convex/auth.ts).
3. Add a client-side idle hook that signs out the current tab when idle for the configured duration (prevents stale tabs displaying sensitive data).

## Acceptance criteria

- [ ] After 30 min of inactivity, the next request returns 401 and the user is redirected to login.
- [ ] Client tab auto-signs-out after the configured idle duration.
- [ ] Setting is configurable by org admin in Settings → Security.

Part of the security compliance baseline.

yannickmonney:
Triaged — valid feature, keeping open. Confirmed there's currently no idle-timeout configuration in `convex/auth.ts` / `convex/betterAuth/auth.ts` (Better Auth defaults apply; the only bound is the absolute session expiry). This would add a configurable idle window with server-side enforcement plus a client tab watchdog. Assigned to me.

PR #1819, opened 2026-06-02
feat(platform): org-configurable session idle timeout (#1502)

yannickmonney:
## What

Adds a `session_idle_timeout` governance policy so org admins can configure the inactivity window from **Settings → Security**, closing the acceptance gap on #1502 (the merged work — #1808/#1810 — was deployment-env-var only).

Architecture (per maintainer decision: **per-org policy + env backstop**):

- `SESSION_IDLE_TIMEOUT_MINUTES` stays the **server-side** sliding-window control (unchanged Better Auth static config, covers all session types) and is the **hard cap** — an org may only *tighten* it, never extend past it.
- The per-org policy drives the **client idle watchdog**, which now resolves its window from `org policy × env backstop` instead of env alone.

## Changes

- `resolveEffectiveIdleMinutes()` — pure resolver combining env backstop × org policy; shared by the watchdog and Convex, unit-tested.
- `sessionIdleTimeoutConfigSchema` + `session_idle_timeout` policy type (shared + Convex schema kept in sync). `upsertPolicy` validates it and audit-logs the change via the existing path.
- Policy added to the member-readable allowlist so the watchdog can size its window.
- `SessionIdleTimeoutEditor` mounted in Settings → Security (mirrors `LoginPolicyEditor`), with loader preload.
- i18n en/de/fr.

## Tests

- Resolver: tighten / can't-loosen-past-cap / disabled / no-env-backstop.
- Watchdog: 4 new cases (per-org tighten, env cap, disabled fallback, org-only when no env).
- Editor: loaded / disabled / skeletonized loading.
- `bun run check`: format + lint + typecheck clean; full test suite green except 3 pre-existing concurrent-load timeouts in unrelated suites (pass 30/30 in isolation).

## Notes

Server-side enforcement intentionally stays on the env backstop (the trusted-headers session-creation path lacks cheap org context); per-org tightening is delivered client-side. True per-request per-org server enforcement was considered and declined to preserve the JWT fast path.

Part of the #1803 security/compliance epic.

## Release Notes

* **New Features**
  * Organization administrators can now configure session idle timeout policy in security governance settings to control automatic session expiration per organization, with ability to tighten timeouts while respecting system-level constraints.

* **Tests**
  * Added test coverage for session idle timeout configuration and session monitoring behavior.

* **Localization**
  * Added UI text for session idle timeout in English, French, and German.


yannickmonney:
CodeRabbit findings addressed in the follow-up commit:

- **Watchdog arming on the env window while the org policy loads** (major) — fixed: the effective-window memo now returns `null` while `organizationId` is set but the `getPolicy` query is unresolved (`policyRow === undefined`), so it never arms on the looser env window first. Added a regression test.
- **Persist normalized `session_idle_timeout` config** (major) — `upsertPolicy` now stores `parsed.data` (schema defaults applied) for this policy type.
- **Heading asserted against English literal** (i18n) — both heading checks now assert structurally (`getAllByRole('heading')`) instead of pinning English copy.
- **Test fixture broad cast** — fixture typed as `SessionIdleTimeoutConfig | undefined`.

`bun run check` green.

PR #1835, opened 2026-06-06
feat(platform): table layout, selection & menu polish

yannickmonney:
## Summary

A batch of UI polish centered on the shared data tables, plus the in-progress chat/notifications/threads work that was already in the working tree.

### Tables (shared `DataTable`)
- **Utility column alignment**: the select-checkbox and 3-dot row-actions columns are now pinned to a fixed width (1% column + fixed-size content box) so they land at the same x on every table; content columns absorb leftover space proportionally instead of one runaway column.
- **Skeletons** use the same width logic, so the loading state matches the real table.
- **Consistent row height** (`h-12` minimum).
- **Border on horizontal scroll**: the rounded border now wraps the full table width and scrolls with the content rather than being pinned to the viewport.
- **Bulk-delete bar** gets a consistent gap below the table in the sticky layout (`empty:hidden` so no phantom gap when hidden).

### Selection & menus
- **Agents** and **Skills** tables gain multi-select + bulk delete (protected agents excluded from selection).
- **Skills** row menu gains **Replace bundle** and **Duplicate**.
- Row-action menus **auto-insert a separator before destructive actions** (delete/archive), keeping consecutive destructive actions grouped.

### Automations
- **Folder-aware select-all** — selecting a folder (or select-all) now reaches the workflows nested inside folders.
- **Global, flat search** — a query matches workflows across all folders (same behavior at root or inside a folder) and each result is prefixed with its folder path, e.g. `github / GitHub Issue Sync`.

### Misc
- `projects` / `automations` pages get the standard page inset (gap restored).
- Tab navigation drops its focus ring.
- Fix untranslated `duplicate`/`rename`/`delete` menu labels and the select-checkbox aria labels.
- `bun run dev` no longer starts docs or web.

## Testing
- `tsc` typecheck + `oxlint` clean on changed files.
- Manually verified in the running app (Playwright): column alignment across tables, skeleton widths, row height, horizontal-scroll border, bulk-delete bars (agents/skills), folder select-all, global search with folder-path prefix, and the menu separators/labels.

## Release Notes

* **New Features**
  * Canvas state persistence for chats
  * Bulk deletion across tables (agents, automations, skills, teams, etc.)
  * Skill duplication support
  * Enhanced offline detection showing device vs. server connectivity status
  * Live relative timestamps in chat history
  * Double-click zoom toggle for image viewer

* **UI & Style**
  * Dark theme grayscale color refinements
  * Improved focus ring styling for keyboard navigation
  * Redesigned offline/connectivity overlay

* **Improvements**
  * Standardized table column sizing across platform
  * Per-row selection control in tables
  * Filter disable state support
  * Error boundary visual redesign

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "tools/plop"
   ],
   "scripts": {
-    "dev": "bunx turbo run dev",
+    "dev": "bunx turbo run dev --filter=!@tale/docs --filter=!@tale/web",
     "serve": "bunx turbo run serve",
     "build": "bunx turbo run build",
     "lint": "bunx turbo run lint",
--- a/packages/ui/src/components/primitives/icon-button.tsx
+++ b/packages/ui/src/components/primitives/icon-button.tsx
@@ -77,12 +77,11 @@ export const IconButton = forwardRef<
       />
     );
 
-    // The base Button uses `focus-visible:ring-ring`, which resolves to the
-    // theme's accent blue in dark mode (`--ring: 217 100% 66%`). On a small
-    // ghost icon button that ring reads as a heavy blue halo around an
-    // otherwise quiet glyph — visually loud, especially for icon-only
-    // toolbar clusters. Override to `border-strong` so focus stays visible
-    // for keyboard users (a11y) without the saturated blue.
+    // The base Button's `focus-visible:ring-ring` previously resolved to a
+    // saturated accent blue that, on a small ghost icon button, read as a
+    // heavy halo around an otherwise quiet glyph — visually loud, especially
+    // for icon-only toolbar clusters. Override to `border-strong` so focus
+    // stays visible for keyboard users (a11y) without the loud ring.
     const focusOverride = 'focus-visible:ring-border-strong';
 
     if (asChild && isValidElement(slotChild)) {
--- a/packages/ui/src/components/primitives/icon-button.tsx
+++ b/packages/ui/src/components/primitives/icon-button.tsx
@@ -77,6 +77,14 @@ export const IconButton = forwardRef<
       />
     );
 
+    // The base Button uses `focus-visible:ring-ring`, which resolves to the
+    // theme's accent blue in dark mode (`--ring: 217 100% 66%`). On a small
+    // ghost icon button that ring reads as a heavy blue halo around an
+    // otherwise quiet glyph — visually loud, especially for icon-only
+    // toolbar clusters. Override to `border-strong` so focus stays visible
+    // for keyboard users (a11y) without the saturated blue.
+    const focusOverride = 'focus-visible:ring-border-strong';
+
     if (asChild && isValidElement(slotChild)) {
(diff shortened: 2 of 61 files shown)


yannickmonney:
## CodeRabbit review — resolution summary

All CI checks are green. Findings addressed below.

### Fixed
- **projects / websites bulk delete** — use the existing `toId<T>()` helper instead of inline `as Id<…>` casts.
- **agents protected-name checks** — drop the `as readonly string[]` cast, use `.some()`.
- **skills bulk delete** — forward the SKILL.md CAS hash (`expectedHash`) so a concurrently-edited skill isn't deleted from a stale view.
- **thread-canvas optimistic update** — guard against a `null` `current` before patching.
- **date-range-picker** — disable the Reset button and early-return `handleClear` when `disabled`/`isLoading`.
- **data-table date filter** — translate the "Date filter unavailable" fallback (`common.dataTable.dateFilterUnavailable`, en/de/fr).
- **member / teams bulk delete** — drop the per-row toast so `BulkDeleteBar` surfaces a single batch-failure toast (no duplicates).
- **notification-bell tooltip** — gate the animation behind `motion-safe:` for `prefers-reduced-motion`.
- **chat** — share `THREADS_PAGE_SIZE` from the queries module (single source of truth).
- **model-selector** — render option descriptions as tooltips (`descriptionMode="tooltip"`).
- **data-table utility cell** — fallback width uses `ACTIONS_COLUMN_SIZE` for action columns.
- **icon-button** — refreshed a stale `--ring` comment.

### Kept intentionally (with rationale)
- **tab-navigation focus ring** — removed deliberately per an explicit product request in this change; the active-tab underline conveys state. (a11y trade-off acknowledged.)
- **entity-row-actions auto-separator** — by design: an explicit `separator` opens the trailing "dangerous" group so consecutive destructive actions (e.g. projects' *Archive* + *Delete*) stay together rather than getting a redundant second divider. Verified against the projects menu.
- **threads `setThreadCanvasState` access gate** — intentionally mirrors `setThreadPinned` / `markThreadRead`; `assertThreadAccess` is the cross-tenant guard, and canvas open / active-file is non-sensitive per-thread UI layout. Tightening only this mutation would diverge from the established pattern.
- **project-avatar cva refactor** — stylistic; current behavior is correct and it's outside this PR's scope.
- **use-relative-now `s`/`m`** — intentional compact ticker format; the ≥1h branch already localizes via `formatRelativeTime`.
- **chat `primeCachedPaginatedQuery` cast** — the bridging cast is isolated to the loader; widening the shared helper's generics affects every caller for no runtime benefit.

### Not reproduced
- **projects "no `getRowId`"** — `getRowId` is supplied by `useListPage` via `tableProps` (defaults to `row._id`), so selection keys are the document `_id`s.
- **api-keys `onDeleteComplete`** — `useRevokeApiKey` already invalidates the keys list on success, so the table refreshes.

PR #1841, opened 2026-06-10
feat(sandbox): docker + exec-free kubernetes dual execution backend

larryro:
## What

Make the sandbox service deployable on **both Docker Compose (today) and Kubernetes (Helm)**, selected by `SANDBOX_BACKEND=docker|kubernetes`. The Compose path is byte-identical to before; the K8s path is a native **Pod-per-exec** backend that needs no `docker.sock`, no host-path workspace, and no `--internal` egress bridge.

## Architecture

A coarse `ExecutionBackend.execute(cfg, req, opts) → ExecuteResponse` seam:

- **`DockerBackend`** — the Compose path, unchanged. Stages onto a host bind-mount, `docker run`, harvests back. The orchestration lives in `local-workspace-run.ts`; existing unit tests are the parity proof.
- **`KubernetesBackend`** — **exec-free, HTTP-only**. Every spawner↔Pod interaction is plain HTTP (`createNamespacedPod` / `readNamespacedPodLog` / `deleteNamespacedPod`) + presigned-URL I/O done **inside** the Pod.

### Why exec-free

Deep local e2e on kind found the K8s **exec websocket unreliable under Bun** (intermittent `Expected 101` upgrade failures; ~0% on a stressed cluster). `@kubernetes/client-node`'s `Log.log` / `readNamespacedPodLog` use a plain HTTPS `fetch` (reliable); only `Exec` uses `isomorphic-ws` (the failure point). So the transport uses **only** HTTP K8s ops — no websocket anywhere.

### Exec-free Pod (one Pod, shared `/workspace` emptyDir, 3 containers)

- **`stage` initContainer** (spawner image): downloads inputs from presigned URLs, writes the multi-step wrapper + prior-stage attestation. Completes before the runner — no sentinel. Required-input failure → `PRE_STAGE_FAILED`.
- **`runner`** (runtime image): command override runs `/entrypoint.sh` as a child of `sh -c` (not `exec`) so the wrapper captures the exit code to a file; stderr → a file (K8s logs stay clean stdout for phase parsing). **No credentials, no callbacks.** No runtime-image change.
- **`harvest`** sidecar (spawner image): owns the token + upload slots, enforces the user timeout, uploads `/workspace/output` via presigned slots + EP1/EP2, and prints one `__TALE_RESULT__` line the spawner reads back from its logs.

### Security

The per-exec **Secret** (presigned URLs + `SANDBOX_TOKEN` + caps) is mounted **only** into `stage`/`harvest`, **never** the `runner` — regression-tested. The runner has no token and no URLs.

### Horizontal scale

The result rides the `harvest` container's logs, read by the **owning** spawner replica — no callback to a Service VIP, no cross-replica affinity. The spawner stays HPA-able; cancel = delete-by-deterministic-name from any replica.

## Platform side

Unchanged. The spawner still emits the same SSE `phase`/`stdout`/`stderr`/`result` events; EP1/EP2 are backend-agnostic.

## Verification

- **Docker path:** `bun run check` green; staging/prior-output unit tests unchanged.
- **K8s path — unit:** Pod shape + the Secret-never-on-runner invariant, the ExecSpec Secret payload round-trip, and the `__TALE_RESULT__` result-line protocol. **156 tests pass**, typecheck/lint/format clean.
- **K8s path — on a real cluster (kind, runc):** drove the real `KubernetesBackend` through the actual `executeRequest` dispatcher.
  - **Reliability bar met:** 60/60, 50/50, +80 @ concurrency-4 — 100%, deterministic (~2.2–3.3 s each, zero flakes). No exec websocket → no flakiness.
  - basic (output uploaded, sha matched) / failure (exit 7 → `RUNTIME_ERROR`) / multistep (steps[] wrapper + output harvested).
  - **Cancel** mid-flight → `cancelled`, Pod + Secret cleaned up.
  - **Runtime security, verified on a live Pod:** the runner has no Secret mount / no `SANDBOX_TOKEN` / no presigned URLs; `stage` + `harvest` do.
  - **Multi-spawner scale:** two independent driver processes, disjoint Pods, no interference (no cross-replica state; cancel = delete-by-deterministic-name).
- **Bug found + fixed by the e2e:** the runner log-follow left a detached `AbortError` on cancel under Bun (client-node 1.4.0's `Log.log` pipes the body but never invokes the done-callback). Replaced the stream with **discrete `readNamespacedPodLog` polling**, so the whole k8s path is request/response HTTP with nothing to abort. Re-validated clean.
- **Prod image build fixed:** `services/sandbox/bun.lock` (the standalone lock the Dockerfile builder uses) was missing `@kubernetes/client-node` — regenerated so `--frozen-lockfile` passes.

The exec-free transport is validated on real k8s — the original "don't enable in production until tested" blocker is cleared. What remains before a production cutover is **ops/integration, not transport reliability**: a smoke against the **real platform** storage + EP1/EP2 in-cluster (the e2e used an in-cluster mock for those; the code that calls them is backend-agnostic and already production-proven on the docker path), plus the operator-applied **gVisor RuntimeClass** + egress **NetworkPolicy** (documented, not yet exercised here).

## Deploy contract

`services/sandbox/docs/kubernetes.md` documents the RBAC (namespaced Role; `pods` create/get/list/delete, `pods/log` get, `secrets` create/update/delete; **no `pods/exec`**), the required env (`SANDBOX_SPAWNER_IMAGE`, `NODE_EXTRA_CA_CERTS`, …), and the recommended egress NetworkPolicy. Helm manifests are authored separately by the deployment owner.


larryro:
Update: validated the backend end-to-end **locally** against a kind cluster with the **real `tale-sandbox-runtime` image** (real Python 3.12 execution, live PHASE markers, separated stderr, harvested `json.dump` output — 5/5 consistent runs). This closed two of the listed follow-ups and surfaced two real bugs, now fixed in cdaa57ec3:

- **`imagePullPolicy: IfNotPresent`** on runner + holder — K8s defaults `Always` for `:latest`, which re-pulls the image on *every* execution and ignores a locally-loaded image (kind/air-gapped). `IfNotPresent` matches the docker path (pull once, reuse).
- **Bun + @kubernetes/client-node abort resilience** — the aborted `Log.log` stream rejected with a detached `AbortError` under Bun and intermittently crashed the process; now handled at the source via the log `done`-callback, with pod-status reads + the log-follow connect wrapped in the same retry as the exec ops, plus an `unhandledRejection` backstop in `server.ts`.

Remaining follow-ups: log-follow mid-stream reconnect (only the connect is retried today), `cacheMode=pvc` (needs RWX storage), and the egress NetworkPolicy/RBAC contract doc (Phase 3). Local e2e is reproducible via token-auth + `NODE_EXTRA_CA_CERTS` against kind.

PR #1872, opened 2026-06-11
feat(sandbox): persistent agent sessions for Claude Code

larryro:
## What

Transforms the sandbox from **one-shot code execution** into a persistent
**"remote computer"** that can run coding agents (Claude Code, OpenCode) — the
base service for "give an agent a task in a sandbox" (e.g. GitHub issue →
Claude Code → PR). Entry-point UI (chat tool / workflow node) is intentionally
**out of scope** per the agreed plan; the service API is entry-agnostic so any
entry reuses it.

## How it fits together

- **`/v1/sessions` API** on the spawner (create/get/list/destroy + exec(SSE)/
  cancel/attach + env/files), proxied to **runnerd** — a small control daemon
  that is PID 1 of a long-lived session container (single-file `runnerd.mjs`,
  run by the image's Node 24). **No `kubectl exec`** anywhere; the K8s backend
  stays exec-free (Pod-IP HTTP). The registry is a cache; backend
  labels/annotations + runnerd's activity clock are the source of truth (boot
  re-adopt + TTL/idle sweep).
- **Single runtime image** extended (one-image decision): pinned Claude Code,
  OpenCode, `gh`, Playwright + Chromium + `@playwright/mcp`, git/rg/fd, plus an
  `agent` user (uid 10001, non-root so `bypassPermissions` is allowed) selected
  per profile. The one-shot path keeps uid 65534 and is byte-identical
  (docker-args snapshot unchanged).
- **`@tale/agent-adapters`** — entry-agnostic package: `buildExec(spec)` →
  session exec (argv/env/stdin), plus a parser turning each agent's native
  stdout (Claude `stream-json`, OpenCode JSONL) into one normalized `AgentEvent`
  union (usage deduped by message id).
- **LLM gateway = Bifrost** (OSS, Apache-2.0; pinned compose service): the only
  path from an in-sandbox agent to an LLM. Platform mints a session-scoped
  virtual key (budget + model allowlist), revoked on destroy. Raw provider keys
  never enter the sandbox.
- **Platform Convex glue**: `sandboxSessions` (polymorphic owner driving
  lifecycle/access/quota) + token/progress/credential-audit tables; reserve +
  watchdog mutations; `session_client` (HMAC platform→spawner), `bifrost_admin`
  (governance API), `session_credentials` (Tier-2 integration-credential broker
  → session env, audited), and `run_agent` (progress bridge: exec → adapter
  parser → throttled `sandboxSessionOps` so any entry renders live progress).
  `usageLedger` stays the single billing truth.

Tiered secret model is documented in `services/sandbox/docs/sessions.md`:
Tier-0 (platform secrets) never enter the sandbox; Tier-1 (LLM keys) proxied;
Tier-2 (integration creds) enter only via explicit per-session grant + broker +
audit + revoke-on-destroy.

## Tests

- Unit/integration (all green): spawner 225, runnerd daemon 20, agent_adapters
  12 — runnerd exec/timeout/cancel/attach + file ops, docker-session-args &
  k8s-session-pod-spec snapshots, the session route layer against a fake
  runnerd, adapter golden fixtures + `buildExec` snapshots.
- Platform `tsc` 0 errors; `convex codegen` bundles clean (new node-runtime
  modules carry `'use node'`).

**Needs live infra (follow-up, can't run in this env):** docker-build image
conformance (`tests/container-sandbox-runtime-test.sh`, syntax-checked), kind
e2e for the K8s session lifecycle, and a secret-gated agent smoke (real
Bifrost + provider key).

## Pre-PR checklist

- [x] Ran `bun run check`. All workspaces this PR touches (sandbox, platform,
  agent_adapters, runnerd daemon) pass. The only failures are a known
  turbo-load `mock.module` flake in `@tale/cli` (`pull-image.test.ts` — passes
  6/6 in isolation, confirmed) and `@tale/rag` interrupted by a local timeout;
  neither workspace is touched by this PR.
- [x] No hand-rolled skeletons / magic `h-[…]` — N/A (no UI in this PR).
- [x] Updated `services/platform/messages/{en,de,fr}.json` — N/A (no UI message
  keys added).
- [ ] Updated `/docs/{en,de,fr}/` — **follow-up.** This PR adds operator env
  vars + a compose service but no end-user-reachable surface (entry points
  deferred); operator docs for the Bifrost service + `SANDBOX_*`/`BIFROST_*`
  env land with the entry-point PR.
- [x] Docs opening/closing prose — N/A (no docs pages touched).
- [x] Ran docs lint/test — N/A.
- [x] Updated READMEs — N/A (no user-facing feature surface yet).

## Release Notes

* **New Features**
  * Added persistent sandbox sessions with long-lived environments and workspace state management.
  * Introduced agent adapter support for Claude Code and OpenCode workflows.
  * Implemented session credential provisioning and virtual key management.

* **Infrastructure Updates**
  * Integrated Bifrost LLM Gateway service for agent traffic routing.
  * Extended sandbox runtime with daemon mode for session lifecycle management.


larryro:
Worked through the CodeRabbit + CodeQL review against current `main`-rebased code in e44883b4a. Most findings predated ~40 commits of hardening, so each was re-verified before acting.

### Fixed
- **compose**: `bifrost` is the only external-image service → `pull_policy` defaults to `missing` instead of the repo-wide `build`.
- **agent_adapters**: `satisfies` on the buildExec fixture; deduped `str(message?.model)`; an OpenCode `tool_use` part with no `id`/`callID` now forwards as `raw` instead of normalizing under an empty `toolUseId` (+regression test).
- **runnerd daemon**: env seed now honours `patch()`'s entry/size caps; output byte-cap clips a cap-crossing chunk to the remaining budget instead of overshooting by a full pipe buffer (+regression test asserting exact-cap); dropped the `as RunnerdExecEvent` and `as readonly string[]` casts for real guards; URL fetches stream with a running byte cap instead of buffering the whole body; `POST /env` shape-validates before `envStore.patch`; logged the empty kill/stdin catches.
- **platform convex**: narrowed `runAgentInSession`'s `status` validator + result type to the 4 statuses it returns; dropped the redundant `as AgentSlug`; credential broker warns+skips on a single grant's decrypt failure instead of aborting the whole create; added the missing `Equal<>` session-profile parity guard; commented the intentional SSE `parseData` silence.
- **spawner**: `SANDBOX_AGENT_USER` validated as `uid:gid >= 1` at config load + chown guard (no silent root); k8s create shares one health deadline across `waitForEndpoint`/`waitForRunnerd`; sweep health-probe failures logged; Tier‑2 doc corrected to the env-backed v1 git-cred lifecycle.

### Already resolved on-branch
AbortController is wired to three cancellation paths; the dead `ResolvedCredential` interface is gone; the Dockerfile already fails hard when Chromium is missing (`fs.accessSync(..., X_OK)`).

### Deliberately not changed
- **CodeQL "unknown directive `'use node'`"** — required by Convex for Node-runtime modules; removing it breaks deploy. False positive.
- **CodeQL "useless conditional `!drainActive`"** — a re-check after an `await` where the drain can resolve mid-await and flip the flag; genuinely reachable, not dead code (CodeQL misses the cross-await mutation).
- **OpenCode prompt via argv** — `opencode run` only accepts the prompt as a positional `[message..]`; it has no stdin prompt source like Claude Code, so the stdin contract can't apply. Documented in the adapter.

All touched workspaces pass typecheck + lint + tests (agent_adapters 26, sandbox 245, daemon 49, platform sandbox suites green).

PR #1904, opened 2026-06-18
feat(cli): cross-terminal clean-output rework for tale + bun run dev

yannickmonney:
## Why

Both developer entry points — the `tale` CLI and `bun run dev` — leaked their machinery into the terminal: raw `stdio:'inherit'` docker/convex/vite firehose, ad-hoc `console.log` + emoji, turbo `<pkg>:<task>:` prefix interleave, and alt-screen rendering unsafe across terminals. This reworks both onto **one shared, cross-terminal output abstraction** so they look and behave identically: a calm step checklist with elapsed times, signal surfaced and noise collapsed, the raw firehose one `--verbose` flag away and always captured for clean failure dumps.

## What

**Shared core (`@tale/shared`)** — node-free where Convex's V8 bundler reaches it, enforced by an import-graph boundary test:
- `terminal/` — one capability probe (`NO_COLOR`/`FORCE_COLOR`/CI/`TERM=dumb`/win32 precedence), safe-ANSI palette/markers, wide-char width/truncate, `LiveRegion` (cursor-up + clear-line only, no alt-screen).
- `classify/` — five subprocess line classifiers (docker-compose / buildkit / convex / vite / platform-container) + a sticky multi-line-error wrapper.
- `process/` — `spawnCaptured` (never inherits), `pipeLines`, `RingBuffer`.
- `tux/` — the shared reporter: `runStep`, line emitters, one configured source honoring `--json` (silent) / `--quiet` / `--verbose`.
- `logging/logger` color de-duped onto the capability probe.

**`tale` CLI (`tools/cli`)** — global flags `-v/-q/-y/--no-color/--json/--ci` via one `preAction` `configureReporter`; `failWith` + stable exit codes (`2` usage / `3` precondition / `4` user-abort / `5` external-dep); `--json` for `status`, `config show`, `migrate status`; every command on a central `action()` dispatch; one prompt primitive that throws (never hangs) without a TTY.

**`bun run dev` (`services/platform/scripts`)** — split the 1.2k-line orchestrator into a tested **`dev-engine.ts` (`runDevFleet`)** + a thin shim (kept at `scripts/dev.ts` so turbo / playwright / the root supervisor keep spawning it), with the Convex restart/health machine, secret derivation, and mode/gate resolution as **pure unit-tested modules** (`convex-supervisor`, `dev-secrets`, `dev-modes`, `dev-gates`). Turbo bypass + single sandbox owner retained; all boot invariants preserved.

**No emoji anywhere** — the container entrypoints (`services/{platform,convex}/docker-entrypoint.sh`) now emit plain text severity words (`INFO`/`OK`/`WARN`/`ERROR`) so raw `docker logs` is clean too, and the classifier maps those to the bracketed glyph marker (`[ ✓ ]`/`[ ! ]`/`[ x ]`, ASCII fallback). Renamed `proc`→`process`, `term`→`terminal` (full words).

## Verification

- ~559 unit/integration + 16 compiled-binary smoke tests green; `turbo run lint typecheck` + `oxfmt --check` + `knip` clean; CLI binary builds (linux/macos/windows) with `check:bundle`.
- Docs structural suite green; CLI global-flags + exit-codes documented in **en/de/fr**.
- **Live** on a Docker + Convex stack: `bun run dev` boots clean to `READY` (step checklist, zero escape bytes when piped, `/health` 200); SIGTERM reaps all host children in ~1s with **zero orphans** and a graceful exit 0; backing containers persist by design.

## Notes

- `-v` now means `--verbose`; `--version` keeps `-V` (Commander's default), so this is additive, not breaking.
- Deliberately out of scope: effect-trace parity goldens for the dev engine (the live boot is the proof) and folding `sync-convex-env` into the DI pattern (it already works and is emoji-free).


yannickmonney:
Addressed the review findings in `24446604b` (verified each against current code; minimal changes; full local validation green — shared 206, CLI 181, platform scripts 25, typecheck/lint/oxfmt/knip clean).

**Fixed**
- `logger.boundary.test`: `existsSync` instead of read-and-discard; node-runtime guard now catches side-effect + dynamic `node:*` imports.
- `pipe-lines`: reject the promise on a Node stream `error` (was hanging the awaiting spawn).
- `ring-buffer`: backing store typed `T[]`, `tail()` no longer drops a stored `undefined` (+ regression test); migration-history wording removed.
- `spawn-captured`: `instanceof ReadableStream` narrowing instead of an unchecked cast.
- `ansi`: trimmed history commentary.
- `tux/context` + `live`: crash teardown via `uncaughtExceptionMonitor` (restores the cursor without swallowing the crash).
- `tux/lines`: table key column measured with `visibleWidth`.
- `index`: `--verbose` is long-only — a root `-v` shadowed `upgrade`'s `-v, --version` (`tale upgrade -v X` errored; fixed).
- `output-mode`: `--quiet` wins over `--verbose`. `start`: `--port` validated (1–65535). `deploy`: non-ENOENT lstat → `warn`. `fail`: `code` excludes `Ok`. `prompt.test`: real TTY flags stubbed off. `smoke`: asserts empty stderr. Test `ProcessEnv` casts dropped.

**Skipped (with reason)**
- `platform-container` warn emoji: stale — the classifier matches text `WARN` and emoji are stripped at source; the `⚠️` branch no longer exists.
- `live`/SIGINT–SIGTERM re-raise: would double-fire `bun run dev`'s graceful shutdown (verified). Applied only the safe `uncaughtExceptionMonitor` half.
- `context` `writeLine` stderr-when-region-active: intentional single-writer invariant; in non-interactive/`--json` there's no active region, so errors reach stderr.
- `dev-engine` best-effort catch: documented E2E-only `renice`; routing it through `warnLine` would spam every E2E boot.
- `dev.ts` `err: unknown`: removing the annotation widens the param to `any` (the annotation is correct).

PR #1913, opened 2026-06-21
fix(sandbox): add per-process ulimit parity with the docker backend (#1851)

### LINKED ISSUE #1851: sandbox/k8s: no per-process ulimit parity with the docker backend
From the 2026-06-10 review (deferred; documented as a known delta in docs/kubernetes.md).

The docker runtime enforces `--pids-limit=128`, `--ulimit fsize=100MB`, `--ulimit cpu=600`, `--ulimit core=0` (docker-args.ts:114-135). The k8s runner pod has cpu/memory limits and (since the fix PR) a workspace emptyDir sizeLimit, but no pids/fsize/cpu-time equivalents — a fork-heavy or single-giant-file workload behaves differently across backends (kubelet eviction / silent file-skip vs prompt RUNTIME_ERROR).

Options to evaluate: gVisor (runsc) resource controls, a tiny setrlimit shim in the runner wrapper (`sh -c 'ulimit -u 128 -f 204800; ...'` — busybox sh supports ulimit), or accepting the delta permanently.

Refs: services/sandbox/src/backend/kubernetes/k8s-pod-spec.ts (RUNNER_WRAPPER, resources), docs/kubernetes.md resource-parity note.

larryro:
## Summary

- Closes #1851
- Injects `ulimit -u 128 -f 204800 -t 600 -c 0` into `RUNNER_WRAPPER` in `k8s-pod-spec.ts`, giving the k8s runner the same per-process limits as the docker backend:
  - `-u 128` → pids (matches `--pids-limit=128`)
  - `-f 204800` → fsize 100 MiB in 512-byte blocks (matches `--ulimit fsize=104857600`)
  - `-t 600` → CPU time seconds (matches `--ulimit cpu=600`)
  - `-c 0` → core dump size (matches `--ulimit core=0:0`)
- Uses busybox sh's built-in `ulimit`, which only lowers limits — no extra capabilities required under the hardened `securityContext`
- Updates the stale workspace `emptyDir` comment (previously said fsize had "no per-process equivalent here")
- Adds a regression test asserting the `ulimit` line is present in `RUNNER_WRAPPER`

## Test plan

- [x] `bun test src/backend/kubernetes/k8s-pod-spec.test.ts` — all 20 tests pass (19 pre-existing + 1 new)
- [x] Full sandbox test suite: 301 pass, 3 pre-existing failures (missing `@kubernetes/client-node` dep + parse-error fixtures) unchanged

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive regression test to validate enforcement of process resource limits in sandbox runner environments

* **Bug Fixes**
  * Runner container now enforces explicit per-process resource constraints, including limits on memory usage, file size, execution time, and core dump generation, ensuring consistent behavior across deployment environments

### DIFF
--- a/services/sandbox/docs/kubernetes.md
+++ b/services/sandbox/docs/kubernetes.md
@@ -37,14 +37,16 @@ the run stops promptly, but the owning replica then sees its Pod vanish and
 resolves the run as `failed`/`HARVEST_READ_FAILED` rather than `cancelled` (it
 never learns the cancel intent — a known, accepted limitation).
 
-**Resource-limit parity (known delta vs docker):** the runner Pod enforces
-cpu/memory limits and a `sizeLimit` on the `/user` emptyDir
-(`SANDBOX_K8S_WORKSPACE_SIZE_LIMIT`, default `4Gi` — exceeding it evicts the
-Pod). There is **no Kubernetes equivalent** of the docker path's per-process
-ulimits (`pids-limit=128`, `fsize=100MB`, `cpu-time=600s`): a fork-heavy or
-single-giant-file workload behaves differently on the two backends. Operators
-needing those bounds should use gVisor (`SANDBOX_RUNTIME=runsc`) and
-namespace-level quotas.
+**Resource-limit parity with docker:** the runner Pod enforces cpu/memory
+limits and a `sizeLimit` on the `/user` emptyDir (`SANDBOX_K8S_WORKSPACE_SIZE_LIMIT`,
+default `4Gi` — exceeding it evicts the Pod). `RUNNER_WRAPPER` also injects
+`ulimit -u 128 -f 204800 -t 600 -c 0` before launching the entrypoint, which
+sets the same per-process limits as the Docker backend's `--pids-limit=128`,
+`--ulimit fsize=104857600`, `--ulimit cpu=600`, and `--ulimit core=0:0`. The
+busybox sh built-in lowers limits without requiring elevated capabilities, and
+all descendant processes (python/node) inherit them. gVisor (`SANDBOX_RUNTIME=runsc`)
+adds further isolation benefits (syscall filtering, separate kernel) but is no
+longer required solely to obtain these resource bounds.
 
 ## RBAC (namespaced Role — no cluster scope, no `pods/exec`)
 
--- a/services/sandbox/src/backend/kubernetes/k8s-pod-spec.test.ts
+++ b/services/sandbox/src/backend/kubernetes/k8s-pod-spec.test.ts
@@ -264,6 +264,14 @@ describe('buildSandboxPod', () => {
     }
   });
 
+  test('RUNNER_WRAPPER enforces per-process ulimits matching docker-args.ts', () => {
+    const c = runner(buildSandboxPod(cfg, goodInput));
+    const script = c.command?.[2] ?? '';
+    // Parity with --pids-limit=128, --ulimit fsize=104857600 (204800 blocks),
+    // --ulimit cpu=600, --ulimit core=0 from docker-args.ts.
+    expect(script).toContain('ulimit -u 128 -f 204800 -t 600 -c 0');
+  });
+
   test('RUNNER_WRAPPER invariant: a surviving wrapper always exits 0 (echo is last)', () => {
     // The runner-dead short-circuit in k8s-backend.ts depends on this: the
     // wrapper's LAST command is the echo into the exit-code file, so a
--- a/services/sandbox/src/backend/kubernetes/k8s-pod-spec.ts
+++ b/services/sandbox/src/backend/kubernetes/k8s-pod-spec.ts
@@ -96,6 +96,11 @@ export function podNameFor(executionId: string): string {
 // clean phase parsing). $0..$3 come from the Pod `args` trailer below.
 const RUNNER_WRAPPER = [
   `mkdir -p ${TALE_DIR}`,
+  // Per-process resource parity with docker-args.ts: --pids-limit=128,
+  // --ulimit fsize=104857600 (204800 × 512-byte blocks), --ulimit cpu=600,
+  // --ulimit core=0. busybox sh's built-in ulimit only lowers limits, so this
+  // never requires extra capabilities under the hardened securityContext.
+  `ulimit -u 128 -f 204800 -t 600 -c 0`,
   `/entrypoint.sh "$0" "$1" "$2" "$3" 2>${STDERR_PATH}`,
   `echo $? > ${EXIT_CODE_PATH}`,
 ].join('\n');
@@ -143,8 +148,8 @@ export function buildSandboxPod(
     // sizeLimit bounds everything the execution writes (deps installs, temp
     // files, outputs — the entrypoint points HOME/TMPDIR/PIP_TARGET here).
     // Exceeding it evicts the pod; the spawner's runner-dead path classifies
-    // that instead of burning the full timeout. K8s analogue of the docker
-    // path's fsize ulimit (which has no per-process equivalent here).
+    // that instead of burning the full timeout. Complements the per-process
+    // fsize ulimit in RUNNER_WRAPPER (total-workspace vs per-file cap).
     {
       name: 'workspace',
       emptyDir: { sizeLimit: cfg.k8s.workspaceSizeLimit },


larryro:
## Code Review: fix(sandbox/k8s): per-process ulimit parity (#1851)

**Verdict: LGTM — ready to merge.**

### CI
All non-skipped checks pass (Analyze, Browser, all Build jobs, Smoke test, Unit, Lint, Opengrep, Scan sandbox, Validate images, Migrations, UI). The skipped jobs are structural (fork-PR gates, Storybook, Trivy) and not relevant here.

### Value parity with docker-args.ts

| Docker flag | Converted | Shell flag | Correct? |
|---|---|---|---|
| `--pids-limit=128` | — | `ulimit -u 128` | ✓ |
| `--ulimit fsize=104857600` (bytes) | ÷ 512 → 204800 blocks | `ulimit -f 204800` | ✓ |
| `--ulimit cpu=600` | — | `ulimit -t 600` | ✓ |
| `--ulimit core=0:0` | — | `ulimit -c 0` | ✓ |

All four values map exactly. The fsize conversion (bytes → POSIX 512-byte blocks) is arithmetic-correct.

### Placement & correctness
The `ulimit` line is placed *before* `/entrypoint.sh` in the `sh -c` wrapper, so limits are set on the current shell and inherited by the entrypoint child. No additional Linux capabilities are needed — `ulimit` only lowers `RLIMIT_*` values, which any process can do unprivileged.

One minor note for the record: Docker's `--ulimit core=0:0` sets both soft **and** hard core limits to 0, while `ulimit -c 0` sets only the soft limit. In practice harmless — container runtimes already enforce a hard core limit of 0 — but worth noting.

### Tests
New test (`RUNNER_WRAPPER enforces per-process ulimits matching docker-args.ts`) correctly extracts `command[2]` from the built pod and asserts the exact ulimit string is present. Local run: **20/20 pass** in `k8s-pod-spec.test.ts`. The three other k8s test files fail due to `@kubernetes/client-node` not being installed in CI, but that failure is pre-existing on `main` and unrelated to this change.

### Comment update
The stale "has no per-process equivalent here" comment on the `emptyDir` sizeLimit is correctly replaced with "Complements the per-process fsize ulimit in RUNNER_WRAPPER (total-workspace vs per-file cap)."

larryro (commented):
## Review: fix(sandbox/k8s): add per-process ulimit parity with the docker backend (#1851)

**Verdict: READY TO MERGE**

---

### CI state

All 27 required checks pass (confirmed with `gh pr checks 1913`). No check is red or pending.

---

### Tests

Ran the project's own test suite locally:

```
bun test src/backend/kubernetes/k8s-pod-spec.test.ts
→ 20 pass / 0 fail  (includes the 1 new regression test)

bun test  (full sandbox suite)
→ 301 pass / 3 fail / 3 errors
```

The 3 failures and 3 errors are pre-existing and unrelated to this change:
- 2 `@kubernetes/client-node` module-not-found errors (missing optional dev dep in sandbox env)
- 1 parse-error fixture test in `k8s-protocol.test.ts`

All 20 tests in `k8s-pod-spec.test.ts` pass, including the new regression test.

---

### Correctness analysis

**Ulimit values match docker-args.ts exactly:**

| Docker flag | Value | k8s ulimit flag | Value | Verified |
|---|---|---|---|---|
| `--pids-limit=128` | 128 processes | `-u 128` | 128 | ✓ |
| `--ulimit fsize=104857600` | 100 MiB in bytes | `-f 204800` | 204800 × 512-byte blocks = 104857600 bytes | ✓ |
| `--ulimit cpu=600` | 600 CPU-seconds | `-t 600` | 600 | ✓ |
| `--ulimit core=0:0` | no core dumps | `-c 0` | 0 | ✓ |

**Placement is correct:** `ulimit` runs *before* `/entrypoint.sh`, so limits apply to the entrypoint and all its descendant processes (python/node).

**Security is sound:** `ulimit` in busybox sh only lowers limits and requires no elevated capabilities. The hardened `securityContext` (`allowPrivilegeEscalation: false`, `capabilities: { drop: ['ALL'] }`) is untouched and compatible.

**Comment update is accurate:** The updated `emptyDir` comment ("Complements the per-process fsize ulimit in RUNNER_WRAPPER — total-workspace vs per-file cap") correctly describes the relationship between the sizeLimit and the new ulimit.

**The `RUNNER_WRAPPER invariant` test still passes:** the `echo $? > EXIT_CODE_PATH` line remains last, preserving the runner-dead detection logic.

---

### Observation (non-blocking)

`docker-args.ts:130` includes `--ulimit nofile=1024:4096` (open-file-descriptor limit) which has no counterpart in the `ulimit` line here. However, issue #1851 explicitly identified the gaps as "pids/fsize/cpu-time" and the suggested approach in the issue body (`ulimit -u 128 -f 204800`) did not include `-n`. This is a pre-existing delta that was out-of-scope for this fix, not a regression.

---

### Summary

The implementation is minimal, correct, and fully addresses what issue #1851 asked for. All ulimit values are mathematically verified. Tests cover the new behaviour and all pre-existing tests continue to pass. CI is green.

larryro:
## Desk Review

**Verdict: NOT READY — one blocking doc update required before merge.**

### CI
All checks green. No pending or failed jobs.

### Tests
`bun test src/backend/kubernetes/k8s-pod-spec.test.ts`: **20 pass, 0 fail** (19 pre-existing + 1 new).
Full sandbox suite: **301 pass, 3 fail** — the 3 failures are pre-existing (missing `@kubernetes/client-node` dep in the k8s-client tests + parse-error fixtures) and unchanged by this PR.

### Code — correct

The math is right:
- `-u 128` → RLIMIT_NPROC 128, matching `--pids-limit=128`. Semantic note: RLIMIT_NPROC is a per-user limit (uid 65534) rather than a cgroup pids.max cap, so a process running under a different UID would not be counted — but the issue body itself suggested this exact ulimit-shim approach, and in this single-uid sandbox it provides the intended fork-bomb protection.
- `-f 204800` → RLIMIT_FSIZE 204800 × 512-byte blocks = 104,857,600 bytes, exactly matching `--ulimit fsize=104857600`. ✓
- `-t 600` → RLIMIT_CPU 600 s, matching `--ulimit cpu=600`. ✓
- `-c 0` → RLIMIT_CORE 0, matching `--ulimit core=0:0`. ✓

The `ulimit` line appears before `/entrypoint.sh` in RUNNER_WRAPPER, so all child processes inherit the limits. The busybox sh built-in lowers limits without requiring extra capabilities. The new test and the comment update on the workspace emptyDir are both correct.

---

### Blocking: stale documentation (`services/sandbox/docs/kubernetes.md`, lines 40–47)

The issue body explicitly says the known delta was **"documented as a known delta in docs/kubernetes.md"**. That section was not updated by this PR, and it still reads:

> **Resource-limit parity (known delta vs docker):** … There is **no Kubernetes equivalent** of the docker path's per-process ulimits (`pids-limit=128`, `fsize=100MB`, `cpu-time=600s`): a fork-heavy or single-giant-file workload behaves differently on the two backends. Operators needing those bounds should use gVisor (`SANDBOX_RUNTIME=runsc`) and namespace-level quotas.

This is now false. The fix exists. An operator reading the docs will be told no equivalent exists and that gVisor is required — directly contradicting what the code does. Update that paragraph to reflect that `RUNNER_WRAPPER` now enforces the per-process ulimits via the busybox sh built-in, and that gVisor is no longer required solely for this purpose (it still adds other isolation benefits).

---

### Required change

**`services/sandbox/docs/kubernetes.md` lines 40–47** — update the "Resource-limit parity" paragraph to remove the false claim that there is no k8s equivalent and document the `ulimit -u 128 -f 204800 -t 600 -c 0` enforcement in RUNNER_WRAPPER.

larryro (commented):
## Review: sandbox/k8s per-process ulimit parity (#1851)

### CI state
All checks green. 26 checks total; 6 skipping (fork-PR variants, Storybook, Docs/Web container tests, Trivy) — all expected to skip for a non-fork PR that doesn't touch docs. No failures.

---

### Implementation correctness

**Value mapping (vs docker-args.ts):**

| Flag | K8s ulimit | Docker arg | Notes |
|---|---|---|---|
| pids | `-u 128` | `--pids-limit=128` | 128 processes |
| fsize | `-f 204800` | `--ulimit fsize=104857600` | 204800 × 512 = 104,857,600 bytes ✓ |
| cpu | `-t 600` | `--ulimit cpu=600` | 600 s ✓ |
| core | `-c 0` | `--ulimit core=0:0` | 0 ✓ |

Conversions are exact. The `ulimit` call is placed BEFORE `/entrypoint.sh` in `RUNNER_WRAPPER`, so all descendant processes (python, node, uv, npm) inherit every limit correctly.

The runner container uses `command: ['/bin/sh', '-c', RUNNER_WRAPPER]` without `exec` in front of the entrypoint — sh remains alive, sets ulimits, then forks the entrypoint as a child — all child processes inherit the limits. ✓

---

### Findings

**1. Comment says "busybox sh" but the runtime image uses Debian `dash` — minor, non-blocking.**

The Dockerfile base is `python:3.12-slim-bookworm` (Debian). The Dockerfile even notes _"Debian slim ships only dash"_. Both `dash` and busybox ash support all four flags (`-u`, `-f`, `-t`, `-c`), so the command works correctly. The comment on lines 99–103 of `k8s-pod-spec.ts` ("busybox sh's built-in ulimit") is factually inaccurate about which shell is executing; it should reference Debian `dash`. Does not affect correctness.

**2. `ulimit -u` semantics differ from Docker's `--pids-limit` — known, acceptable delta.**

Docker's `--pids-limit` uses the cgroup v1/v2 pids controller: limit is **per-container**. `ulimit -u` sets `RLIMIT_NPROC`: limit is **per-UID on the host kernel** (no pod-level boundary on the default `runc` class). With default `maxConcurrent=4` and each pod using ~10–20 UID-65534 processes (sh + entrypoint + python/node + harvest container), the total stays well below 128. With gVisor (separate kernel per pod) this difference vanishes entirely. The issue itself proposed this exact mechanism and acknowledged the semantic delta — not a blocker, but operators running many concurrent executions on runc should be aware.

**3. No `set -e` in `RUNNER_WRAPPER` — if `ulimit` failed, the entrypoint would run without limits.**

A non-root process lowering its own soft limits never fails (`setrlimit` always succeeds when only lowering), so this is theoretical in practice. Pre-existing pattern in the file; not introduced by this PR. Low risk.

**4. `ulimit -c 0` sets only the soft limit in most POSIX shells — minor divergence from Docker.**

Docker's `--ulimit core=0:0` explicitly sets both soft and hard limit to 0. `ulimit -c 0` (without `-H`) in dash sets soft=0; the hard limit remains at the container default. If hard > 0, user code could raise the soft limit and write a core dump. Mitigated in practice by the fsize cap (100 MB per file), the 4 GiB emptyDir sizeLimit, and `readOnlyRootFilesystem: true`. Low practical risk. Could be tightened to separate `ulimit -Hc 0; ulimit -Sc 0` as a follow-up.

---

### Tests

New test in `k8s-pod-spec.test.ts`:

```ts
test('RUNNER_WRAPPER enforces per-process ulimits matching docker-args.ts', () => {
  const c = runner(buildSandboxPod(cfg, goodInput));
  const script = c.command?.[2] ?? '';
  expect(script).toContain('ulimit -u 128 -f 204800 -t 600 -c 0');
});
```

Exact-value regression gate — will catch any drift vs docker-args.ts values. CI **Unit** check confirms all 20 tests pass (19 pre-existing + this new one).

---

### Documentation

The stale note in `docs/kubernetes.md` ("no Kubernetes equivalent of the docker path's per-process ulimits") is correctly removed and replaced with an accurate description of the new behaviour. The comment on the `workspace` emptyDir ("Complements the per-process fsize ulimit in RUNNER_WRAPPER (total-workspace vs per-file cap)") is accurate and helpful.

---

### Verdict

**READY TO MERGE.**

All values are correct, tests pass, CI is green, and the documentation is consistent with the implementation. The four findings are non-blocking minor notes or pre-existing patterns; none is a regression introduced by this PR. The one item worth a follow-up (outside this PR's scope) is explicitly hard-zeroing the core limit (`ulimit -Hc 0`), but that is optional hardening, not a correctness gap.

larryro:
## Desk Review: fix(sandbox): add per-process ulimit parity with the docker backend (#1851)

**Verdict: READY TO MERGE**

### CI & Tests

CI: all checks green (26 jobs — all substantive checks pass; 5 fork/docs/storybook jobs that are expected to skip on non-fork PRs are skipped).

Tests run locally:
- `bun test src/backend/kubernetes/k8s-pod-spec.test.ts` → **20/20 pass** (19 pre-existing + 1 new)
- Full sandbox suite → **301/304 pass**, 3 pre-existing failures (missing `@kubernetes/client-node` dep ×2, parse-error fixtures ×1) — unchanged, unrelated to this change

---

### Correctness

The four numeric translations against `docker-args.ts:118–136` are all correct:

| Docker flag | ulimit flag | Arithmetic |
|---|---|---|
| `--pids-limit=128` | `-u 128` | RLIMIT_NPROC = 128 |
| `--ulimit fsize=104857600` | `-f 204800` | 104,857,600 ÷ 512 bytes/block = 204,800 ✓ |
| `--ulimit cpu=600` | `-t 600` | CPU seconds, 1:1 ✓ |
| `--ulimit core=0:0` | `-c 0` | both soft+hard = 0 ✓ |

**Placement** before `/entrypoint.sh` is correct — the entrypoint is a child process of the shell and inherits all rlimits. The `echo $? > exit-code` line correctly captures the entrypoint's exit code, not ulimit's.

**Hard-limit enforcement:** busybox sh `ulimit` sets both soft and hard limits to the given value when no `-S`/`-H` qualifier is used. Since the container runs as UID 65534 with `allowPrivilegeEscalation: false` and `capabilities: drop: ALL` (`CAP_SYS_RESOURCE` absent), user code cannot raise the hard limits — the ceiling is permanent for the sandbox lifetime.

**Mechanism note:** `ulimit -u` enforces RLIMIT_NPROC (per-UID process count) rather than Docker's cgroup `pids.max`. Functionally equivalent for this container — all processes run as UID 65534, no privilege escalation — and is the correct mechanism for shell-injected limits without elevated capabilities.

### Tests

The new test (`k8s-pod-spec.test.ts:267–273`) asserts the exact four-flag substring, pinning all four numeric values against drift. The static array literal for `RUNNER_WRAPPER` guarantees line order, so the `toContain` approach is sufficient for this static construction.

### Non-blocking observations

1. **`nofile` limit not mapped.** `docker-args.ts:129–130` also carries `--ulimit nofile=1024:4096`. The issue body explicitly named only `pids`, `fsize`, `cpu`, and `core` as the missing k8s limits; the updated docs paragraph correctly reflects that scope. This is a remaining delta but is intentionally out of scope for this fix.

2. **No `set -e` in wrapper.** If `ulimit` failed (e.g., node hard limits already below the requested values), execution would continue without limits applied. This is a pre-existing wrapper pattern, not introduced here. In practice, lowering rlimits from container-runtime defaults always succeeds; all four limits here are restrictions.

3. **gVisor RLIMIT_NPROC.** gVisor's RLIMIT_NPROC enforcement depends on the runsc version. The PR correctly frames gVisor as providing *additional* isolation rather than being *required* for these bounds. Operators should validate RLIMIT_NPROC enforcement on their specific gVisor build if a hard pids ceiling is a security requirement.

### Documentation

`docs/kubernetes.md` correctly replaces the "known delta vs docker" paragraph with an accurate description of the four mapped limits and the complementary `emptyDir sizeLimit` role. The comment update at `k8s-pod-spec.ts:152` ("total-workspace vs per-file cap") is accurate.

---

No code changes required. The fix is correct, complete for the stated scope, and all tests pass with green CI.

PR #1915, opened 2026-06-21
fix(sandbox): accumulate canonical stdout from polled deltas, not final re-read (#1850)

### LINKED ISSUE #1850: sandbox/k8s: stdout beyond kubelet log rotation returns a mid-stream window
From the 2026-06-10 review (deferred; the fix PR ships detection only).

The k8s backend reconstructs canonical stdout from `readNamespacedPodLog` (head-capped at stdoutMaxBytes, default 5MiB). Kubelet rotates container logs at containerLogMaxSize (default 10Mi) and the log API serves only the current file — so for runs emitting more than the rotation threshold, the final read returns a window from the middle of the stream instead of docker's deterministic first-5MiB head. The fix PR detects log shrinkage and sets `truncated.stdout = true`, but the window itself remains rotation-dependent.

Possible directions: accumulate canonical stdout spawner-side from the polled deltas (capped, like docker's drainAndCap) instead of re-reading the full log; or document head-of-current-file semantics as the contract.

Refs: services/sandbox/src/backend/kubernetes/k8s-backend.ts (pollRunnerStdout, final read + logShrunk flag).

larryro:
## Summary

Fixes #1850: `sandbox/k8s` stdout beyond kubelet log rotation returns a mid-stream window.

### Root cause

In `KubernetesBackend.execute()`, after the polling loop ends, the code performed a final `readPodLog()` call to get the canonical stdout. When kubelet had rotated the container log (due to output exceeding `containerLogMaxSize`, default 10Mi), this final read returned only the content of the **current rotated file** — a window from the **middle of the stream** rather than the deterministic first-N bytes that the docker backend's `drainAndCap` produces.

The previous fix PR detected log shrinkage and set `truncated.stdout = true`, but the content itself was still wrong.

### Fix

Accumulate canonical stdout **incrementally from polled deltas** in `pollRunnerStdout`, capped at `stdoutMaxBytes`, instead of re-reading the full log at the end:

- Added `canonicalChunks: Buffer[]` and `canonicalByteCount` accumulators inside the closure.
- Each delta (`logs.slice(lastLogLen)`) is appended to the accumulator up to the cap.
- On rotation detection (`logs.length < lastLogLen`): set `logShrunk = true` and **reset `lastLogLen = 0`** so subsequent polls pick up post-rotation content from byte 0 of the new file.
- Replaced the final `readPodLog()` call with `Buffer.concat(canonicalChunks).toString('utf8')`.

This gives deterministic first-N-bytes semantics regardless of when kubelet rotates the container log.

### Tests

Two regression tests added to `k8s-backend.test.ts`:
1. Pre-rotation head bytes are preserved when rotation is detected in the final poll — `truncated.stdout` is set.
2. Post-rotation deltas are appended correctly when the canonical cap has not yet been reached.

## Test plan

- [x] `bun test` in `services/sandbox` — 327 pass, 0 fail
- [x] `bunx tsc --noEmit` — no type errors
- [x] `bunx oxlint --type-aware` — 0 warnings, 0 errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runner stdout collection to reliably handle Kubernetes log rotation and truncation scenarios.
  * Enhanced incremental stdout accumulation with better tracking across container lifecycle events.

* **Tests**
  * Added comprehensive test coverage for stdout behavior during log rotation, truncation, and rotation recovery.

### DIFF
--- a/services/sandbox/src/backend/kubernetes/k8s-backend.test.ts
+++ b/services/sandbox/src/backend/kubernetes/k8s-backend.test.ts
@@ -409,7 +409,8 @@ describe('stdout accumulation across kubelet log rotation', () => {
         if (p?.container === 'harvest') {
           return Promise.resolve(harvestResultLine);
         }
-        // Runner log sequence: empty → pre-rotation → rotation (shorter, final poll)
+        // Runner log sequence: pre-rotation ('AAAAA') → rotation ('BBB',
+        // shorter, returned on the final poll after the loop exits).
         runnerLogCount += 1;
         if (runnerLogCount === 1) return Promise.resolve('AAAAA');
         // Final poll after loop: kubelet has rotated — shorter than 'AAAAA'.
@@ -462,6 +463,37 @@ describe('stdout accumulation across kubelet log rotation', () => {
     const stdout = Buffer.from(res.stdoutBase64, 'base64').toString('utf8');
     // Pre-rotation 'AAAAAA' + post-rotation delta 'BBBBBBBB' (full new file on reset).
     expect(stdout).toBe('AAAAAABBBBBBBB');
+    expect(res.truncated.stdout).toBe(true);
+  });
+
+  test('canonical accumulation is capped at stdoutMaxBytes; truncated.stdout is set', async () => {
+    // A tiny stdoutMaxBytes forces the cap branch: the runner emits more bytes
+    // than the cap, so only the first N are kept and truncated.stdout is set.
+    const cappedCfg: SpawnerConfig = { ...cfg, stdoutMaxBytes: 4 };
+    let readPodCount = 0;
+    const stub = stubClient({
+      readPod: () => {
+        readPodCount += 1;
+        // waitForRunnerStart sees runner running; main loop then breaks.
+        return Promise.resolve(
+          readPodCount <= 1 ? runningPod() : terminatedPod(),
+        );
+      },
+      readLog: (p) => {
+        if (p?.container === 'harvest') {
+          return Promise.resolve(harvestResultLine);
+        }
+        // 10 bytes of output, but the cap is 4.
+        return Promise.resolve('AAAAAAAAAA');
+      },
+    });
+    const backend = new KubernetesBackend(cappedCfg, stub);
+    const res = await backend.execute(cappedCfg, req, execOpts());
+
+    expect(res.status).toBe('completed');
+    const stdout = Buffer.from(res.stdoutBase64, 'base64').toString('utf8');
+    // Only the first 4 bytes are retained.
+    expect(stdout).toBe('AAAA');
     expect(res.truncated.stdout).toBe(true);
   });
 });
--- a/services/sandbox/src/backend/kubernetes/k8s-backend.test.ts
+++ b/services/sandbox/src/backend/kubernetes/k8s-backend.test.ts
@@ -22,6 +22,7 @@ import {
   staleLifetimeCutoffMs,
 } from './k8s-backend.ts';
 import { podNameFor } from './k8s-pod-spec.ts';
+import { formatResultLine } from './k8s-protocol.ts';
 
 const cfg: SpawnerConfig = {
   backend: 'kubernetes',
@@ -227,7 +228,7 @@ function stubClient(behavior: {
   createSecret?: () => Promise<unknown>;
   createPod?: () => Promise<unknown>;
   readPod?: () => Promise<V1Pod>;
-  readLog?: () => Promise<string>;
+  readLog?: (params?: { container?: string }) => Promise<string>;
 }): { core: CoreV1Api; namespace: string; calls: StubCalls } {
   const calls: StubCalls = { deletes: [], replaces: 0 };
   // oxlint-disable-next-line typescript-eslint/no-unsafe-type-assertion -- test stub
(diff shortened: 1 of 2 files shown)


larryro:
## Desk Review: fix(sandbox) — accumulate canonical stdout from polled deltas (#1850)

**Verdict: READY TO MERGE**

---

### CI Status — ALL REQUIRED CHECKS PASS

| Check | Result |
|---|---|
| Analyze (javascript-typescript) | ✅ pass |
| Browser | ✅ pass |
| Build controller / convex / db / sandbox / proxy / sandbox-runtime / sandbox-egress / platform | ✅ pass |
| Smoke test | ✅ pass |
| Scan sandbox | ✅ pass |
| Validate images | ✅ pass |
| Lint commits | ✅ pass |
| Opengrep / Opengrep OSS | ✅ pass |
| Unit / UI / Migrations | ✅ pass |
| CodeRabbit | ✅ pass |
| Smoke test (fork PR) / Docs container / Storybook / Trivy / Web container | skipping (expected for this PR type) |

---

### Test Run

```
bun test services/sandbox/src/
327 pass, 0 fail — 22 files [3.29s]
```

The two new tests in `k8s-backend.test.ts` (`'pre-rotation head bytes are preserved; truncated.stdout is set'` and `'post-rotation deltas are appended when cap has not been reached'`) both pass and correctly pin the fix's two key scenarios.

---

### Correctness Review

**Root cause correctly targeted.** The old code's final `readPodLog(limitBytes: stdoutMaxBytes)` after the poll loop returned the rotated (mid-stream) file when kubelet had rotated the container log between the last poll and the final read — producing a window of a different position rather than the canonical head. The fix eliminates that re-read entirely and builds the canonical buffer from incremental polled deltas.

**Happy path (no rotation):** Deltas are computed correctly as `logs.slice(lastLogLen)` where `lastLogLen` tracks string-character positions consistently. Accumulation is correctly capped at `stdoutMaxBytes` via the `remaining` guard.

**Rotation path:** When `logs.length < lastLogLen` is detected (log shrunk), `lastLogLen = 0` is now set (previously it was left at the old value and subsequent polls computed a negative or zero delta). This correctly causes the next poll to treat the full new file content as a fresh delta. Pre-rotation bytes in `canonicalChunks` are preserved. `logShrunk = true` is a one-way latch, correct even for multiple sequential rotations.

**Cap path:** Once `canonicalByteCount >= stdoutMaxBytes`, the `remaining <= 0` guard prevents further accumulation. The `limitBytes` API parameter also caps the raw log read from the START of the log (confirmed in `k8s-client.ts`: "caps the response from the START (matches our 'truncate the tail' stdout policy)"), so the two caps are complementary.

**Truncation flag:** `canonicalByteCount >= cfg.stdoutMaxBytes || logShrunk` is correct — rotation always sets the flag because bytes between the last successful old-file poll and the rotation point may be unobserved.

**`limitBytes` semantics confirmed:** Delta computation (`logs.slice(lastLogLen)`) only makes sense if the API returns the FIRST N bytes (not a tail window). The `k8s-client.ts` comment explicitly confirms this.

---

### Minor Observations (non-blocking)

1. **Double allocation:** `Buffer.from(delta, 'utf8')` is created twice per delta — once for `scanner.onStdoutChunk?.(...)` and once for accumulation as `deltaBuf`. Trivially fixable, but both buffers are bounded by `stdoutMaxBytes` and the overhead is negligible.

2. **Slightly misleading test comment:** In the first new test, "Runner log sequence: empty → pre-rotation → rotation" doesn't match the actual call sequence — call 1 returns `'AAAAA'` (pre-rotation), not an empty probe. The behavior tested is correct; only the comment is off.

3. **No dedicated test for the `canonicalByteCount >= stdoutMaxBytes` branch:** The cap logic is simple and correct; with `stdoutMaxBytes = 5 MiB` in the test config, small test data never exercises this branch. Not blocking.

---

### Summary

The fix is correct on all paths (happy, rotation, cap, multiple rotation). The new tests pin both primary scenarios. All 327 sandbox tests pass locally and all required CI checks are green. No blocking findings.

larryro:
## Desk Review — PR #1915 (fixes #1850)

**Verdict: READY TO MERGE.**

Reviewed the full implementation (not just the diff) over two rounds — 6 parallel breadth subagents across correctness / robustness / elegance / tests / boundaries, then independent confirmation of every candidate blocking finding.

### CI & tests
- **CI: green.** Every required check passes (Build*, Unit, UI, Smoke test, Validate images, Scan sandbox, Analyze, Opengrep, Lint commits, Migrations). The `skipping` entries are fork-PR-only conditional jobs that don't apply here.
- **Tests run locally:** `bun install && bun test` in `services/sandbox` → **328 pass, 0 fail**. `bun test src/backend/kubernetes/k8s-backend.test.ts` → **16 pass, 0 fail**.

### What the fix does, and why it's correct
Replaces the post-loop `readPodLog()` re-read with incremental accumulation of polled deltas into `canonicalChunks` (capped at `stdoutMaxBytes`). On kubelet rotation (`logs.length < lastLogLen`) it sets `logShrunk=true` and resets `lastLogLen=0` so subsequent polls pick up the post-rotation file from byte 0. This yields deterministic first-N-bytes semantics (docker `drainAndCap` parity) instead of the mid-stream window the old final re-read returned. Verified:
- Non-rotation happy path accumulates contiguous first-N bytes; no double-counting on the reset (traced `AAAAAABBBBBBBB`).
- Cap enforcement: once `canonicalByteCount` hits `stdoutMaxBytes`, `remaining` is 0 and nothing further is appended; `truncated.stdout` set via `canonicalByteCount >= cap || logShrunk`.
- Rotation detected on the final poll forfeits that poll's post-rotation window but flags `truncated.stdout=true` — by design, consistent with the issue's deterministic-head goal.

### Findings examined and cleared (non-blocking)
1. **UTF-8 split at the cap boundary** — `deltaBuf.subarray(0, remaining)` can cut a multibyte char, but the truncated chunk is always the **last** chunk, so any `U+FFFD` is **trailing, not mid-stream**. The old code's `readPodLog(..., {limitBytes})` kubelet cut had the same trailing-`�` possibility through the identical `assemble`→`capText` path. **Not a regression.** Optional polish: route the cap cut through `capText` to trim a trailing `�`.
2. **Empty stdout on total poll failure** — identical to old behavior (old final read also swallowed errors → `''`). Acceptable.
3. **Inter-poll / multi-rotation data loss** — inherent to polling a rotating log; always flagged via `truncated.stdout`.
4. **Tests** — rotation, post-rotation append, and cap are directly covered and fail against the old code. Optional additions: a multi-poll non-rotation accumulation assertion and a multibyte-at-cap case.

No blocking issues. The change is minimal, well-named, well-commented, leaves no dead code, and resolves the reported issue.

**READY TO MERGE**

PR #1917, opened 2026-06-22
fix(sandbox/k8s): accumulate canonical stdout from polled deltas to avoid mid-stream window after log rotation

### LINKED ISSUE #1850: sandbox/k8s: stdout beyond kubelet log rotation returns a mid-stream window

larryro:
## Summary

Fixes #1850.

When kubelet rotates a container log (at `containerLogMaxSize`, default 10Mi), the Kubernetes log API serves only the current/new file. The previous implementation did a final `readPodLog` call to get the canonical stdout — after rotation, this returned content from the middle of the stream (the new file's head), not the deterministic first-`stdoutMaxBytes` bytes.

**Root cause:** `pollRunnerStdout` accumulated deltas into the `scanner` for live progress callbacks, but the canonical `stdout` at the end was taken from a fresh full `readPodLog` call that bypassed the spawner-side accumulation.

**Fix:** Accumulate the canonical stdout spawner-side into `logBuf` from the same polled deltas, capped at `stdoutMaxBytes` bytes (using `capText`). After rotation (`logs.length < lastLogLen`), `logBuf` already holds the pre-rotation head; we reset `lastLogLen = 0` to continue accumulating from the new file within the remaining cap. The final `readPodLog` call for canonical stdout is removed — `logBuf` is used directly.

This is the "accumulate canonical stdout spawner-side from the polled deltas" direction described in the issue, analogous to docker's `drainAndCap` approach.

## Changes

- `services/sandbox/src/backend/kubernetes/k8s-backend.ts`: Add `logBuf` accumulation in `pollRunnerStdout`; reset `lastLogLen = 0` on rotation; replace the final `readPodLog` call with `const stdout = logBuf`.
- `services/sandbox/src/backend/kubernetes/k8s-backend.test.ts`: Add `stdout log-rotation pinning` test that simulates log shrinkage and verifies the canonical stdout is the pre-rotation head (not the rotated window).

## Test plan

- [x] `bun run --filter @tale/sandbox test` — 301 pass, 3 pre-existing failures (all due to `@kubernetes/client-node` not installed in the test environment, unrelated to this change)
- [x] New unit test `stdout log-rotation pinning` added to `k8s-backend.test.ts` covering the rotation scenario
- [x] Logic verified: with the bug, final stdout would be the rotated file content; with the fix it is the pre-rotation head

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive test coverage for stdout handling during Kubernetes pod log rotation events.

* **Bug Fixes**
  * Fixed incomplete stdout logs during Kubernetes log rotation. Stdout is now reliably preserved across log rotation boundaries using a persistent buffer mechanism, ensuring complete execution logs.

### DIFF
--- a/services/sandbox/src/backend/kubernetes/k8s-backend.test.ts
+++ b/services/sandbox/src/backend/kubernetes/k8s-backend.test.ts
@@ -22,6 +22,7 @@ import {
   staleLifetimeCutoffMs,
 } from './k8s-backend.ts';
 import { podNameFor } from './k8s-pod-spec.ts';
+import { formatResultLine } from './k8s-protocol.ts';
 
 const cfg: SpawnerConfig = {
   backend: 'kubernetes',
@@ -227,7 +228,7 @@ function stubClient(behavior: {
   createSecret?: () => Promise<unknown>;
   createPod?: () => Promise<unknown>;
   readPod?: () => Promise<V1Pod>;
-  readLog?: () => Promise<string>;
+  readLog?: (p: { container: string }) => Promise<string>;
 }): { core: CoreV1Api; namespace: string; calls: StubCalls } {
   const calls: StubCalls = { deletes: [], replaces: 0 };
   // oxlint-disable-next-line typescript-eslint/no-unsafe-type-assertion -- test stub
@@ -238,7 +239,8 @@ function stubClient(behavior: {
     readNamespacedPod:
       behavior.readPod ??
       (() => Promise.reject(Object.assign(new Error('404'), { code: 404 }))),
-    readNamespacedPodLog: behavior.readLog ?? (() => Promise.resolve('')),
+    readNamespacedPodLog: (p: { container: string }) =>
+      behavior.readLog?.(p) ?? Promise.resolve(''),
     replaceNamespacedSecret: () => {
       calls.replaces += 1;
       return Promise.resolve({});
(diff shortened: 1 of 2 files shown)


larryro:
## Desk Review — PR #1917

**VERDICT: READY TO MERGE**

CI: all required checks green (Analyze, Browser, Build ×8, Opengrep, Smoke test, Unit, Lint commits, Validate images, Migrations, UI, Scan sandbox — all pass; 5 jobs correctly skip on fork/non-applicable). No red or pending checks.

Tests: `bun run --filter @tale/sandbox test` — 301 pass, 3 fail. The 3 failures are pre-existing and unrelated to this change (the three `@kubernetes/client-node` tests that fail because the module is absent from the sandbox CI test environment; all 3 identical failures exist on `main`). The CI Unit check is green in the proper environment.

---

### What the fix does

`pollRunnerStdout` now accumulates an incremental `logBuf` from each polled delta, capped at `stdoutMaxBytes` via `capText`. On a kubelet log rotation (`logs.length < lastLogLen`), `logBuf` already holds the deterministic pre-rotation head; the code resets `lastLogLen = 0` so the new file's content continues accumulating into the remaining cap. The final `readPodLog('runner')` call is replaced with `const stdout = logBuf`.

The PR **also fixes a secondary bug** in the old code that went undocumented in the issue: the old `else if` branch set `logShrunk = true` but did **not** reset `lastLogLen`. On every subsequent poll after rotation the new file starts from zero and grows slowly — always shorter than the pre-rotation `lastLogLen` — so `logs.length < lastLogLen` stays true and `logs.length > lastLogLen` can never fire. `scanner.onStdoutChunk` would go permanently silent and no new post-rotation content would reach live-progress callbacks. The `lastLogLen = 0` reset at line 466 closes this gap.

---

### Findings by dimension

**Correctness — no new bugs.** Walked all branches:

- Happy path: delta accumulates correctly into `logBuf`.
- Rotation: `logShrunk = true` + `lastLogLen = 0` → correct. Pre-rotation head preserved; new file accumulates within remaining cap.
- Multi-rotation: each shrink resets cursor; invariant holds.
- Error path: poll failure leaves `logBuf`/`lastLogLen` unchanged; next poll catches up.
- `stdoutStreamTruncated = Buffer.byteLength(stdout) >= stdoutMaxBytes || logShrunk` propagates correctly through all four return paths (aborted, runnerDead, harvestDone, harvestMissing).
- `assemble()` re-applies `capText(stripPhaseMarkers(stripControlChars(logBuf)), stdoutMaxBytes)`; since `logBuf` is already capped and stripping can only shorten it, `stdoutCapTrunc` is always `false` — no double-truncation issue.

Two **pre-existing issues** (not introduced by this PR):
- `lastLogLen` tracks `.length` (character count) while `limitBytes` is a byte budget — incommensurable for multi-byte UTF-8. After rotation, `lastLogLen = 0` makes the mismatch slightly more load-bearing, but the practical impact is narrow (only matters when the post-rotation file immediately exceeds the byte cap between polls with multi-byte content) and a fix is out of scope here.
- Rotation is silently undetected when the new file happens to be ≥ old length at the moment of polling. Pre-existing limitation of length-based detection; not worsened by this PR.

**Tests — one gap, not blocking.** The new `stdout log-rotation pinning` test correctly pins the core scenario; the `res.stdoutBase64` / `res.truncated.stdout` assertions are sound. The `stubClient` signature change (`readLog` now receives `{ container }`) is backward-compatible with all existing callers.

Minor coverage gap: the test is structured so rotation is always detected in the **final** `pollRunnerStdout()` after the loop, never during an in-loop poll. The post-rotation accumulation branch (`lastLogLen = 0` → delta from new file flows into `logBuf`) is never exercised by the test. The code is correct by inspection, and the existing 301-test suite covers no-rotation accumulation, so this is not blocking — worth a follow-up test if the rotation scenario ever sees regressions.

**Elegance — cosmetic only.** `logBuf` is declared two lines below its three sibling poll-state variables, separated by its comment block. Convention in the file is to group related `let` declarations first. The `Buffer.byteLength(logBuf, 'utf8') < cfg.stdoutMaxBytes` guard allocates a `Buffer` per poll; capturing `capText`'s returned `truncated` boolean as a `logBufFull` sentinel would avoid the redundant allocation. Neither item is blocking at 500 ms poll intervals.

**Issue resolution — complete.** The fix matches the "accumulate canonical stdout spawner-side" direction in #1850. All `readPodLog('runner')` call sites verified: the only remaining one is the poll inside `pollRunnerStdout` itself (correct, it must stay). Consistent with the Docker backend's `drainAndCap` approach.

---

The implementation is correct, the secondary live-tail-dark regression is a genuine bonus fix, and CI is fully green. Ready to merge.

larryro:
Superseded by #1915, which fixes the same issue (#1850) with the same approach but a more comprehensive test suite (covers both pre-rotation preservation **and** post-rotation delta append) and cleaner accumulation logic (`canonicalChunks`/`canonicalByteCount` with an explicit byte limit).

Closing this to avoid merging the same fix twice. If you prefer this implementation, reopen and we'll close #1915 instead.

PR #1919, opened 2026-06-22
fix(sandbox): verify skipTLSVerify/caFile behavior of kubeconfig under Bun (#1849)

### LINKED ISSUE #1849: sandbox/k8s: verify skipTLSVerify/caFile behavior of kubeconfig under Bun

larryro:
## Summary

Resolves #1849.

`@kubernetes/client-node@1.4.0` routes every API call through **node-fetch v2** (`gen/http/isomorphic-fetch.js: import fetch from "node-fetch"`), which calls `https.request()` with an `https.Agent` — **not** Bun's native `fetch()`. Bun's `node:https` implementation honours `rejectUnauthorized` and `ca` on the Agent, so `skipTLSVerify` and `caFile` in the kubeconfig **are** effective and carry real security semantics — they are not dead code.

The limitation documented in the AUTH NOTE is specific to **client-cert auth** (`cert`/`key` options on the Agent), which Bun's TLS stack does not support. `skipTLSVerify` and `caFile` are distinct knobs that work as intended.

## Changes

- **`k8s-client.test.ts`** — three new tests under `kubeconfig TLS knobs under Bun` that call `applySecurityAuthentication` against a test `RequestContext` and assert the `https.Agent` options (`rejectUnauthorized`, `ca`) are set correctly for `skipTLSVerify:true`, `caFile`, and the no-override case.
- **`k8s-client.ts`** — AUTH NOTE rewritten to be precise: client certs are inert; `skipTLSVerify`/`caFile` are effective via `node:https`; `NODE_EXTRA_CA_CERTS` serves as belt-and-suspenders for native `fetch()` paths (harvest presigned-URL uploads).
- **`docs/kubernetes.md`** — new **TLS contract under Bun** section with a table of kubeconfig knobs, their Bun-effective status, and guidance on `SANDBOX_K8S_CAFILE` vs `NODE_EXTRA_CA_CERTS`.

## Test plan

- [x] `bun run --filter '@tale/sandbox' test` — 345 pass, 0 fail (3 new tests added)
- [x] `bun run --filter '@tale/sandbox' lint` — 0 warnings, 0 errors
- [x] `bun run --filter '@tale/sandbox' typecheck` — clean
- [x] `bun run --filter '@tale/sandbox' format:check` — clean
- [x] `bun run lint:sast -- --include services/sandbox/` — 0 findings

## Summary by CodeRabbit

* **Documentation**
  * Expanded Kubernetes sandbox documentation with detailed TLS behavior specification, including which configuration options are supported and practical consequences of various authentication setups.

* **Tests**
  * Added comprehensive test coverage for TLS configuration option handling in Kubernetes client implementations.


larryro:
## Code Review: fix(sandbox): verify skipTLSVerify/caFile behavior of kubeconfig under Bun (#1849)

**CI:** All required checks passed (Unit ✅, UI ✅, Browser ✅, Migrations ✅, Smoke test ✅, Build sandbox ✅, Opengrep ✅, CodeRabbit ✅, Analyze JS/TS ✅). Fork-PR skips are expected.

**Tests run locally:** `bun test services/sandbox/src/backend/kubernetes/k8s-client.test.ts` — **11 pass, 0 fail**.

---

## Verdict: NOT READY — 1 blocking issue

---

## Blocking

### 1. `services/sandbox/docs/kubernetes.md:92` — Internal contradiction: env table says the opposite of the new TLS section

The `NODE_EXTRA_CA_CERTS` row in the Environment table was not updated and still reads:

> **Bun's fetch ignores the kubeconfig CA**, so without this the apiserver TLS isn't trusted.

But the PR's own new "TLS contract under Bun" section (same file, ~35 lines below) now correctly states that the kubeconfig CA *is* honoured for kubernetes API calls — through node-fetch's `https.Agent`. A reader who skims the env table first walks away with the old, wrong understanding — directly undermining the documentation fix this PR is supposed to deliver.

The reason `NODE_EXTRA_CA_CERTS` is still needed is different now: native `fetch()` calls — like harvest presigned-URL uploads in `exec-common.ts` / `sandbox-callback.ts` — bypass the Agent and go through Bun's native TLS stack, which reads `NODE_EXTRA_CA_CERTS` but ignores `agent.options.ca`.

**Fix for `kubernetes.md:92`:**
```
| `NODE_EXTRA_CA_CERTS` | in-cluster | Point at the SA `ca.crt` (`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`). Kubernetes API calls are trusted via `caFile` → Agent (see TLS note below), but native `fetch()` calls (e.g. harvest presigned-URL uploads) bypass the Agent and require this variable to trust the apiserver CA. |
```

---

## Non-blocking (worth fixing but should not hold the merge)

### 2. `k8s-client.test.ts:189` — `unlinkSync` in `finally` will throw `ENOENT` if a concurrent runner already deleted the file

Use `rmSync(caPath, { force: true })` instead of `unlinkSync(caPath)`.

### 3. `k8s-client.test.ts:186–187` — `ca` buffer content not verified; stale `/tmp` file could cause false-pass

The test confirms `ca` is a Buffer but not that it contains the expected bytes. A pre-existing `/tmp/k8s-client-test-ca.pem` from a prior failed run would cause the test to pass even if `caFile` plumbing was broken. Add:

```ts
expect(agentOpts?.ca).toEqual(Buffer.from('placeholder-pem-content'));
```

### 4. `k8s-client.ts:92–94` — `skipTLSVerify: true` fallback is silent in logs

When `SANDBOX_K8S_SERVER`+`SANDBOX_K8S_TOKEN` are set but `SANDBOX_K8S_CAFILE` is absent, TLS verification is silently disabled. An operator who misconfigures production gets no signal in the live log. A `console.warn('[sandbox.k8s] SANDBOX_K8S_CAFILE not set — TLS verification DISABLED ...')` before `kc.loadFromOptions` would surface this immediately.

### 5. `k8s-client.test.ts` — `caData` path documented as honoured but not tested

`kubernetes.md:134` lists `caFile / caData` together in the "Yes — honoured" row. The `caData` path (base64-encoded inline CA) takes the `bufferFromFileOrString(null, caData)` branch rather than the `bufferFromFileOrString(caFile, null)` branch — a distinct code path. A parallel test to the existing `caFile` test would give this claim the same evidential weight.

---

## Observations (no action needed)

- **Tests (a) and (b) are sound:** Assertions on specific non-undefined values (`toBe(false)`, `.toBeDefined()`) would fail if `agent.options` were absent or plumbing broken — not vacuous.
- **Test (c) mild vacuous-pass risk:** `.toBeUndefined()` assertions pass if `getAgentOpts` returns `undefined`. Library source confirms `context.setAgent(...)` is unconditional so this is not a present risk, but an `expect(agentOpts).toBeDefined()` guard would make it explicit.
- **AUTH NOTE in k8s-client.ts:** The rewritten comment is accurate and an improvement over the old text.
- **`agent.options` cast at line 156:** The `oxlint-disable` is correct. Tests confirm this holds under the current Bun version.

larryro:
## Desk Reviewer Report — PR #1919

**Branch:** `tale/wx76d456sa63vjy2ath8g3pspn895vya`  
**Issue:** #1849 — sandbox/k8s: verify skipTLSVerify/caFile behavior of kubeconfig under Bun

---

### CI State

All applicable CI checks **pass**. The "skipping" jobs (Docs container test, Smoke test (fork PR), Storybook, Validate images (fork PR), Web container test) are fork-PR-specific jobs skipped by design — not failures.

| Check | Result |
|---|---|
| Unit | ✅ pass |
| Browser | ✅ pass |
| UI | ✅ pass |
| Build (all targets) | ✅ pass |
| Migrations | ✅ pass |
| Analyze (JS/TS + Python) | ✅ pass |
| Opengrep / Opengrep OSS | ✅ pass |
| Trivy | ✅ pass |
| Scan sandbox | ✅ pass |
| Smoke test | ✅ pass |
| Lint commits | ✅ pass |

---

### Test Run (local)

After installing dependencies:

```
bun run --filter '@tale/sandbox' test
346 pass, 0 fail, 1052 expect() calls — Ran 346 tests across 22 files.
```

The three new TLS-knob tests pass.

---

### Findings

#### BLOCKING — must be fixed before merge

**1. `k8s-client.test.ts:207–218` — `caData` test does not verify decoded content**

The `caFile` test (line 188) correctly asserts:
```ts
expect(agentOpts?.ca).toEqual(Buffer.from('placeholder-pem-content'));
```

The `caData` test uses the same plaintext encoded as base64 but only checks:
```ts
expect(agentOpts?.ca).toBeDefined();
expect(Buffer.isBuffer(agentOpts?.ca)).toBe(true);
```

A bug where `caData` is stored as raw base64 bytes (not decoded PEM) would produce a defined Buffer and pass both assertions — the decode is not actually verified. Add:
```ts
expect(agentOpts?.ca).toEqual(Buffer.from('placeholder-pem-content'));
```
This mirrors the `caFile` test's rigor and makes the base64 decode path actually checked.

---

**2. `k8s-client.test.ts:173` — Hardcoded `/tmp/k8s-client-test-ca.pem` is not parallel-run-safe**

The path `/tmp/k8s-client-test-ca.pem` is a fixed global string. If two CI shards run the `caFile` test concurrently, one worker's `rmSync` in its `finally` block can delete the file while the other worker's `readFileSync` inside `applySecurityAuthentication` is still pending — causing a non-deterministic `ENOENT`. Use a unique suffix:
```ts
import { randomUUID } from 'node:crypto';
const caPath = `/tmp/k8s-client-test-ca-${randomUUID()}.pem`;
```

---

#### ADVISORY — not blocking

**A. `k8s-client.ts:87–91` — `console.warn` fires on every `makeK8sClient` call**

If `makeK8sClient` is ever called more than once (per-request or per-namespace), the same warning emits to stderr repeatedly, burying real log signals. A one-shot module-level guard is the standard mitigation (`let _warned = false`). Not blocking because current callers appear to call `makeK8sClient` once at startup.

**B. `k8s-client.test.ts:120–132` — 13-line describe-block preamble duplicates the file header and `kubernetes.md`**

The comment block explains library internals already covered verbatim in `k8s-client.ts`'s AUTH NOTE (lines 9–16) and the new TLS contract section of `kubernetes.md`. One sentence of unique context ("The limitation is specific to client-cert auth") is all that's needed here. The rest will go stale if the transport layer changes.

---

### Assessment

The approach is sound: testing `applySecurityAuthentication` and asserting `https.Agent` options is the right unit-test strategy (Bun TLS dial-level verification would require a mock TLS server and is out of scope). The AUTH NOTE rewrite is accurate, the documentation follows existing conventions, and the new `console.warn` is correctly conditioned.

Both blocking items are one-line fixes to the test file.

---

**VERDICT: NOT READY — CHANGES REQUIRED**

1. `k8s-client.test.ts:207–218` — Add `expect(agentOpts?.ca).toEqual(Buffer.from('placeholder-pem-content'))` to the `caData` test to verify base64 decoding, matching the `caFile` test's rigor.
2. `k8s-client.test.ts:173` — Replace the hardcoded `/tmp/k8s-client-test-ca.pem` path with a UUID-suffixed unique path to prevent parallel-run collisions.

larryro:
⚠️ **Conflicting PRs for #1849 — do NOT merge any of these until resolved.**

Three open PRs address the same question, and two reach **opposite conclusions** about whether Bun honors kubeconfig TLS knobs:

- **#1919** — concludes `skipTLSVerify`/`caFile` **ARE** honored under Bun.
- **#1916** — concludes Bun **silently ignores** them; only `NODE_EXTRA_CA_CERTS` works, and removes the "dead" code.
- **#1926** — docs companion (aligns with #1919).

This is security-critical: merging the wrong conclusion could leave TLS verification **silently disabled in production**, or remove code that was actually doing the verification. This cannot be resolved by review alone — it needs an **empirical test under the real Bun version** (issue a request with a kubeconfig carrying `caFile`/`skipTLSVerify` and observe whether TLS is actually enforced/skipped). Holding all three for a human decision; only the PR matching the verified behavior should be merged, the others closed.

larryro:
Superseded by #1916 — closing as a duplicate fix for issue #1849.

This PR concluded that `skipTLSVerify`/`caFile` **are** honored under Bun, but that conclusion is empirically incorrect. Its tests only assert the option is stored on the `https.Agent` object (`agent.options.rejectUnauthorized === false`); they never verify TLS is actually enforced/skipped on the wire. An end-to-end test under Bun 1.3.14 (real self-signed HTTPS server + real `CoreV1Api`) shows the library's `node-fetch` transport **drops** the Agent TLS options, so the knobs are INERT — a request still fails with `DEPTH_ZERO_SELF_SIGNED_CERT` even with `skipTLSVerify: true`. Only `NODE_EXTRA_CA_CERTS` provides working CA trust. See the verification write-up on #1916.

PR #1934, opened 2026-06-23
feat(platform): unified chat panel + branch-aware workspace inheritance

Israeltheminer:
## What & why

Two coherent, seam-coupled changes that ship together because the frontend panes consume the new backend queries.

### Right-side panes → one unified Chat panel
Replaces the four independent resizable panes (Plan, Canvas, Workspace files, Live browser) with a single `ChatPanel` shell + context. Each pane is now a **registrar**: it owns its data/state and publishes a `ChatPaneDescriptor`, rendering nothing itself; the shell owns the shared strip, tab bar, minimize/maximize, and resize handle, and keeps every body mounted so the Live browser RFB WebSocket survives tab switches. The per-pane mobile `Sheet`s are gone — the unified panel handles every viewport.

- Canvas file picker rebuilt: horizontal tabs for agent + user-uploaded files plus a bottom "Code output" dock for run outputs, replacing the collapsible `WorkspaceFileSidebar` (deleted).
- `RenderableFileViewer`: Source/Preview toggle moved to a floating control with a live file-size readout and `aria-pressed` state; chosen mode is now sticky per file path.

### Branch-aware workspace inheritance (Convex)
- `listThreadFilesForUser` / `thread_todos.get` now walk the branch ancestor chain (`getBranchAncestorThreadIds`) and union delegate sub-thread artifacts (`getDelegateSubThreadIds`), so files/plans written before a fork point or inside a delegated sub-thread stay visible. Branches record `forkOrderCreatedAt` (optional, additive — **no migration**) to cut ancestor files created after the split.
- Forking a thread snapshots the workspace: `snapshotThreadFiles` copies blobs to fresh `storageId`s (respecting count/byte caps, preserving provenance); `copyThreadTodos` clones the plan verbatim but resets the idempotency ring and integration budget.
- **Fork-cutoff fix** (from review): both fork paths now cut the file snapshot at their conversation boundary so a fork can't inherit files written after it — `forkOwnThread({ upToMessageOrder })` (the "fork from this message" action) by the cutoff message's `_creationTime`, and `forkThread` (share-link fork) by `sharedAt` (the same boundary it already cuts messages at). Both skip the plan copy at a boundary, since a single mutable todos doc can't be reconstructed as-of-cutoff. Previously both copied the *entire current* workspace.

### Polish
Agent-selector "Add agent" footer opens the agent catalog (catalog header gains the create-agent menu); branding form surfaces the "app name required" error while dirty; PDF viewer bootstraps `pdfjsLib` before importing the viewer; Tailwind link-hover selector fix in markdown/release/workflow/docx renderers.

## Verification
- `bun run check` — format, lint, typecheck, **73,349 platform tests pass** (incl. dedicated `ChatPanel`, `WorkspaceFileTabs`, PDF-bootstrap, branding regression, and backend branch-ancestor / snapshot / copy-todos + partial-fork-cutoff suites).
- `bun run lint:sast` — Opengrep, **0 findings** in changed files.
- `bun run --filter @tale/platform migrations:check` — OK (11/11 registered; no schema migration needed).
- Security review of the ancestor/delegate traversal + fork snapshot: **clean** — per-row org filter is the access gate, org is membership-pinned, branch rows are same-owner by write-time invariant, blob copy is org-scoped + count/byte capped, traversal is depth-bounded.
- Docs test suite passes in isolation (the full-monorepo run hit a known-flaky `deploy-sim` temp-dir race under concurrent load — unrelated to this diff).
- ⏳ **Live authenticated UI walk-through pending** (open chat → unified panel → fork) — to be confirmed by reviewer/QA; the dev instance needs sign-in credentials not available in the authoring session.

## Definition of Done
- [x] Translations — `en`, `de`, `fr`, `de-CH` (new `chatPanel`/`canvas` keys; removed pane keys cleaned from the `de-CH` delta)
- [x] Docs — `canvas-pane.md` Source/Preview toggle wording updated in `en`, `de`, `fr`
- [x] Migration — N/A; `forkOrderCreatedAt` is optional + additive; `migrations:check` OK
- [x] Tests — chat-panel, workspace-file-tabs, pdf bootstrap, branding regression, and backend branch-ancestor / snapshot (incl. partial-fork cutoff) / copy-todos suites
- [x] Accessibility — vitest-axe blocks on `ChatPanel` + `WorkspaceFileTabs`; `aria-pressed` added to the file-viewer toggle
- [x] Storybook — N/A; feature components, not `packages/ui` primitives


Israeltheminer:
Thanks @coderabbitai — addressed the bugs in `7ed2f523d`:

**Fixed**
- 🔴 **Fork-point cut now uses `updatedAt`, not `createdAt`** (`thread_files/queries.ts`, both cut sites + `snapshot_thread_files.ts`). `upsertThreadFile` patches a same-path rewrite in place (preserving `createdAt`), so a pre-fork file edited post-fork was leaking its later content. Added a regression test for exactly that (old `createdAt`, new `updatedAt`).
- **Partial fork no longer widens to a full snapshot** (`fork_own_thread.ts`). An empty carried window now yields `createdAtCutoff: 0` (copy nothing) instead of `undefined`. Added the `upToMessageOrder: -1` edge test.
- **`workspace-file-tabs` keyboard trap** — derive an effective active tab (first tab fallback) when `activePath` is null/stale, so exactly one tab keeps `tabIndex={0}`.
- **`workspace-files-pane` stale state / stopped-session recovery** — key the body by `threadId:refreshNonce` so it remounts on thread switch (resets `selectedPath`) and on Refresh (resets `sessionRunning`, letting a stopped session recover).
- **Live browser RFB socket** now disconnects when the session goes inactive (`sessionActive` added to the connect-effect deps).
- **Live browser auto-open** — track the menu-open and control-request edges separately (two `useAutoOpen` calls), so a control request re-surfaces the pane even if it was opened/minimized before.

**Deferred** (not bugs — style/coverage; out of this PR's scope per the author):
- `get_branch_ancestor_thread_ids` ctx typing (`Pick<QueryCtx,'db'>`), `get_delegate_sub_thread_ids` catch `console.warn`, and the extra pdf/todos coverage asks — happy to take these in a follow-up.

The `E2E` failure is a pre-existing onboarding-wizard first-paint flake (the shared `helpers/auth.ts` setup, unrelated to this diff — that workflow fails/cancels on `main` in most recent scheduled runs); all chat-test failures are downstream `(1ms)` setup bailouts from it.

PR #1937, opened 2026-06-23
fix(platform): keep zero-cost fallback models after a credit-exhausted credential (#1454)

### LINKED ISSUE #1454: Bug: Chat requests fail with "AI provider's credit limit exceeded" – no successful response returned
### What happened?

While using the chat feature in the Tale platform, all requests fail with an error indicating that the AI provider’s credit limit has been exceeded. The system attempts multiple model fallbacks, but none succeed, resulting in no response being generated.

### Steps to reproduce

1. Steps to Reproduce
2. Start platform locally
3. Open Chat
4. Enter any prompt (e.g., “Please tell me about yourself”)
5. Send message

### Expected behavior

At least one configured model should return a valid response
If fallback models exist, chat should still work
System should not completely fail due to provider credit issue

### Affected service

Platform

### Version or commit

v.2.66

### Browser and OS

Chrome

### Relevant log output

```shell

```

### Screenshots or recordings

Image

### Before submitting

- [x] I searched existing issues and this hasn't been reported yet.

larryro:
## Problem

Closes #1454.

Chat requests fail with **"The AI provider's credit limit was exceeded"** and **no successful response is returned**, even when fallback models are configured.

In the common single-credential setup (one OpenRouter key, several models on it), the adaptive fallback loop treats an out-of-funds (`credit_exhausted` / HTTP 402) failure as a property of the whole **credential**: it retires `credentialScopeKey(provider+key)` and skips every remaining model that shares that key. That is correct for paid models — but it also skips **zero-cost** models (OpenRouter `:free` variants, or models priced at 0) that draw **no credits** and would have answered. The turn ends with a failed message and nothing tried.

## Fix

Make credit retirement spare zero-cost models, while leaving auth / unreachable retirement untouched.

`services/platform/convex/providers/failure_scope.ts`:
- Split the credit scope into its own namespace `creditScopeKey()` (distinct from the auth `credentialScopeKey()`) so the dead-set records *why* a credential died.
- `retiredScopeKey('credit_exhausted', …)` now retires the **credit** scope; `auth_error` still retires the **auth** credential; `provider_unreachable` still retires the **endpoint**.
- New `isFreeModel()` helper: a model is free when its id carries the OpenRouter `:free` suffix **or** both token prices are explicitly `0`. Unconfigured (`undefined`) pricing is *not* treated as free (conservative).
- `isModelScopeRetired()` skips a model on a credit-dead credential **only when it is not free**. Auth-dead and endpoint-dead resources still skip every model (a bad key / down host kills free models too).

The resolved `ResolvedModelData` already carries `modelId` / `inputCentsPerMillion` / `outputCentsPerMillion`, so no call-site changes were needed.

## Tests

`failure_scope.test.ts` extended: paid sibling still skipped after credit death; `:free` and zero-priced siblings now attempted; free models still skipped on auth / endpoint death; `isFreeModel` truth table.

- `vitest --project server` providers + agent_chat suites: **230 passed**
- `tsc --noEmit`: clean
- `oxlint --type-aware`: 0 warnings / 0 errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model fallback strategy to better distinguish between credential authentication failures and credit exhaustion, allowing free models to remain available for selection even when paid options are unavailable due to credit limits.

### DIFF
--- a/services/platform/convex/lib/agent_chat/internal_actions.ts
+++ b/services/platform/convex/lib/agent_chat/internal_actions.ts
@@ -509,6 +509,9 @@ export async function runGenerationCore(
     // later model sharing that resource is skipped instead of waiting on a
     // doomed request. Keying by credential (not provider name) means a sibling
     // model with its own `secretsEnv` key is still tried after another key dies.
+    // An out-of-funds (credit) failure is the one exception that does NOT doom
+    // every sibling: a zero-cost model on the same credential (`:free` / priced
+    // at 0) draws no credits, so `isModelScopeRetired` still attempts it (#1454).
     let lastFallbackError: unknown;
     const deadScopes = new Set<string>();
 
--- a/services/platform/convex/providers/failure_scope.test.ts
+++ b/services/platform/convex/providers/failure_scope.test.ts
@@ -1,15 +1,24 @@
 import { describe, expect, it } from 'vitest';
 
 import {
+  creditScopeKey,
   credentialScopeKey,
   endpointScopeKey,
+  isFreeModel,
   isModelScopeRetired,
   modelScopeKeys,
   retiredScopeKey,
 } from './failure_scope';
 
 const model = (
-  over: Partial<{ providerName: string; apiKey: string; baseUrl: string }> = {},
+  over: Partial<{
+    providerName: string;
+    apiKey: string;
+    baseUrl: string;
+    modelId: string;
+    inputCentsPerMillion: number;
+    outputCentsPerMillion: number;
+  }> = {},
 ) => ({
   providerName: 'openrouter',
   apiKey: 'key-A',
@@ -47,12 +56,15 @@ describe('endpointScopeKey', () => {
 });
 
 describe('retiredScopeKey', () => {
-  it('retires the CREDENTIAL for funds and auth failures', () => {
-    expect(retiredScopeKey('credit_exhausted', model())).toBe(
-      credentialScopeKey(model()),
-    );
+  it('retires the AUTH credential for an auth failure', () => {
     expect(retiredScopeKey('auth_error', model())).toBe(
       credentialScopeKey(model()),
+    );
+  });
+
+  it('retires the CREDIT credential for an out-of-funds failure', () => {
+    expect(retiredScopeKey('credit_exhausted', model())).toBe(
+      creditScopeKey(model()),
     );
   });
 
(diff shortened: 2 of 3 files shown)


larryro:
## Desk Review — #1454 fix (keep zero-cost fallback models after a credit-exhausted credential)

**Verdict: READY TO MERGE.**

I reviewed the whole implementation (not just the diff), traced both consumers, ran the project's tests, and confirmed CI.

### What the fix does, and why it's correct
`credit_exhausted` previously retired the entire `credentialScopeKey`, so `isModelScopeRetired` skipped **every** sibling model on the same key — including zero-cost ones that draw no credits. That's the root cause of "fallbacks attempted, none succeed." The fix splits credential death into two namespaces — `cred:` (auth) and `credit:` (funds) — via the new `creditScopeKey`, and makes credit retirement conditional: `isModelScopeRetired` still skips paid siblings but spares free ones (`isFreeModel`: OpenRouter `:free` suffix, or explicit `0`/`0` token pricing). Auth and endpoint deaths still retire free models too (checked first, unconditionally).

Verified:
- `retiredScopeKey` auth path is byte-identical to before; only the credit path moved to a new prefix. No code reads the raw `cred:`/`credit:` prefixes — both consumers (`agent_chat/internal_actions.ts` and `workflow_engine/.../execute_llm_node.ts`) interact only through the helpers, so the split is transparent and consistent across both.
- `ResolvedModelData.modelId` is always present and pricing flows from `definition.cost?.{input,output}CentsPerMillion`, so `isFreeModel` reads real data. Strict `=== 0` means *unconfigured* pricing is correctly **not** treated as free → paid-but-unpriced models stay retired (safe direction).
- 402 / credit-phrase errors classify to `credit_exhausted` (`lib/error_classification.ts`) ahead of the auth branch — the exact path this fixes.
- Loop is bounded: each model index is attempted at most once; a free model that itself 402s retires only the `credit:` scope and doesn't wrongly block other frees or un-block paid siblings.

### Tests / CI
- `bunx vitest --run --project server`: **501 files, 6135 tests, all pass** (no regressions).
- `failure_scope.test.ts`: **22/22 pass**, covering paid-still-skipped, `:free`/zero-priced spared, free-still-skipped on auth & endpoint death, and `isFreeModel` truth table.
- CI on the PR: **46 pass, 5 skipping (conditional jobs), 0 fail, 0 pending** — green.

### Scope note (non-blocking)
`fallbackModels` is purely operator-configured (`supportedModels.slice(1)`); there's no hardcoded free default. So this fix rescues chat **when a zero-cost fallback is configured**. A workspace whose only models are paid on the one exhausted key will still fail — but that's correct behavior (no model *can* run when the funds are gone and everything costs money), not a bug. The fix correctly targets the actual defect: wrongly skipping viable zero-cost siblings.

### Optional polish (nits — not blocking)
1. `creditScopeKey` vs `credentialScopeKey` differ by one substring and produce near-identical strings; a typo would silently misroute. Consider `fundsScopeKey`/`authScopeKey` for harder-to-confuse names.
2. `isFreeModel` uses `.includes(':free')`; `.endsWith(':free')` matches the OpenRouter suffix convention more precisely.
3. Add unit assertions for one-side-zero pricing (`{input:0, output:250}` → not free) and mixed-case `:FREE` to lock in the `.toLowerCase()` path.

None of these affect correctness. **Approving.**

PR #1940, opened 2026-06-23
fix(platform): surface actionable error when whole fallback chain is unconfigured (#1455)

### LINKED ISSUE #1455: Bug: Fallback model chain fails across all providers – no successful response even after multiple retries
### What happened?

The system attempts multiple fallback models across different providers when a request fails, but none succeed. This results in a complete failure of the chat system instead of graceful degradation.

### Steps to reproduce

1. Start platform locally
2. Open Chat
3. Enter any prompt (e.g., “Please tell me about yourself”)
4. Send message

### Expected behavior

At least one fallback model should succeed
Or system should stop early with meaningful configuration error

### Affected service

Platform

### Version or commit

0.2.26

### Browser and OS

_No response_

### Relevant log output

```shell
gemini → gemini-flash → mistral-large → mistral-medium → llama → qwen → qwen-next
```

### Screenshots or recordings

_No response_

### Before submitting

- [x] I searched existing issues and this hasn't been reported yet.

larryro:
## Summary

Fixes #1455.

When the agent fallback model chain (e.g. `gemini → gemini-flash → mistral-large → … → qwen-next`) is exhausted because **every** model fails to *resolve* for a configuration reason — most commonly no API key is configured on any provider — the loop previously threw the **last** entry's per-model `MissingApiKeyError`. The user saw a confusing message implicating only the tail model (`qwen-next`) after "multiple retries", with no hint that the real problem is that *no* provider is configured.

This matches the issue's expected behaviour: *"the system should stop early with a meaningful configuration error."*

## What changed

- **`providers/errors.ts`**
  - Added `buildChainExhaustionError(...)`: when no model ever reached a live provider call (`attemptedCount === 0`) and every failure was a config/resolution error, it collapses the chain into a single actionable `NoProviderAvailableError` (which `classifyChatErrorCode` maps to `missing_api_key` → the localized "Open Settings → AI providers" hint). An already-terminal `NoProviderAvailableError` is passed through untouched; a chain where a real attempt happened still surfaces the genuine runtime error.
  - Moved the shared `FRIENDLY_NO_PROVIDER` copy here so the provider loader and the fallback chain share one source of truth (no string drift).
- **`providers/file_actions.ts`** — imports `FRIENDLY_NO_PROVIDER` instead of redefining it.
- **`lib/agent_chat/internal_actions.ts`** — the fallback loop now records each resolution (config) failure and routes the chain-exhausted throw through `buildChainExhaustionError`.

No new user-visible strings (the friendly copy already existed and is rendered via the existing `missing_api_key` chat-error code), so no i18n/docs changes are required.

## Tests

- Added unit coverage in `providers/errors.test.ts` for `buildChainExhaustionError` (all-unconfigured collapse, runtime-error passthrough, terminal passthrough, fallbacks). `errors.test.ts`: 51 passed.
- `oxlint --type-aware` clean on all changed files.

### DIFF
--- a/services/platform/convex/lib/agent_chat/internal_actions.ts
+++ b/services/platform/convex/lib/agent_chat/internal_actions.ts
@@ -31,6 +31,7 @@ import { routeSeedValidator, routeTuningValidator } from '../../agents/schema';
 import { resolveOrgSlug } from '../../organizations/resolve_org_slug';
 import { recordFailure } from '../../providers/circuit_breaker';
 import {
+  buildChainExhaustionError,
   classifyFailureScope,
   isTransientProviderError,
 } from '../../providers/errors';
@@ -511,6 +512,12 @@ export async function runGenerationCore(
     // model with its own `secretsEnv` key is still tried after another key dies.
     let lastFallbackError: unknown;
     const deadScopes = new Set<string>();
+    // Resolution (config) failures collected across the chain. When NO model
+    // ever reaches a live provider call (`attemptedCount === 0`) these are the
+    // whole story, and the chain collapses into one actionable
+    // "configure a provider" error instead of the last entry's per-model
+    // message (issue #1455).
+    const configFailures: Array<{ model: string; message: string }> = [];
 
     // RESOLUTION (no HTTP) is memoized per model index so the catch can look
     // ahead for the next attemptable model without paying for it twice. A
@@ -568,6 +575,13 @@ export async function runGenerationCore(
       const resolution = await resolveAt(attempt);
       if (!resolution.ok) {
         lastFallbackError = resolution.error;
+        configFailures.push({
+          model: currentModelId ?? 'default',
+          message:
+            resolution.error instanceof Error
+              ? resolution.error.message
+              : String(resolution.error),
+        });
         if (attempt < modelsToTry.length - 1) {
           debugLog('SKIP_UNCONFIGURED_MODEL', {
             model: currentModelId ?? 'default',
@@ -579,7 +593,10 @@ export async function runGenerationCore(
           });
           continue;
         }
-        throw resolution.error;
+        // Last entry also failed to resolve. Fall through to the unified
+        // exhaustion handler below so an all-unconfigured chain surfaces one
+        // actionable error rather than this tail model's per-model message.
+        break;
       }
       const resolved = resolution.resolved;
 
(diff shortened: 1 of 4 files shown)


larryro:
## Desk Review — #1455 fallback chain unconfigured-error fix

**Verdict: READY TO MERGE.**

### CI
All required checks **green** (`gh pr checks`): Type check, Lint, Format, Knip, Unit, Build (all services), 16× Playwright platform shards, Smoke test, CodeRabbit, Opengrep — all `pass`. No failed or pending checks; only the expected fork-PR / container variants are `skipping`.

### Tests (run locally on this branch)
- `bunx vitest --run --project server convex/providers/errors.test.ts` → **51 passed**
- `bunx vitest --run --project server convex/providers convex/lib/agent_chat lib/agent_response lib/shared` → **82 files, 1302 passed**

### What the change does
When the whole fallback chain is exhausted with **no model ever reaching a live provider call** (`attemptedCount === 0` — every entry failed during in-process resolution because no usable API key is configured), `buildChainExhaustionError` collapses the chain into a single actionable `NoProviderAvailableError('missing_api_key')` carrying the `FRIENDLY_NO_PROVIDER` copy, instead of surfacing the tail model's confusing per-model message. The tail-entry resolution failure now `break`s into this unified handler rather than throwing directly.

### Verified correct
- **Trigger path is real:** `MissingApiKeyError` is thrown during resolution (`resolveModelDataInline` → `resolveModelApiKey`), so `attemptedCount` stays 0 for an all-unconfigured chain and the collapse fires. Confirmed by reading `resolve_model_node.ts` / `file_actions.ts`.
- **No false-positive collapse:** `attemptedCount` (init 0 @ line 347) increments only inside the `try` after a successful resolution and before the provider call (line 621). Any chain with a configured model that reached a provider returns the genuine runtime error unchanged (`attemptedCount > 0` gate).
- **Error-code routing:** the `FRIENDLY_NO_PROVIDER` string matches `classifyChatErrorCode`'s `/no api key is configured for this organization/i`, mapping to `missing_api_key` and the actionable UI hint. Notably also improves the single-unconfigured-model case (previously fell through to `generic`).
- **No cleanup regression:** the `break` lands directly on the `buildChainExhaustionError` throw with no intervening code; the outer catch (stream error, status clear, failed-message save, `NonRetryableError`) is error-source-agnostic.
- **`NoProviderAvailableError` passthrough** preserved; dead-scope/runtime-error paths still surface the true cause.
- **No secret leakage:** detail list contains model ids and the `secretsEnv` variable *name* only — never key values.
- **Elegance:** `FRIENDLY_NO_PROVIDER` now has exactly one definition (moved to `errors.ts`, re-imported in `file_actions.ts`); no duplication/dead code; Knip green.

### Non-blocking notes (optional)
1. The loop→`buildChainExhaustionError` integration (the tail-resolution `break`) has no end-to-end test; the helper itself is fully unit-covered. An integration test would harden against future loop refactors.
2. `errors.test.ts` asserts only the first/last model appear in `details`; adding `mistral-large` would fully match the "every failed model is preserved" comment.

These do not affect merge-readiness.

**READY TO MERGE.**

PR #2101, opened 2026-06-24
fix(platform): require organization name (#1951)

### LINKED ISSUE #1951: Improvement: Make organization name required
## Problem
The organization name is not enforced as required.

## Current behavior
Org settings form `services/platform/app/features/settings/organization/components/organization-settings.tsx` — name not enforced as required.

## Proposed change
Add `.min(1)` (required) validation to the org name field and mark the label required; enforce on the backend mutation too.

## Acceptance criteria
- [ ] Saving an empty org name is blocked client- and server-side with a clear message.

larryro:
## Summary

Resolves #1951 — the organization name is now enforced as **required**, client- and server-side.

## Changes

- **Client validation** (`organization-settings.tsx`): added a Zod schema (`name: z.string().trim().min(1)`) wired through `useFormEditor`, so an empty/whitespace name blocks Save and surfaces an inline error via the existing `Input` `errorMessage` path. The save now sends the trimmed name unconditionally (no longer `|| undefined`).
- **Required label** (`settings-row.tsx`): added an optional `required` prop that renders the red `*` asterisk (mirrors the `Label` primitive, with a translated `aria-label`); the org-name row opts in.
- **Server enforcement** (`convex/auth.ts`, `beforeUpdateOrganization`): rejects an org rename that clears the name (empty/whitespace) with `APIError('BAD_REQUEST')` and trims the stored value, so the rule can't be bypassed via a direct API call. Only enforced when `name` is part of the patch (locale-only updates are untouched).
- **i18n**: added `settings.organization.nameRequired` to `en`/`de`/`fr`. Also fixed the German/French `settings.organization.title`, which still read "Organisation (optional)" / "(facultatif)" — corrected to "Organisationsname" / "Nom de l'organisation" to match the now-required English label.

## Tests

- New unit tests in `organization-settings.test.tsx`: empty name → inline error + `isValid: false`; entering a name clears the error and re-validates.
- Ran the platform UI suite (`organization-settings`, `settings-row`) — 12 passing — plus the i18n parity suite (34 passing) and `bun run typecheck` (clean).

## Acceptance criteria

- [x] Saving an empty org name is blocked client- and server-side with a clear message.

## Notes

- Docs: N/A — no docs page documented the org-name field's optionality.
- Migrations: N/A — no schema change.

## Summary by CodeRabbit

* **New Features**
  * Organization name field now requires input with validation that displays error messages when empty or whitespace-only.

* **Localization**
  * Added organization name validation messages in English, German, and French.

### DIFF
--- a/services/platform/app/features/settings/components/settings-row.tsx
+++ b/services/platform/app/features/settings/components/settings-row.tsx
@@ -3,6 +3,7 @@
 import { Description } from '@tale/ui/description';
 import { forwardRef, useId, type HTMLAttributes, type ReactNode } from 'react';
 
+import { useT } from '@/lib/i18n/client';
 import { cn } from '@/lib/utils/cn';
 
 interface SettingsRowProps extends Omit<
@@ -11,6 +12,8 @@ interface SettingsRowProps extends Omit<
 > {
   label: ReactNode;
   description?: ReactNode;
+  /** Append a red required asterisk to the label (mirrors `Label`'s required). */
+  required?: boolean;
   /** Right-side control (switch, button, copy field, link). */
   children: ReactNode;
 }
@@ -21,7 +24,8 @@ interface SettingsRowProps extends Omit<
  * narrow viewports so the right-side control wraps cleanly.
  */
 export const SettingsRow = forwardRef<HTMLDivElement, SettingsRowProps>(
-  ({ label, description, children, className, ...props }, ref) => {
+  ({ label, description, required, children, className, ...props }, ref) => {
+    const { t } = useT('common');
     const id = useId();
     const labelId = `${id}-label`;
     const descId = description ? `${id}-desc` : undefined;
@@ -47,6 +51,14 @@ export const SettingsRow = forwardRef<HTMLDivElement, SettingsRowProps>(
             className="text-foreground text-sm leading-none font-medium"
           >
             {label}
+            {required && (
+              <span
+                className="ml-1 text-red-600"
+                aria-label={t('aria.required')}
+              >
+                *
+              </span>
+            )}
           </span>
           {description && <Description id={descId}>{description}</Description>}
         </div>
--- a/services/platform/app/features/settings/organization/components/organization-settings.test.tsx
+++ b/services/platform/app/features/settings/organization/components/organization-settings.test.tsx
@@ -1,11 +1,18 @@
 import { describe, it, expect, vi } from 'vitest';
-import { z } from 'zod';
+import { z } from 'zod/v4';
 
 import { useFormEditor } from '@/app/components/ui/editor';
+import { organizationNameSchema } from '@/lib/shared/schemas/organizations';
+import enMessages from '@/messages/en.json';
 import { checkAccessibility } from '@/tests/utils/a11y';
 import { fireEvent, render, screen, waitFor } from '@/tests/utils/render';
 
 import { OrganizationSettingsView } from './organization-settings';
+
+// Resolve user-facing labels through the translation layer rather than
+// hardcoding English, matching the platform i18n rule for tests.
+const orgNameLabel = enMessages.settings.organization.title;
+const nameRequiredMessage = enMessages.settings.organization.nameRequired;
 
 // The view now embeds the Members section, which subscribes to Convex via
 // `useMembers`. This suite renders the view in isolation (no Convex provider)
@@ -40,6 +47,7 @@ function Harness() {
       memberContext={null}
       canDelete={false}
       isCurrentOrganization
+      onSave={save}
     />
   );
 }
@@ -65,6 +73,7 @@ function LoadHarness({ orgName }: { orgName: string }) {
       memberContext={null}
       canDelete={false}
       isCurrentOrganization
+      onSave={save}
     />
   );
 }
@@ -85,7 +94,7 @@ describe('OrganizationSettingsView page load', () => {
     // The org-name field, labeled by `settings.organization.title`, resolves to
     // the current (non-empty) org name once the form applies its baseline.
     const orgNameField = screen.getByRole('textbox', {
-      name: 'Organization name',
+      name: orgNameLabel,
     });
     await waitFor(() => expect(orgNameField).toHaveValue('Acme Industries'));
     // Mirror the E2E's "non-empty value" intent.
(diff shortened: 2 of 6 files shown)


larryro:
## Desk Review — #1951 "Make organization name required" (PR #2101)

**Verdict: NOT READY — one change required (test coverage).** The implementation itself is correct, well-scoped, and on-convention; CI is fully green. The single blocker is a missing automated test for the server-side half of the acceptance criterion.

### CI & tests (observed)
- `gh pr checks` — **all green** (Build platform, 16 Playwright platform shards, Lint, Format, Knip, Migrations, CodeQL/Opengrep, CodeRabbit). Only skip is the fork-PR smoke test (expected).
- Ran locally on the branch:
  - `bunx vitest --run --config vitest.ui.config.ts organization-settings` → **5 passed**
  - `bunx vitest --run --config vitest.ui.config.ts settings-row` → **7 passed**
  - `bunx vitest --run --project server convex/auth.test.ts` → **3 passed**

### What's correct (verified by tracing every branch, twice)
- **Client gate works.** `organizationSchema` (`z.string().trim().min(1)`) is wired via `useFormEditor` → `zodResolver` (`mode: onChange`); `handleSubmit`'s invalid path rejects with `VALIDATION_FAILED` so the network `save` cannot run on an empty/whitespace name. `isValid` reflects the schema. (`organization-settings.tsx:325-333,396`, `use-form-editor.ts:157-191,223`)
- **Server hook is correct on all branches** (`convex/auth.ts:716-724`, `beforeUpdateOrganization`): name check runs *before* the slug early-return so name-only updates are still validated; locale-only updates (no `name` key) correctly skip; whitespace-only, non-string, and null are rejected; valid name is trimmed and persisted via the same live-payload pattern as slug normalization. Error type `APIError('BAD_REQUEST', …)` is consistent with the file.
- **`name: data.name.trim() || undefined` → `data.name.trim()`** is the right consequence of the requirement; no previously-allowed path breaks.
- **i18n complete.** `settings.organization.nameRequired` present in en/de/fr; de/fr titles updated from "(optional)/(facultatif)" to a name, consistent with en ("Organization name"). `common.aria.required` (used by the new asterisk) exists; `de-CH.json`/`global.json` omissions resolve via i18next base-locale fallback.
- **Elegance.** `SettingsRow required` asterisk faithfully mirrors the platform `Label` (same class/`aria-label`); no double-asterisk (Input renders its own label only with a `label` prop, here it uses `aria-label`); schema-in-`useMemo([tSettings])` matches the `account-form` / `team-edit-dialog` convention.

### Blocking finding
1. **Server-side enforcement is untested.** The acceptance criterion requires the empty name to be blocked **client- AND server-side**. The client side has new tests; the server-side branch in `beforeUpdateOrganization` (`convex/auth.ts:716-724`) has **no coverage** — `auth.test.ts` only tests `trustedOrigins`, and the Playwright E2E only drives the client-gated UI, so neither exercises this error branch. The existing `auth.test.ts` mock stubs `organization` as `vi.fn(() => ({}))`, discarding the hook, so a test must capture the hook config and invoke `beforeUpdateOrganization` directly. **Add a server test asserting:** (a) a `{ organization: { name: '   ' } }` update throws `APIError('BAD_REQUEST')`, (b) a non-string/empty name is rejected, (c) a valid name is trimmed-and-persisted, (d) a name-absent (locale-only) payload passes untouched.

### Non-blocking notes (optional)
- Test re-declares a stand-in `organizationSchema` with a hardcoded English message rather than the container's real schema (the real one isn't exportable since it closes over `tSettings`); these client tests therefore don't guard the container's actual schema wiring/i18n key. Consider extracting the schema if you want that guarded.
- The `SettingsRow` `required`-asterisk render branch is untested in `settings-row.test.tsx`.
- Asymmetry (out of issue scope): org **create** (`beforeCreateOrganization`) validates only slug, not name — a direct-API create could still set an empty name.

**NOT READY — CHANGES REQUIRED:** add the server-side test for the org-name-required branch in `convex/auth.ts` `beforeUpdateOrganization`.

larryro:
## Desk Review — fix(platform): require organization name (#1951)

**Verdict: READY TO MERGE.**

### CI
All checks green on PR #2101 (`gh pr checks`): Build (all images), Lint, Type check, Format, Knip, Unit/UI/Browser tests, all 16 Playwright platform shards, Playwright web/docs, CodeQL, Opengrep, Smoke test, Scan platform. Trivy is `NEUTRAL` (informational, not a failure); fork-only jobs are skipped as expected. No red, no pending.

### Tests run locally (this review)
- `bunx vitest --run --project server convex/auth.test.ts` → **7 passed**
- `bunx vitest --run --config vitest.ui.config.ts organization-settings.test.tsx input.test.tsx` → **37 passed**

### What I verified (two rounds, fan-out + confirmation)
- **Shared rule, single source.** `organizationNameSchema(message?) = z.string().trim().min(1, message)` (`lib/shared/schemas/organizations.ts`) — `.trim().min(1)` correctly rejects empty AND whitespace-only. Used by both the client form and the server hook, so they can't drift.
- **Client (blocked).** `useFormEditor` runs `zodResolver` in `mode: 'onChange'`; the inline error renders via `errorMessage={errors.name?.message}`, and the Save button is disabled when `!isValid` (`editor-actions.tsx:147`) while `doSave` rejects with `VALIDATION_FAILED` on an invalid submit. An empty/whitespace name cannot reach `authClient.organization.update`. The save payload change `data.name.trim() || undefined` → `data.name.trim()` is correct and actually removes a latent bug where an emptied field used to send `undefined` (silently skipping the update).
- **Server (blocked).** `beforeUpdateOrganization` (`convex/auth.ts:717`) re-validates with the same schema when `name` is present in the patch, throws `BAD_REQUEST` on failure, and writes the trimmed value back (`orgPatch.name = parsedName.data`) using the same project-back pattern as the established slug normalization. Name-absent (locale-only) updates pass through untouched. This hook is the chokepoint for all better-auth org updates, so a direct API call can't bypass it.
- **i18n.** `settings.organization.nameRequired` added to en/de/fr; the `title` label renamed from "(optional)/(facultatif)" to the required form in all three; `common.aria.required` exists in all three for the asterisk. `de-CH` correctly falls back. No stale "(optional)" references remain.
- **Tests.** New client tests (empty → error+invalid; whitespace → invalid; error clears on valid) build the harness from the real shared schema and resolve labels via the i18n layer. New server tests cover whitespace, non-string, trim+persist, and name-absent. Both reach the real code, not reimplementations.

1. `beforeCreateOrganization` does not validate the name server-side. Org creation already enforces a non-empty name client-side in the onboarding flow (`workspace-step.tsx:78`), and #1951 targets the settings rename, so this is out of scope — but server-side parity for creation would be a reasonable defense-in-depth follow-up.
2. The server error string `'Organization name is required.'` is hardcoded English, consistent with the other `APIError` messages in `auth.ts`. Client-facing copy is localized.
3. Optional extra coverage: an explicit `name: null` server case and a direct "Save disabled when invalid" assertion would document intent, but behavior is already correct and covered indirectly.

**READY TO MERGE.**

PR #2120, opened 2026-06-24
fix: seed enterprise sso form with defined defaults to avoid controlled warnings (#2095)

### LINKED ISSUE #2095: Bug: Enterprise SSO settings form logs React uncontrolled→controlled warnings (Select + Switch initialized undefined)
**Area / route**
`/dashboard/{org}/settings/enterprise-sso` — `app/features/settings/enterprise-sso/components/enterprise-sso-form.tsx` (new feature, commit b4180755d).

**Steps to reproduce**
1. Open Settings → Enterprise SSO with the console open.
2. Watch the page load.

**Expected**
No controlled/uncontrolled warnings; fields initialized with defined defaults so state does not churn (no risk of reset on first interaction).

**Actual**
On load, 5 console warnings fire: "Select/Switch is changing from uncontrolled to controlled". The `data` useMemo returns undefined while config loads, so the Select/Switch fields mount with value/checked undefined then hydrate.

**Evidence**
- `enterprise-sso-form.tsx:266-267` (data useMemo returns undefined while loading); Select value={field.value} at :558,:749 (2); Switch checked={field.value} at :672,:770,:781 (3) = 5 warnings (matches observed). Screenshot settings/enterprise-sso-controlled-warning.png.

**Notes**
Live browser finding (B3); NEW feature, NOT in the dedup baseline. Fix: provide defined default values (empty strings for Selects, false for Switches) before config resolves so the controls are controlled from first render. Same new form is the subject of the protocol-switch validation gap [08].

---

larryro:
## Summary

Fixes #2095 — the Enterprise SSO settings form logged five React *"changing from uncontrolled to controlled"* warnings on load (2× Select, 3× Switch).

### Root cause
`EnterpriseSsoForm`'s `data` useMemo returns `undefined` while the connection config is loading, and `useFormEditor` passed that straight to RHF's `defaultValues`. The controlled `Select`/`Switch` inputs therefore mounted with `value`/`checked` = `undefined` (uncontrolled), then flipped to controlled once `data` resolved and the form reset.

### Fix
- `useFormEditor` gains an optional `defaultValues` argument used purely for RHF's initial `defaultValues` (read once at mount). `data` stays the loading sentinel, so `isLoading` / dirty-suppression / the remote-update reset logic are unchanged.
- `EnterpriseSsoForm` passes a fully-defined `EMPTY_FORM_DATA` (empty strings for text/Selects, booleans for Switches) so every control is controlled from the first render.

### Verification
- `bunx tsc --noEmit` (platform): clean.
- `bunx oxlint --type-aware` on touched files: 0 warnings, 0 errors.
- `oxfmt --check`: clean.
- Unit tests (vitest.ui.config.ts): `use-form-editor` 10/10 (added a regression test asserting defined defaults while loading) and existing `enterprise-sso` suite 8/8.

### DIFF
--- a/services/platform/app/components/ui/editor/use-form-editor.test.tsx
+++ b/services/platform/app/components/ui/editor/use-form-editor.test.tsx
@@ -135,7 +135,7 @@ describe('useFormEditor', () => {
           schema,
           save: vi.fn().mockResolvedValue(undefined),
         }),
-      { initialProps: { data: undefined } },
+      { initialProps: { data: undefined as Form | undefined } },
     );
 
     // Loading, but the controlled fields already read defined values rather
--- a/services/platform/app/components/ui/editor/use-form-editor.test.tsx
+++ b/services/platform/app/components/ui/editor/use-form-editor.test.tsx
@@ -126,6 +126,31 @@ describe('useFormEditor', () => {
     expect(result.current.reset).toBe(reset0);
   });
 
+  it('seeds defined defaultValues while data is still loading (controlled from first render)', () => {
+    const { result, rerender } = renderHook(
+      ({ data }: { data: Form | undefined }) =>
+        useFormEditor<Form>({
+          data,
+          defaultValues: { name: '', color: '' },
+          schema,
+          save: vi.fn().mockResolvedValue(undefined),
+        }),
+      { initialProps: { data: undefined } },
+    );
+
+    // Loading, but the controlled fields already read defined values rather
+    // than undefined — no uncontrolled→controlled churn when data arrives.
+    expect(result.current.isLoading).toBe(true);
+    expect(result.current.form.getValues('name')).toBe('');
+    expect(result.current.form.getValues('color')).toBe('');
+    expect(result.current.isDirty).toBe(false);
+
+    rerender({ data: { name: 'A', color: '#FF0000' } });
+    expect(result.current.isLoading).toBe(false);
+    expect(result.current.form.getValues('name')).toBe('A');
+    expect(result.current.isDirty).toBe(false);
+  });
+
   it('reports isValid:false for schema-invalid input', async () => {
     const { result } = mount({ name: 'A', color: '#FF0000' });
     act(() => result.current.form.setValue('name', '', { shouldDirty: true }));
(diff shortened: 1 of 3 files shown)


larryro:
## Desk Review — #2095 (uncontrolled→controlled warnings on Enterprise SSO form)

**Verdict: READY TO MERGE.**

**CI:** All checks green (`gh pr checks` — Build platform, type-aware Lint, Knip, Format, Browser, UI, all 16 Playwright platform shards, etc.; "Docs container test" skipped as expected). No red, no pending.

**Tests run locally (once):** `bunx vitest --run --config vitest.ui.config.ts app/components/ui/editor/use-form-editor.test.tsx` → **10 passed**.

### What the fix does
- `use-form-editor.ts`: adds an optional `defaultValues` arg and changes `defaultValues: data as …` → `defaultValues: (data ?? defaultValues) as …`. RHF reads `defaultValues` once at mount, so while `data` is `undefined` the form is seeded with defined values; the existing `useEffect` still `form.reset(data)`s once data arrives.
- `enterprise-sso-form.tsx`: adds `EMPTY_FORM_DATA: SsoFormData` (every field defined) and passes it as `defaultValues`.

### Why it's correct (verified across all branches)
- **Resolves all 5 cited warnings.** Each Select/Switch field now mounts controlled: `protocol`→`'entra-id'` (Select), `defaultRole`→`'member'` (Select), `pkce`→`true`, `autoRole`→`false`, `autoTeam`→`false` (Switches). No controlled input can mount `undefined`.
- **Production path is real:** `enterprise-sso-settings.tsx` passes `config=undefined` during load straight into the form, so the targeted loading window is genuinely hit.
- **No collateral on the shared hook.** `defaultValues` is optional; for the ~20 other `useFormEditor` callers `data ?? undefined` is byte-for-byte the prior behavior. `isLoading` (`data === undefined`), the `isDirty` suppression-while-loading, `hasInitializedRef`/`lastDataRef`, and the Convex baseline-drift / remote-update logic are all untouched — the seed is inert after the first `form.reset(data)`.
- **Type-safe & DRY-safe:** `EMPTY_FORM_DATA: SsoFormData` forces the compiler to cover every field if the interface grows; type-aware Lint is green.
- **Genuine regression guard:** reverting the one-line change makes the new test fail (`getValues` returns `undefined`) — confirmed empirically.

### Non-blocking observations (optional, do not gate merge)
1. The new test asserts defined values rather than spying on `console.error`/`warn` for the actual "uncontrolled to controlled" string — a `vi.spyOn(console,'error')` assertion would tie the test more directly to the symptom.
2. Coverage is hook-level only; no component-level test renders `EnterpriseSsoForm` through the `config=undefined`→loaded transition. Acceptable given the hook test pins the mechanism and Browser/Playwright CI is green.

The related "protocol-switch validation gap [08]" is correctly out of scope here.

PR #2130, opened 2026-06-24
fix(ui): add disabledReason tooltip explaining why a control is disabled (#1949)

### LINKED ISSUE #1949: Improvement: Add tooltips explaining why a disabled control is disabled
## Problem
Disabled buttons / fields / controls give no reason for being disabled.

## Current behavior
Tooltip primitives exist (`packages/ui/src/components/overlays/tooltip.tsx`, app wrapper `services/platform/app/components/ui/overlays/tooltip.tsx`) but are not systematically attached to disabled controls. Note: disabled elements don't emit pointer events, so the tooltip trigger needs a wrapper span / `pointer-events` handling.

## Proposed change
- Add a shared `disabledReason` (tooltip) affordance to the shared `Button`, `IconButton`, and form-field primitives that, when disabled, wraps the control so a tooltip + `aria-describedby` explains why.
- Apply it to **all** controls that have a disabled state.

## Acceptance criteria
- [ ] Every disabled button/field/control surfaces a reason on hover/focus.
- [ ] Works for keyboard users (accessible name/description).

larryro:
## What & why

Resolves #1949. Disabled buttons/controls gave no reason for being disabled. A natively-`disabled` button emits no pointer events and leaves the tab order, so a tooltip could never reach it on hover *or* focus.

## Change

Adds a shared `disabledReason` affordance to the `@tale/ui` `Button` primitive (inherited for free by `IconButton`, which composes `Button`). When a button is `disabled` **and** carries a `disabledReason`:

- it is kept focusable and switched from the native `disabled` attribute to `aria-disabled` (still rendered visually disabled via new `aria-disabled:*` variants, and inert to click / Enter / Space);
- the reason is shown through the existing shared Tooltip on **hover and focus**, and Radix wires `aria-describedby` so screen-reader users hear it too.

`disabledReason` is ignored while the button is enabled (callers can pass it unconditionally) and suppressed under `asChild` (the button is then a Radix `Slot`). Existing `disabled` behaviour is unchanged when no reason is supplied.

## Acceptance criteria

- [x] Every disabled button/control can surface a reason on hover/focus via one shared prop.
- [x] Works for keyboard users (focusable, `aria-disabled`, `aria-describedby` via Tooltip).

## Verification

- `bun run --filter @tale/ui test` — full unit suite green, incl. 6 new `Button` cases + 1 `IconButton` case and an axe audit (`checkAccessibility`) on the soft-disabled markup.
- `bun run --filter @tale/ui typecheck` and `lint` (oxlint, type-aware) — clean.
- Added a `DisabledWithReason` Storybook story (Storybook a11y project runs in CI; its browser couldn't be downloaded in the local sandbox).

## Notes / Definition of Done

- No user-facing strings added — the reason text is caller-provided — so i18n / docs are **N/A**.
- Not a new primitive (a prop on an existing one); story added, a11y covered, `Skeletonize` already wraps the button.

### DIFF
--- a/packages/ui/src/components/forms/slider.tsx
+++ b/packages/ui/src/components/forms/slider.tsx
@@ -230,6 +230,12 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
               value={value}
               disabled={softDisabled ? undefined : disabled}
               aria-disabled={disabled || undefined}
+              // The range stays value-controlled while soft-disabled, so React
+              // would warn about a `value` with no `onChange`. `readOnly`
+              // silences that without changing behaviour — drag is already
+              // blocked via `onPointerDown` and arrow keys via `onKeyDown` —
+              // matching how `input.tsx`/`textarea.tsx` freeze their fields.
+              readOnly={softDisabled ? true : undefined}
               onChange={softDisabled ? undefined : handleChange}
               onPointerDown={(event) => {
                 if (softDisabled) {
--- a/packages/ui/src/components/forms/slider.tsx
+++ b/packages/ui/src/components/forms/slider.tsx
@@ -1,6 +1,7 @@
 import {
   type ChangeEvent,
   type InputHTMLAttributes,
+  type KeyboardEvent,
   type ReactNode,
   forwardRef,
   useEffect,
@@ -10,6 +11,7 @@ import {
 
 import { cn } from '../../lib/cn';
 import { SkeletonBox } from '../feedback/skeleton';
+import { DisabledReasonTooltip } from '../overlays/disabled-reason';
 
 export interface SliderProps extends Omit<
   InputHTMLAttributes<HTMLInputElement>,
@@ -34,10 +36,36 @@ export interface SliderProps extends Omit<
   valueLabel?: ReactNode;
   /** Time in ms to keep the value label visible after interaction. Default 1500. */
   valueLabelHideDelay?: number;
+  /**
+   * Explains *why* the slider is disabled, surfaced in a tooltip on hover AND
+   * focus and to screen readers (#1949). Only takes effect while the slider is
+   * `disabled`; ignored otherwise, so callers can pass it unconditionally.
+   *
+   * A natively-`disabled` range emits no pointer events and leaves the tab
+   * order, so no tooltip could reach it. When a disabled slider carries a
+   * `disabledReason` we therefore keep it focusable (still rendered visually
+   * disabled and inert — the value can't change via drag or arrow keys), swap
+   * `disabled` for `aria-disabled`, and let the shared Tooltip wire up
+   * `aria-describedby`.
+   */
+  disabledReason?: ReactNode;
 }
 
 const THUMB_DIAMETER_PX = 20;
 const THUMB_HALF_PX = THUMB_DIAMETER_PX / 2;
+
+// Keys a native range responds to by changing its value — blocked while a
+// slider is soft-disabled so keyboard users can't nudge an inert control.
+const VALUE_KEYS = new Set([
+  'ArrowLeft',
+  'ArrowRight',
+  'ArrowUp',
+  'ArrowDown',
+  'Home',
+  'End',
+  'PageUp',
+  'PageDown',
+]);
 
 function clampPct(pct: number): number {
   if (pct < 0) return 0;
@@ -95,9 +123,12 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
       valueLabel,
       valueLabelHideDelay = 1500,
       className,
+      disabled,
+      disabledReason,
       onPointerDown,
       onFocus,
       onBlur,
+      onKeyDown,
       ...rest
     },
     ref,
@@ -124,6 +155,12 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
       flashLabel();
       onChange(event);
     }
+
+    // Soft-disable keeps the range focusable + hoverable (so the reason tooltip
+    // reaches pointer and keyboard users) while making it inert: the control is
+    // value-controlled, so swallowing onChange freezes the thumb against drag,
+    // and blocking the value keys stops keyboard nudges.
+    const softDisabled = Boolean(disabled) && disabledReason != null;
 
     const range = max - min;
     const pct = range > 0 ? ((value - min) / range) * 100 : 0;
@@ -183,33 +220,52 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
             </div>
           ) : null}
 
-          <input
-            ref={ref}
-            type="range"
-            min={min}
-            max={max}
-            step={step}
-            value={value}
-            onChange={handleChange}
-            onPointerDown={(event) => {
-              flashLabel();
-              onPointerDown?.(event);
-            }}
-            onFocus={(event) => {
-              flashLabel();
-              onFocus?.(event);
-            }}
-            onBlur={(event) => {
-              setShowLabel(false);
-              if (hideTimerRef.current) {
-                clearTimeout(hideTimerRef.current);
-                hideTimerRef.current = null;
-              }
-              onBlur?.(event);
-            }}
-            className={cn(inputClasses, className)}
-            {...rest}
-          />
+          <DisabledReasonTooltip reason={disabledReason} active={softDisabled}>
+            <input
+              ref={ref}
+              type="range"
+              min={min}
+              max={max}
+              step={step}
+              value={value}
+              disabled={softDisabled ? undefined : disabled}
+              aria-disabled={disabled || undefined}
+              onChange={softDisabled ? undefined : handleChange}
+              onPointerDown={(event) => {
+                if (softDisabled) {
+                  event.preventDefault();
+                  return;
+                }
+                flashLabel();
+                onPointerDown?.(event);
+              }}
+              onKeyDown={(event: KeyboardEvent<HTMLInputElement>) => {
+                if (softDisabled && VALUE_KEYS.has(event.key)) {
+                  event.preventDefault();
+                  return;
+                }
+                onKeyDown?.(event);
+              }}
+              onFocus={(event) => {
+                flashLabel();
+                onFocus?.(event);
+              }}
+              onBlur={(event) => {
+                setShowLabel(false);
+                if (hideTimerRef.current) {
+                  clearTimeout(hideTimerRef.current);
+                  hideTimerRef.current = null;
+                }
+                onBlur?.(event);
+              }}
+              className={cn(
+                inputClasses,
+                'aria-disabled:cursor-not-allowed aria-disabled:opacity-50',
+                className,
+              )}
+              {...rest}
+            />
+          </DisabledReasonTooltip>
         </div>
       </div>
     );
--- a/packages/ui/src/components/forms/slider.tsx
+++ b/packages/ui/src/components/forms/slider.tsx
@@ -229,7 +229,10 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
               step={step}
               value={value}
               disabled={softDisabled ? undefined : disabled}
-              aria-disabled={disabled || undefined}
+              // Only the soft-disabled branch needs `aria-disabled`; a natively
+              // `disabled` control already conveys the state, so emitting it
+              // there too would be redundant with the native attribute.
+              aria-disabled={softDisabled || undefined}
               // The range stays value-controlled while soft-disabled, so React
               // would warn about a `value` with no `onChange`. `readOnly`
               // silences that without changing behaviour — drag is already
--- a/packages/ui/src/components/forms/slider.tsx
+++ b/packages/ui/src/components/forms/slider.tsx
@@ -11,7 +11,10 @@ import {
 
 import { cn } from '../../lib/cn';
 import { SkeletonBox } from '../feedback/skeleton';
-import { DisabledReasonTooltip } from '../overlays/disabled-reason';
+import {
+  DisabledReasonTooltip,
+  hasDisabledReason,
+} from '../overlays/disabled-reason';
 
 export interface SliderProps extends Omit<
   InputHTMLAttributes<HTMLInputElement>,
@@ -160,7 +163,7 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
     // reaches pointer and keyboard users) while making it inert: the control is
     // value-controlled, so swallowing onChange freezes the thumb against drag,
     // and blocking the value keys stops keyboard nudges.
-    const softDisabled = Boolean(disabled) && disabledReason != null;
+    const softDisabled = Boolean(disabled) && hasDisabledReason(disabledReason);
 
     const range = max - min;
     const pct = range > 0 ? ((value - min) / range) * 100 : 0;
--- a/packages/ui/src/components/forms/checkbox.test.tsx
+++ b/packages/ui/src/components/forms/checkbox.test.tsx
@@ -1,7 +1,7 @@
-import { describe, it } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 
-import { checkAccessibility } from '@/tests/utils/a11y';
-import { render } from '@/tests/utils/render';
+import { checkAccessibility, expectFocusable } from '@/tests/utils/a11y';
+import { render, screen } from '@/tests/utils/render';
 
 import { Checkbox } from './checkbox';
 import { Label } from './label';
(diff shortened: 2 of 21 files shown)


larryro:
## Desk Review — #1949 `disabledReason` tooltip (PR #2130)

**Verdict: NOT READY — CHANGES REQUIRED.** The `Button`/`IconButton` affordance is well-built, correct, accessible, and well-tested — but the change covers only ~⅓ of the issue's stated scope. CI is green and all tests pass; the block is **incomplete scope**, not code defects.

### CI & tests (observed)
- **CI: all green.** `gh pr checks` reports ~40 checks **pass**, 0 failing/pending (Analyze, Build *, Lint, Format, Knip, Typecheck, Opengrep, Smoke, and Playwright platform 1–16/16 all `pass`).
- **Tests run locally:** `bunx vitest --run --project unit button.test.tsx icon-button.test.tsx` → **71 passed**. Full UI suite `bunx vitest --run --project unit` → **109 files, 1004 passed**. No regressions.

### What's good (no rework needed on the Button side)
- Soft-disable design is sound: disabled + `disabledReason` keeps the control focusable/hoverable, swaps native `disabled` for `aria-disabled`, blocks activation via `blockActivation`, and surfaces the reason through the shared Radix tooltip with `aria-describedby`. axe audit passes.
- Activation is genuinely inert on **all** paths — click, Enter, **and Space** (the synthesized click on Space keyup is caught by the `onClick={blockActivation}` path) and form-submit (preventDefault). Dropping the caller's `onClick`/`onKeyDown` while disabled is correct, native-equivalent behavior — **not** a bug.
- IconButton inherits `disabledReason` via `ButtonProps`; `disabledReason` correctly overrides the visible tooltip while disabled; `asChild`/Slot suppression is sound and documented. No DOM leak, no injection risk.

### Blocking points
1. **Form-field primitives were not touched.** The issue explicitly requires the affordance on *form-field primitives*, but `forms/input.tsx`, `forms/textarea.tsx`, `forms/checkbox.tsx`, and `forms/slider.tsx` have **0 changed lines** on this branch. Acceptance criterion "Every disabled **button/field/control** surfaces a reason" is unmet for fields. Add an equivalent disabled-reason affordance (tooltip + `aria-describedby`, keyboard-reachable) to these disabled-capable form primitives, with tests for the disabled+reason branch of each.
2. **The new affordance is defined but never applied to a real disabled control.** No production JSX passes the Button-primitive `disabledReason`. (Note: the existing `disabledReason` props in `chat-input.tsx`, `chat-interface.tsx`, `discussion-thread-view.tsx`, and `guardrails-overview.tsx` are a *pre-existing, unrelated* prop — an enum/string rendered as plain text — not this new tooltip affordance.) Acceptance criterion "*Every* disabled control surfaces a reason" cannot be satisfied while adoption is zero. Wire `disabledReason` into the real disabled Button/IconButton/field instances that are disabled for a known reason.

### Non-blocking notes
- Naming collision: the new `Button.disabledReason` (`ReactNode`) shares a name with a pre-existing `disabledReason` enum/string on other components — keep distinct when wiring real controls to avoid confusion.
- The branch carries an unrelated `style(workflow): apply oxfmt to desk-process.json` commit — scope creep for a UI PR; consider dropping it.
- Test coverage nits (not blocking): add explicit tests for Space-on-soft-disabled, `isLoading + disabledReason`, `asChild + disabledReason` (suppression), and assert `aria-describedby` is present on the focused soft-disabled button.

**NOT READY — CHANGES REQUIRED** (see blocking points 1–2 above).

larryro (commented):
## Desk review — `disabledReason` tooltip for disabled controls (#1949)

Reviewed the full implementation (not just the diff), set up the branch, and ran the project's own tests once.

**CI:** green — `gh pr checks` reports 47 passed, 3 skipping, 0 failing/pending.
**Tests I ran (this branch):**
- `bunx vitest --run --project unit` on the 6 changed UI files → **102 passed**.
- `bunx vitest --run wizard.test` (services/platform) → **13 passed**.

I confirmed each finding below independently (re-ran the exact path), including refuting one candidate blocker.

### ✅ Verified correct
- **Soft-disable blocks activation on every path.** I wrote a throwaway test putting a soft-disabled `Button` in a `<form>`: it does **not** submit on direct click **nor** on implicit submission (Enter in another field), while an enabled control test confirmed implicit submit does fire — so `blockActivation`'s `preventDefault` on the click event correctly cancels both. The default `type="submit"` is **not** a bug here. (A candidate "blocking" finding — refuted.)
- i18n: `common.actions.completeStepToContinue` is correctly placed in all four shipped locales (en/de/fr/de-CH), resolves via `tCommon = useT('common')`, passes the parity check, and the ß/ss split (de vs de-CH) is right.
- Wizard footer: reason is correctly suppressed while submitting (spinner explains), passed in both layouts, undefined-safe for callers that omit it.
- a11y basics: `aria-disabled` set, control kept focusable / in tab order, native `disabled` removed in soft mode, `asChild`/Slot suppression correct, icon-button accessible name preserved.

### 🔴 Blocking
**1. `Slider` soft-disabled emits a React console error on every use.** `packages/ui/src/components/forms/slider.tsx:230-233`: the controlled `<input type="range" value={value}>` gets `onChange={undefined}` while soft-disabled with **no** `readOnly`. I reproduced the warning:
> "You provided a `value` prop to a form field without an `onChange` handler … set either `onChange` or `readOnly`."
The sibling primitives (`input.tsx`, `textarea.tsx`) avoid this by setting `readOnly` while soft-disabled; the slider deviates and is the only one that warns. The value is still correctly frozen, but this is a confirmed defect + dev-console spam, uncaught by any test. Fix: set `readOnly={softDisabled ? true : undefined}` on the range input (it's ignored by the browser for drag, which is already blocked via `onPointerDown`/key handling, but it silences React's warning) — or pass a no-op `onChange` when soft-disabled. Add a `vi.spyOn(console,'error')` assertion so it can't regress.

### 🟡 Recommended (non-blocking)
2. **Form-submission semantics change in `Input`/`Textarea`.** Swapping native `disabled`→`readOnly` (input.tsx / textarea.tsx) flips the field from *excluded* from form submission to *included*. No product code passes `disabledReason` to these yet, so there's no live regression — but it's a latent footgun in a shared primitive. Worth a deliberate decision/doc note (or omit the field from submission another way).
3. **Screen-reader browse mode.** The reason is exposed via Radix tooltip `aria-describedby`, which only exists while the tooltip is open (hover/focus). A browse-mode SR user who moves the virtual cursor without focusing won't hear the reason. Keyboard-focus users do get it (acceptance criterion met). Consider an always-on visually-hidden description if broader SR support is intended.
4. **Test coverage gaps:** hover (only focus is tested), `aria-describedby` wiring (asserted for Button only, not the shared `DisabledReasonTooltip`/form controls), slider drag-blocking, and behavioral edit-blocking for input/textarea (only the `readonly` attribute is asserted).
5. **Nits:** `checkbox.tsx:47` comment says "Space activates a checkbox" but the code also blocks Enter (and the Enter-block is dead weight — activation is already stopped via `onClick`). `Button` doesn't reuse the shared `DisabledReasonTooltip` (inlines its own Provider/Trigger) — defensible since its tooltip is multi-purpose, but a small shared shell would de-dupe. `blockActivation` is duplicated in button/checkbox.

### Scope
The PR adds the shared affordance to Button/IconButton + Input/Textarea/Checkbox/Slider and adopts it in the onboarding wizard. The library's other disabled-capable primitives (`navigation/tabs.tsx`, `overlays/dropdown-menu.tsx` items) are not covered, and product adoption is the single wizard flow. Reasonable as a first increment delivering the mechanism; full rollout to "every disabled control" (per the issue wording) would be follow-up.

**Verdict:** NOT READY — fix the `Slider` React-warning/`readOnly` inconsistency (#1), then it's in good shape.

packages/ui/src/components/forms/slider.tsx
```diff
@@ -183,33 +220,52 @@ const SliderBase = forwardRef<HTMLInputElement, SliderProps>(
...
+            <input
+              ref={ref}
+              type="range"
+              min={min}
+              max={max}
+              step={step}
+              value={value}
+              disabled={softDisabled ? undefined : disabled}
+              aria-disabled={disabled || undefined}
+              onChange={softDisabled ? undefined : handleChange}
```
larryro, line 245: **Blocking:** soft-disabled slider emits a React console error.

Here the controlled `<input type="range" value={value}>` gets `onChange={undefined}` with no `readOnly`, so React logs:

> You provided a `value` prop to a form field without an `onChange` handler … set either `onChange` or `readOnly`.

(Reproduced locally.) `input.tsx`/`textarea.tsx` avoid this by setting `readOnly` while soft-disabled — the slider should match: `readOnly={softDisabled ? true : undefined}` on the range input (browser ignores it for drag, which is already blocked; it just silences React). Add a `console.error` spy so it can't regress.

larryro:
## Desk Review — #1949 disabledReason tooltips

**Verdict: NOT READY — one CI gate is red (the code itself meets the bar).**

### CI state (`gh pr checks`)
- **Validate images — FAIL**: `sandbox-runtime: 3114 MB exceeds 2800 MB budget`. This PR changes **no** Docker/sandbox files, and `main` is currently red on the same check (introduced by the unrelated `feat(sandbox): bake document toolchain into runtime image`). So this is a pre-existing repo-level failure, **not** caused by this PR — but it is still a red required check, so merge is gated until it's green (raise the budget or slim the sandbox-runtime image at the repo level; rebasing alone won't help while main is red).
- All other checks pass: Type check, Lint, Unit, UI, Storybook, all 16 Playwright (platform) shards, Knip, Format, Opengrep, Build *, Smoke test, etc.

### Tests (run locally)
`cd packages/ui && bunx vitest --run --project unit` → **1029 passed / 109 files**, incl. axe audits and the new `disabledReason` suites for Button, IconButton, Checkbox, Input, Textarea, Slider.

### What's good
- Sound soft-disable pattern: when (and only when) `disabled` **and** `disabledReason` are set, the control drops native `disabled`, gains `aria-disabled`, stays focusable/hoverable, and blocks activation (click/Space/Enter for button & checkbox; drag + value-keys + swallowed `onChange` + `readOnly` for slider; `readOnly` for input/textarea). Plain `disabled` (no reason) behavior is unchanged.
- Keyboard a11y: tooltip opens on focus and Radix wires `aria-describedby` while open (covered by `button.test.tsx`). axe passes for every soft-disabled variant.
- Primitive coverage is complete (button, checkbox, slider, input, textarea); i18n key `common.actions.completeStepToContinue` resolves in en/de/fr/de-CH; wizard adoption correctly suppresses the reason while `submitting` (spinner already explains).

### Non-blocking notes (nice-to-have, not required for merge)
1. `ButtonBase` sets no default `type`, so a soft-disabled `<Button>` inside a `<form>` is an implicit submit button. Submission is currently prevented by `blockActivation`'s `preventDefault` on the click, but a default `type="button"` would be more robust defense-in-depth.
2. `aria-disabled` is emitted even on natively-`disabled` fields (when no reason), which is redundant with the native attribute. Harmless (axe passes); consider gating it on the soft-disabled branch only.
3. Duplication: `Button` inlines the Tooltip Provider/Root/Trigger/Portal structure rather than reusing `DisabledReasonTooltip`. Intentional (Button blends tooltip vs disabledReason), but a shared helper would keep `delayDuration`/structure in one place.
4. `readOnly` is a no-op on non-text input types (`range` is handled separately; `color`/`file` etc. wouldn't be inert). No current consumer hits this, but worth a guard/doc if `Input` is later used with such types.
5. Soft-disabled user-supplied `onClick`/`onKeyDown` (and slider `onKeyDown` for value keys) are dropped, not composed — intended for inert controls and documented, just noting the behavior.

None of (1)–(5) block merge. The only gate is the red **Validate images** check above.

larryro:
## Desk review — `disabledReason` tooltips for disabled controls (#1949)

**Verdict: NOT READY — CHANGES REQUIRED** (the implementation is correct; the blockers below are one small code fix + targeted test coverage, not a redesign.)

### What I verified
- **CI: green** — 47 checks pass, 3 skipped, 0 failed/pending (`gh pr checks`).
- **Tests: pass** — ran the project's own suites once on this branch:
  - `bunx vitest --run --project unit <changed ui files>` → **105 passed** (checkbox/input/textarea/slider/button/icon-button).
  - `bunx vitest --run --config vitest.ui.config.ts wizard.test.tsx` → **9 passed**.
- **Soft-disable correctness** (Button/IconButton + Input/Textarea/Checkbox/Slider): walked every branch (enabled / native-disabled / soft-disabled / isLoading / asChild / type override). Activation is correctly blocked (click + Space/Enter for buttons & checkbox; `readOnly` + `onPointerDown`/`onKeyDown` guards for slider; `readOnly` for text fields), `disabled`→`aria-disabled` swap is correct, `type` defaults to `button` under soft-disable, and `disabledReason` does not leak to the DOM. No correctness defect found.
- **i18n: complete** — `common.actions.completeStepToContinue` present in all four user-facing locales (en/de/de-CH/fr) at the path the code reads. (`global.json` is a separate namespace.)
- **Form-submission semantic** — soft-disabled fields use `readOnly` (not native `disabled`), so they would be *included* in a `<form>` submission. Confirmed **dormant**: no form field (Input/Textarea/Checkbox/Slider) currently receives `disabledReason`; the only live consumer is the wizard's Next **button**. Not an active bug — see observation (B).

### Blocking changes required
1. **Empty / whitespace `disabledReason` produces an unexplained inert control.** Confirmed at runtime: `<Input disabled disabledReason="">` renders `disabled=false`, `aria-disabled="true"`, `readonly`, with an **empty tooltip and empty `aria-describedby` text**. The control is removed-from-native-disable and made inert/focusable but conveys *no reason* — strictly worse than a native disable and contrary to the feature's purpose. Affects the `softDisabled` gate in all five primitives (`forms/input.tsx`, `forms/textarea.tsx`, `forms/checkbox.tsx`, `forms/slider.tsx`, and the wrapper gate in `primitives/button.tsx`), which all use `disabledReason != null`. **Fix:** treat an empty/whitespace-only string the same as "no reason" (fall back to native disabled), so a control is only soft-disabled when there is a real reason to show.
2. **Slider — core inert behaviors are untested** (`forms/slider.test.tsx`). The soft-disabled slider only has tests for `ArrowRight`/`ArrowLeft`. Add tests that the value does **not** change for the other `VALUE_KEYS` — `Home`, `End`, `PageUp`, `PageDown` — and that **pointer drag is blocked** (`onPointerDown` preventDefault). Drag is the slider's primary interaction and is the central promise of soft-disable, yet has zero coverage.
3. **Checkbox — asymmetric/missing coverage** (`forms/checkbox.test.tsx`). Add: (a) `Enter` is blocked (source blocks Space **and** Enter; only Space is tested); (b) "ignores the reason while enabled" (present for input/textarea/slider/button, missing here); (c) `aria-describedby` wires the reason on focus (present for button, missing for the form fields).
4. **IconButton — thin coverage** (`primitives/icon-button.test.tsx`). Only one `disabledReason` test exists. Add parity with Button for at least: native-disabled (no reason) keeps native `disabled`, and activation is blocked (click + Space/Enter).

### Non-blocking observations (recommended, not required for merge)
- **(A) Scope vs. acceptance criterion.** The issue's *Proposed change* names "Button, IconButton, and form-field primitives" — all delivered. But the broader criterion ("apply to **all** controls with a disabled state") leaves `navigation/tabs.tsx` and `overlays/dropdown-menu.tsx` (both have `disabled` props) without a `disabledReason`. Reasonable as a follow-up; flagging so it's a conscious decision.
- **(B) Document the `readOnly` submission semantic.** Since soft-disabled form fields submit their value inside a `<form>` (unlike native `disabled`), add a one-line caveat to the form-field JSDoc so a future caller isn't surprised.
- **(C) Dead param.** `DisabledReasonTooltip`'s `side` prop is never passed by any caller — remove or document.
- **(D) Nit.** A natively-disabled `Button` emits both `disabled` and a redundant `aria-disabled="true"`, while the form fields (correctly) omit `aria-disabled` on native disable. Pre-existing for Button and axe-clean; consider aligning for consistency.

**Verdict: NOT READY — CHANGES REQUIRED** (blocking items 1–4 above).

larryro (commented):
## Desk review — `disabledReason` soft-disable affordance (#1949)

**Verdict: READY TO MERGE.** Reviewed the whole implementation (not just the diff) across two rounds with fan-out review subagents, then independently re-verified every candidate blocker. No blocking defects found.

### Evidence
- **CI:** all checks green on the latest run — `gh pr checks` reports `pass` for all 40+ jobs (Analyze JS/Python, Browser, all Build *, Lint, Format, Knip, Opengrep, Migrations, and all 16 Playwright platform shards + web/docs).
- **Tests (run locally on the branch):**
  - `cd packages/ui && bunx vitest --run --project unit` → **109 files, 1048 tests passed**.
  - `cd services/platform && bunx vitest --run --config vitest.ui.config.ts app/components/ui/wizard/wizard.test.tsx` → **9 passed**.

### What's correct
- **Soft-disable pattern** (Button/IconButton/Input/Textarea/Checkbox/Slider): when `disabled` + a meaningful `disabledReason`, the native `disabled` is swapped for `aria-disabled`, the control stays focusable/hoverable, activation is blocked, and Radix wires `aria-describedby` on focus — so the reason reaches pointer **and** keyboard/SR users. Walked every branch; no half-states.
- **`hasDisabledReason` gate** correctly falls back to native disable for null/boolean/empty/whitespace reasons (verified the empty-string and whitespace branches are tested).
- **Checkbox** `preventDefault` genuinely stops Radix's composed toggle (confirmed against `@radix-ui/react-checkbox` source — toggle lives in the composed `onClick`, and `composeEventHandlers` skips on `defaultPrevented`); works in controlled and uncontrolled modes.
- **Slider** freeze holds: the range is value-controlled with `onChange=undefined` while soft-disabled, so drag, arrow/Home/End/PageUp/PageDown keys (blocked via `VALUE_KEYS`), track-click and wheel can never commit a value change — React reverts the DOM value. `readOnly` silences the controlled-without-onChange warning.
- **Wizard adoption** is gated correctly; `aria-disabled` mode switching is purely prop-derived (no cached state), so toggling `disabled`/`reason` at runtime can't strand a stuck `aria-disabled` or stuck tooltip.
- **i18n:** `common.actions.completeStepToContinue` present in every loaded locale (en/de/de-CH/fr); no missing-locale fallback. `disabledReason` only ever receives static/developer strings — no injection surface.

### Non-blocking notes (optional follow-ups, do not block merge)
1. **Slider (`slider.tsx`):** mouse-wheel and track-click can momentarily move the thumb (visual flicker) before the controlled value reverts. Committed value is never affected. An `onWheel` preventDefault when soft-disabled would give pointer parity with the keyboard guard. Also, the comments at ~`162-165`/`239-243` credit `onPointerDown` with "freezing the thumb" — the real freeze is the controlled value; consider tightening the wording.
2. **Wizard (`wizard-footer.tsx:66`):** the `!submitting` term in `primaryDisabledReason` is defensive — because `goNext` (`wizard.tsx:148`) only sets `submitting` after the step is valid, `stepInvalid && submitting` is unreachable, so the guard never changes the outcome. Harmless; an explicit test isn't required for an unreachable state, but the guard could be simplified or annotated.
3. **`disabled-reason.tsx`:** `hasDisabledReason` returns `true` for empty-but-truthy nodes (`[]`, `<></>`, `<span/>`), which could in theory soft-disable with a blank tooltip. No caller does this (all pass strings); a one-line doc caveat would close the gap.
4. **Minor elegance:** `blockActivation` is duplicated in `button.tsx` and `checkbox.tsx`; `soft` (base) vs `softDisabled`/`showDisabledReason` (wrapper) in `button.tsx` are overlapping-but-not-equal predicates worth a one-line comment. Input/Textarea tests assert the `readonly` attribute rather than observing a blocked `user.type`.

None of these affect correctness, accessibility, security, or CI. **READY TO MERGE.**

PR #2140, opened 2026-06-25
fix: export-chat "Deselect all" no-op and dead download-disabled state (#2080)

### LINKED ISSUE #2080: Bug: Export-chat dialog "Deselect all" is a no-op (empty-set sentinel re-selects everything); download-disabled state is dead code
**Area / route**
Chat → Export. `services/platform/app/features/chat/components/export-chat-dialog.tsx` (opened from a thread's More actions → Export).

**Steps to reproduce**
1. Open a thread that has messages.
2. More actions → Export to open the **Export chat** dialog.
3. Click **Deselect all**. — OR — Deselect per-message checkboxes one at a time, all the way down to the last remaining message.

**Expected**
Per manual test `services/platform/tests/manual/chat.md` F15: "**Deselect all** clears the per-message checkboxes." Once nothing is selected, **Download Markdown** / **Download PDF** disable (their `disabled={noneSelected}` guard fires), and the counter reads `0 of N messages selected`. Reaching zero selected should stay at zero.

**Actual**
Nothing clears. `effectiveSelected` treats an empty `selectedIds` Set as a sentinel meaning "default = all selected", so the moment the selection becomes empty it instantly resolves back to `allIds`:
- Clicking **Deselect all** runs `setSelectedIds(new Set())`, which re-resolves to `allIds` on the next render — every checkbox stays checked, the toggle stays "Deselect all", and the counter stays `N of N`.
- Deselecting messages one-by-one works until the last remaining one; removing it empties `selectedIds`, which re-triggers the sentinel and re-checks every message. The selection can never reach zero.

Because `effectiveSelected` can never be empty while messages exist, `noneSelected` is always `false`, so the `disabled={noneSelected}` guards on both download buttons are dead code (the buttons never disable).

**Evidence** (confirmed file:line — `export-chat-dialog.tsx`)
- Empty-set sentinel: line 171 — `if (selectedIds.size === 0 && messages.length > 0) { return allIds; }` inside `effectiveSelected` (170-180).
- Deselect all no-op: lines 185-191, `handleToggleAll` does `setSelectedIds(new Set())` when `allSelected`.
- Last-message re-select: lines 193-209, `handleToggleMessage` seeds from all when `prev` is empty, but deleting the final id empties `prev` → next render hits the sentinel.
- Dead disabled state: line 183 `const noneSelected = effectiveSelected.size === 0;` is unreachable while messages exist; consumed at lines 253 (`disabled={noneSelected}` on Download Markdown) and 262 (Download PDF).
- Expectation source: `services/platform/tests/manual/chat.md:80` (F15) — "Deselect all clears the per-message checkboxes".

**Notes**
- Distinct from #1973, whose documented root cause is the invalid nested `<button>` hydration warning (chat.md:140); that is already mitigated by the `pointer-events-none` Checkbox wrapper with `tabIndex={-1}` at lines 292-297. This is a separate selection-state logic defect.
- The existing component test `export-chat-dialog.test.tsx` only asserts the toggle label reads "Deselect all" when all are selected (lines 69-72); it never clicks Deselect all nor deselects down to zero, so the bug is unguarded. A fix should add a regression test that clicks Deselect all and asserts `0 of N` + both download buttons disabled.
- Suggested fix: represent "all selected" with an explicit non-empty state (e.g. initialize `selectedIds` to `allIds` once messages load, or use a discriminated `'all' | Set` selection model) so an empty set genuinely means zero selected instead of being overloaded as a default-all sentinel.

---

larryro:
## Summary

Fixes #2080.

The export-chat dialog used an **empty `Set` as a sentinel meaning "default = all selected"**. As a result:
- **Deselect all** was a no-op — `setSelectedIds(new Set())` instantly re-resolved to `allIds` on the next render.
- Deselecting messages one-by-one never reached zero — removing the last id emptied the set, re-triggering the sentinel.
- `noneSelected` was therefore always `false`, making the `disabled={noneSelected}` guards on both download buttons dead code.

## Fix

Replaced the overloaded empty-set sentinel with an explicit discriminated selection model: `selection` is either the `'all'` sentinel (default, selects every message including ones that load later) or an explicit `Set<string>`. An explicit **empty** set now genuinely means "nothing selected", so:
- **Deselect all** sets `selection = new Set()` → counter reads `0 of N`, toggle flips to **Select all**, both download buttons disable.
- Deselecting the final message empties the set instead of re-selecting everything.

## Tests

Added two regression tests in `export-chat-dialog.test.tsx`:
- Clicking **Deselect all** → `0 of N`, no checkboxes checked, both downloads disabled.
- Deselecting messages one-by-one reaches zero (final message no longer re-selects all).

All 5 UI tests pass; `tsc --noEmit` and `oxlint --type-aware` are clean on the changed files.

### DIFF
--- a/services/platform/app/features/chat/components/export-chat-dialog.test.tsx
+++ b/services/platform/app/features/chat/components/export-chat-dialog.test.tsx
@@ -116,4 +116,66 @@ describe('ExportChatDialog', () => {
       mockMessages.length,
     );
   });
+
+  it('clears all checkboxes and disables downloads when Deselect all is clicked', async () => {
+    const { user } = render(
+      <ExportChatDialog
+        open={true}
+        onOpenChange={vi.fn()}
+        threadId="thread-1"
+        organizationId="org-1"
+      />,
+    );
+
+    // Starts with everything selected.
+    expect(
+      screen.getByText(
+        `${mockMessages.length} of ${mockMessages.length} messages selected`,
+      ),
+    ).toBeInTheDocument();
+
+    await user.click(screen.getByRole('button', { name: 'Deselect all' }));
+
+    // Selection genuinely reaches zero and stays there.
+    expect(
+      screen.getByText(`0 of ${mockMessages.length} messages selected`),
+    ).toBeInTheDocument();
+    // The toggle flips to "Select all"; no checkbox remains checked.
+    expect(
+      screen.getByRole('button', { name: 'Select all' }),
+    ).toBeInTheDocument();
+    for (const checkbox of screen.getAllByRole('checkbox')) {
+      expect(checkbox).not.toBeChecked();
+    }
+    // Both download buttons disable while nothing is selected.
+    expect(
+      screen.getByRole('button', { name: 'Download Markdown' }),
+    ).toBeDisabled();
+    expect(screen.getByRole('button', { name: 'Download PDF' })).toBeDisabled();
+  });
+
+  it('reaches zero when messages are deselected one by one', async () => {
+    const { user } = render(
+      <ExportChatDialog
+        open={true}
+        onOpenChange={vi.fn()}
+        threadId="thread-1"
+        organizationId="org-1"
+      />,
+    );
+
+    const checkboxes = screen.getAllByRole('checkbox');
+    for (const checkbox of checkboxes) {
+      await user.click(checkbox);
+    }
+
+    // Removing the final message empties the selection instead of re-selecting.
+    expect(
+      screen.getByText(`0 of ${mockMessages.length} messages selected`),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole('button', { name: 'Download Markdown' }),
+    ).toBeDisabled();
+    expect(screen.getByRole('button', { name: 'Download PDF' })).toBeDisabled();
+  });
 });
--- a/services/platform/app/features/chat/components/export-chat-dialog.tsx
+++ b/services/platform/app/features/chat/components/export-chat-dialog.tsx
@@ -165,49 +165,51 @@ function ExportChatDialogContent({
     [data?.messages],
   );
 
-  const [selectedIds, setSelectedIds] = useState<Set<string>>(() => new Set());
+  // Selection is either the `'all'` sentinel (the default before the user
+  // touches anything — every message selected, including ones that load later)
+  // or an explicit Set. An explicit empty Set genuinely means "nothing
+  // selected", so deselecting down to zero stays at zero.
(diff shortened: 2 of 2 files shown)


larryro:

Reviewed the fix for #2080 on branch `tale/xs764ehrzrdnft95849dd2x4j1898pm3` using a two-round, fan-out method (5 dimension reviewers + adversarial confirmation).

### CI & tests
- **CI: all checks green** (Type check, Lint, Format, Knip, UI, Unit, Storybook, Playwright ×16, builds, scans — every required check passes; only fork-only/optional jobs skipped).
- **Tests run locally:** `bunx vitest --run --config vitest.ui.config.ts app/features/chat/components/export-chat-dialog.test.tsx` → **5 passed (5)**.
- **Regression genuineness confirmed:** ran the two new tests against the *pre-fix* (`origin/main`) implementation — both fail ("2 of 2" instead of "0 of 2"), proving they pin the exact bug. They pass on the fix.

### Assessment by dimension
- **Correctness:** The `selection: 'all' | Set<string>` discriminated model cleanly separates "default-all (incl. later-loaded messages)" from "explicit empty = nothing". All branches traced — `'all'` default, Deselect-all → empty Set, one-by-one to zero, individual toggle from `'all'` vs explicit Set, stale-ID pruning, and `messages.length === 0`. Both reported sub-bugs are resolved: Deselect-all genuinely reaches `0 of N`, and `disabled={noneSelected}` is now live, not dead.
- **Robustness:** `useMemo`/`useCallback` dependency arrays are tightened correctly (`[selection, allIds]`, `[allSelected]`, `[allIds]`); functional state updaters avoid stale closures; no leaks introduced. Async message loading is handled — explicit selections persist and stale IDs are filtered.
- **Elegance:** Minimal, focused change; old empty-set sentinel fully removed (no dead code); accurate explanatory comment; clear rename `selectedIds → selection`. oxlint + tsc clean on the changed file.
- **Tests:** Two added regression tests cover both bug facets and correctly assert disabled download buttons. Untested branches (PDF download path, select-all-after-deselect, empty dialog) are low-risk nits given symmetric logic and the dual guard (`disabled` + `if (selected.length === 0) return`).
- **Boundaries:** Export handlers are double-guarded against empty selection; `getSelectedMessages` filters strictly by `effectiveSelected` membership, so no stale/unselected message can be exported.

### Adversarial pass
A skeptic agent tried 13 interaction sequences (deselect→toggle→select-all→deselect, async load races, stale-id size collisions, disabled-state inversions) and could not produce a wrong selection, counter, button state, or export.

**Verdict: READY TO MERGE.**

PR #2142, opened 2026-06-25
fix(platform): auto-retry convex deploy on transient module-analyze timeout (#2117)

### LINKED ISSUE #2117: Bug: tale deploy hard-fails on transient Convex module-analyze timeout instead of auto-retrying
## Summary

`tale deploy` hard-fails (exit 1, no retry) when the Convex push hits a **module-analyze timeout** — even though the condition is transient and a plain re-run succeeds with zero code changes. The deploy entrypoint classifies this error as `unclassified` and skips the auto-retry path that every other transient failure already uses.

## Observed error

During a normal `tale deploy`, the `platform-blue` color aborted on the Convex push step:

```
platform-blue  ✖ Error: Unable to start push to http://convex:3210
platform-blue  ✖ Error fetching POST http://convex:3210/api/deploy2/evaluate_push 400 Bad Request:
               InvalidModules: Hit an error while pushing:
               Loading the pushed modules encountered the following error:
               Function execution timed out (maximum duration: 2s)
platform-blue  ERROR Convex deploy failed (exit code: 1)
platform-blue  ━━━ Error diagnosis ━━━
platform-blue  Reason: unclassified. See full deploy log above.
               fix: try RUST_LOG=debug on the convex service and re-run deploy.
```

Failure occurred **before** the blue-green traffic switch, so the live color kept serving (no rollback needed). **Re-running `tale deploy` immediately succeeded with no code change.**

## Root cause

- `…/api/deploy2/evaluate_push … Function execution timed out (maximum duration: 2s)` is Convex's **analyze step**: on push, the backend imports each function module once in a V8 isolate to discover its exports, under a hard ~2s per-module evaluation budget.
- This is **not** the runtime/action timeout (`ACTIONS_USER_TIMEOUT_SECS=1800`) nor the deploy wall-clock (`CONVEX_DEPLOY_TIMEOUT=300`). The 2s analyze budget is baked into the pinned convex-backend image and is not configurable from this repo.
- Investigation found **no top-level `await`** and **no genuinely 2s-heavy module-scope work** anywhere under `services/platform/convex/` — the heaviest top-level costs (a single `renderPrompt`, a few compiled regexes, auth role construction) are all sub-millisecond. None of them is a deterministic 2s offender, and none changed in the deployed version.
- The convex container has **no CPU limit** (`compose.yml`, `mem_limit: 12g` only). During a deploy it competes for CPU with concurrent image pulls and the in-flight chat-generation drain. Under that contention a module that normally analyzes in ~1.5s tips over 2s — which is exactly why it is non-deterministic and why an immediate re-run (images already local, less contention) succeeds.

## Why this is a bug

In `services/platform/docker-entrypoint.sh`, the deploy error classifier routes genuinely-transient failures (`RaceDetected`, `SearchIndexesUnavailable`, `fetch failed`/`ECONNREFUSED`) to `retry=true`, which then runs a 3× exponential-backoff retry loop (`services/platform/docker-entrypoint.sh:485-508`).

The analyze timeout instead falls through to the final `else` branch (`services/platform/docker-entrypoint.sh:477-480`):

```bash
else
  log_error "Reason: unclassified. See full deploy log above."
  echo "  fix: try RUST_LOG=debug on the convex service and re-run deploy."
fi
```

`retry` stays `false`, so the deploy **exits 1 immediately** and forces a manual re-run — for a condition we've shown is transient and self-healing. This makes deploys flaky for operators with no upside.

## Proposed fix

Classify the analyze timeout as retryable so it reuses the existing backoff loop, mirroring the `SearchIndexesUnavailable` precedent (`services/platform/docker-entrypoint.sh:453-456`). Add a branch **before** the final `else` in `deploy_convex_functions`:

```bash
elif grep -q "InvalidModules" "$deploy_log" && grep -q "Function execution timed out" "$deploy_log"; then
  log_error "Reason: module analyze timed out (2s isolate budget) — usually transient deploy-time CPU contention"
  echo "  → A module's import-time evaluation exceeded the 2s analyze budget,"
  echo "    typically because convex was CPU-starved during image pulls / chat drain."
  echo "  fix: auto-retrying; if persistent, set RUST_LOG=debug to find the slow module and defer its top-level work."
  retry=true
```

Matching `InvalidModules` **and** `Function execution timed out` together is specific to the push-analyze timeout and won't catch unrelated runtime timeouts.

## Secondary hardening (optional, not required)

- Give the convex container some CPU priority during the deploy window (it currently has no CPU reservation), or stagger image pulls vs. the push, to reduce how often the 2s budget is tipped in the first place.
- The auto-retry above is the primary fix; this is belt-and-suspenders.

## Acceptance criteria

- A Convex push that fails with `InvalidModules` + `Function execution timed out (maximum duration: 2s)` is logged as a transient/retryable reason and auto-retried up to 3× with backoff, instead of exiting 1.
- A deploy that previously needed a manual re-run now self-heals on retry.
- Non-transient push failures (`InvalidSchema`, `ModulesTooLarge`, `AuthConfigMissingEnvironmentVariable`, etc.) still fail fast — no change to their behavior.

## References

- `services/platform/docker-entrypoint.sh:411-512` — push error classifier + retry loop
- `services/platform/docker-entrypoint.sh:477-480` — current `unclassified` fallthrough (no retry)
- `services/platform/docker-entrypoint.sh:453-456` — `SearchIndexesUnavailable` retryable precedent
- `compose.yml` (convex service) — `mem_limit: 12g`, no CPU limit, `env_file: - .env`

larryro:
## Summary

Fixes #2117.

`tale deploy` hard-failed (exit 1, no retry) when the Convex push hit a **module-analyze timeout** (`InvalidModules` + `Function execution timed out (maximum duration: 2s)`), even though the condition is transient — an immediate re-run succeeds with zero code changes. The deploy error classifier in `services/platform/docker-entrypoint.sh` routed this error to the final `unclassified` `else` branch, which leaves `retry=false` and exits 1.

## Change

Add a classification branch in `deploy_convex_functions`, before the final `else`, that matches the push-analyze timeout (both `InvalidModules` **and** `Function execution timed out` present) and sets `retry=true`, so it reuses the existing 3× exponential-backoff retry loop — mirroring the `SearchIndexesUnavailable` retryable precedent.

Matching both strings together keeps this specific to the push-analyze timeout and avoids catching unrelated runtime/action timeouts. Non-transient push failures (`InvalidSchema`, `ModulesTooLarge`, `AuthConfigMissingEnvironmentVariable`, …) still fail fast — unchanged.

## Acceptance criteria

- ✅ A Convex push failing with `InvalidModules` + `Function execution timed out (maximum duration: 2s)` is logged as transient/retryable and auto-retried up to 3× with backoff instead of exiting 1.
- ✅ A deploy that previously needed a manual re-run now self-heals on retry.
- ✅ Non-transient push failures still fail fast — no behavior change.


larryro (commented):
## Desk Review — #2117 auto-retry convex deploy on transient module-analyze timeout

**Verdict: READY TO MERGE.** ✅

Reviewed the whole `deploy_convex_functions` classification chain in `services/platform/docker-entrypoint.sh`, not just the diff. The 7-line addition is correct, minimal, and idiomatic.

### CI
All required checks **pass** (Type check, Lint, Format, Knip, Unit, Migrations, Build *, Playwright *, Opengrep, Analyze). Only "skipping" entries are `Smoke test (fork PR)` and `Docs container test`, which are expected skips, not failures. CI confirmed green.

### Verification performed
- `bash -n services/platform/docker-entrypoint.sh` → syntax OK.
- Re-ran each earlier branch's grep pattern against the real error text `InvalidModules: Hit an error while pushing: Function execution timed out (maximum duration: 2s)`: **none** of the earlier branches match; the new branch is the **only** match → reachable. The internal 2s isolate timeout surfaces as a non-124 exit, so the `deploy_exit -eq 124` branch does not intercept it.
- No project test harness exists for any `docker-entrypoint.sh` (no `*.bats`, no shellcheck in CI, no entrypoint script is tested). Absence of a unit test here is **convention-consistent**, not a gap. The log-output shape is already covered generically by `platform-container.ts` classifier (keys off the `Reason:`/severity vocabulary, not specific strings) — no classifier change needed.

### What's correct
- Matches `InvalidModules` **and** `Function execution timed out` together → tight signature for the push-analyze timeout; sets `retry=true`, reusing the existing bounded 3× backoff loop (10→20→40s). Mirrors the `SearchIndexesUnavailable` retryable precedent.
- Fail-fast preserved for non-transient push errors (`InvalidSchema`, `ModulesTooLarge`, `AuthConfigMissingEnvironmentVariable`) — they sit earlier in the `elif` chain with no `retry=true`.
- A persistent (non-transient) module error still terminates with `exit 1` after 3 retries — bounded added latency (~70s), no masked success, no infinite loop. The branch's own `fix:` message points at `RUST_LOG=debug`.
- No security/input-boundary concern: greps run over an internal `mktemp` deploy log, never user input; patterns are constant literals.

### Non-blocking observations (optional polish, not required to merge)
1. The two `grep -q` calls scan the whole accumulated log independently, so in theory `InvalidModules` and `Function execution timed out` could come from different lines (e.g. across retry attempts appended via `tee -a`). Worst case is a few wasted retries before `exit 1`. A single same-line pattern (`grep -q "InvalidModules:.*Function execution timed out"`) would be tighter. Low impact.
2. `SEARCH_INDEX_RETRY_BACKOFF` is not declared `local` and never reset. Harmless today (function is called exactly once per process and `exit 1`s; the search-index and module-analyze branches are mutually exclusive), but declaring it `local` would make the new branch's documented "default 10s" backoff robust-by-construction rather than incidental, if the function is ever called twice in one process.
3. Pure style nit: the branch wraps one sentence across two `echo` lines with a trailing comma, unlike the one-thought-per-line siblings.

None of the above blocks the merge.

larryro:
## Desk Review — #2117 auto-retry convex deploy on transient module-analyze timeout

**Verdict: READY TO MERGE.** ✅

Reviewed the whole `deploy_convex_functions` classification chain in `services/platform/docker-entrypoint.sh` (not just the diff), via breadth review across correctness / robustness / elegance / tests / boundaries, then independently re-verified every potential blocker. The 11-line change (one new `elif` branch + one `local` declaration, one file) is correct, minimal, and idiomatic. It also already incorporates the prior review's two asks (single pattern instead of two `grep -q`; `SEARCH_INDEX_RETRY_BACKOFF` made `local`).

### CI
`gh pr checks` → **46 pass, 5 expected skips (fork-PR / docs / web container / trivy), 0 fail, 0 pending.** `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. CI confirmed green.

### Tests
- Project test suites ran green in CI (Unit, Migrations, UI, Browser, Smoke test, 16× Playwright, Type check, Lint, Format, Knip).
- No shell/`bats` test harness exists for any `docker-entrypoint.sh`, and the entrypoint isn't shellcheck'd in CI — so the absence of a per-branch unit test is **convention-consistent**, not a gap. The new `Reason: …` log line is already covered generically by `packages/shared/src/classify/sources/platform-container.ts` (keys off the `Reason:` prefix, not the specific string; tested in `platform-container.test.ts`), so no classifier change is needed.

### What I verified (with GNU grep 3.8 — the production family; note plain `grep` ≠ GNU in some shells where `-z` means *decompress*)
- **Matches the reported error.** `grep -qz "InvalidModules.*Function execution timed out"` matches the exact multi-line error from #2117 (`InvalidModules:` … `Function execution timed out (maximum duration: 2s)` on separate lines). Crucially, `-z` is **required** — the same pattern *without* `-z` does **not** match across newlines, so a plain same-line `grep -q` would have silently failed to fix the bug. The implementer's `-z` choice is the correct one.
- **Branch is reachable.** None of the nine earlier `elif` patterns (RaceDetected, ConcurrentPush, ModulesTooLarge, InvalidSchema, Text/VectorIndexTooLarge, Search/VectorIndexesUnavailable, AuthConfigMissingEnvironmentVariable, fetch failed/ECONNREFUSED, invalid admin key/unauthorized) match the analyze-timeout text; the `deploy_exit -eq 124` first branch keys off exit code, not the log, and the analyze timeout surfaces as a non-124 exit. The new branch is the only match.
- **Retry path is sound & bounded.** `retry=true` reuses the existing 3× exponential-backoff loop (10→20→40s, capped 90s). A genuinely non-transient failure still terminates with `exit 1` after 3 attempts — no masked success, no infinite loop. `rm -f "$deploy_log"` runs on every exit path (no temp leak).
- **No regression to the search-index path.** Adding `local SEARCH_INDEX_RETRY_BACKOFF=""` is safe: the loop reads `${SEARCH_INDEX_RETRY_BACKOFF:-10}` (`:-` defaults on empty too), and the `SearchIndexesUnavailable` branch's `=45` assignment still takes effect. Making it `local` is a small, legit robustness win (directly addressing prior feedback).
- **`set -eo pipefail` safe.** A non-zero `grep -qz` inside an `elif` condition does not abort under `set -e` (verified).
- **No injection / boundary risk.** `$deploy_log` is an internal `mktemp` capturing only `bunx convex deploy` output; the grep pattern is a constant literal and the filename is quoted.

### Non-blocking observations (optional polish — do NOT block merge)
1. **Order-dependence:** `InvalidModules.*Function execution timed out` requires `InvalidModules` to appear *before* the timeout line. In the real push-analyze error it always does (the `InvalidModules` wrapper precedes the nested message), so this is fine — just slightly more brittle than the order-independent form described in the PR body.
2. **Theoretical false-positive:** if an unrelated `Function execution timed out` co-occurred with `InvalidModules` in the same deploy log it would also match. This is implausible — the deploy log captures only the push/analyze phase, and runtime/action timeouts (1800s budget) don't land there. Even if it did match, the consequence is bounded (≤3 retries, then correct `exit 1`). Optionally tighten by anchoring on the `2s` budget.
3. **Style nit:** this is the only `grep -qz` in the chain (others use `grep -q`); a one-line comment noting "`-z`: error spans multiple lines" would aid the next reader.

None of the above blocks the merge.

**READY TO MERGE.**

PR #2147, opened 2026-06-25
fix: chat attachment cap & cross-batch dedup bypassable by concurrent batches (#2078)

### LINKED ISSUE #2078: Bug: Chat attachment 10-file cap and cross-batch duplicate detection are bypassable by concurrent attach batches (and compressed-image dedup misses)
**Area / route**
Chat composer attachment handling. `use-convex-file-upload.ts`, `chat-input.tsx`. Two related dedup/cap defects preserved below.

**Sites**

- **[24] Chat 10-file cap and cross-batch duplicate detection are bypassable by concurrent attach batches (in-flight uploads not counted)**
  - Evidence: count check use-convex-file-upload.ts:230-247 and dedup use-convex-file-upload.ts:208-225 both read attachmentsRef.current only; uploadingFiles tracked separately at lines 282-283 and committed to attachments only at line 335; attach button not disabled during upload — chat-input.tsx:326 (`attachDisabled = disabled || isLoading`) and chat-input.tsx:1050
- **[23] Re-attaching an image > 1MB is not detected as a duplicate (dedup compares original name/size against the stored compressed/renamed values)**
  - Evidence: use-convex-file-upload.ts:208-225 (existingKeys from `att.fileName:att.fileSize`, incoming key from `entry.file.name:entry.file.size`); compression rename+resize at lib/utils/compress-image.ts:129-142; stored attachment uses fileToUpload.name/.size at use-convex-file-upload.ts:325-333

**Notes**
[24] fix: fold isUploading into attachDisabled and/or count in-flight files in both the count gate and the dedup gate. [23] fix: compute the dedup key from the same canonical values (track the original file identity on the stored attachment, or dedup after compression). Severity medium: worst outcome is an extra valid chip / skipped dedup — no corruption.

---

larryro:
## Summary

Fixes two related attachment dedup/cap defects in the chat composer (`use-convex-file-upload.ts`).

**[24] 10-file cap & cross-batch dedup bypassable by concurrent attach batches.**
The count, dedup, and total-size gates only read the committed `attachments` state. In-flight uploads (tracked separately in `uploadingFiles` and committed to `attachments` only after the upload finishes) were invisible to a second attach batch fired while the first was still uploading — letting a user exceed the 10-file cap, the total-size cap, and skip duplicate detection across batches.

Fix: a `pendingUploadsRef` reserves a slot for every accepted file *before* its upload starts (recording the original dedup key + source size), and the count/dedup/total-size gates now fold those reservations in alongside committed attachments. Reservations are released when the file commits (or fails).

**[23] Re-attaching an image > 1MB not detected as a duplicate.**
Images are compressed and renamed (`.png` → `.jpg`, resized) before storage, so the stored attachment's `fileName`/`fileSize` no longer match the source file. The dedup key compared the incoming *original* name/size against the stored *compressed* values, so a re-attach of the same image was never caught.

Fix: the stored attachment now keeps `originalFileName`/`originalFileSize`, and the dedup key is computed from those canonical source values on both sides.

## Tests

Added `use-convex-file-upload.test.ts` (jsdom / `test:ui`) covering:
- in-flight uploads counted against the 10-file cap across concurrent batches (6 + 6 → capped at 10, not 12);
- a file uploading in one batch dedups a re-attach in a second batch;
- a re-attached compressed image dedups against its original identity.

Verified locally: `test:ui` (3/3 pass), `tsc --noEmit`, `oxlint --type-aware`, `oxfmt --check` all green.

Closes #2078

### DIFF
--- a/services/platform/app/features/chat/hooks/use-convex-file-upload.test.ts
+++ b/services/platform/app/features/chat/hooks/use-convex-file-upload.test.ts
@@ -1,0 +1,236 @@
+// @vitest-environment jsdom
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { toast } from '@/app/hooks/use-toast';
+import { compressImage } from '@/lib/utils/compress-image';
+
+import { useConvexFileUpload } from './use-convex-file-upload';
+
+// ---------------------------------------------------------------------------
+// Module mocks. The hook leans on Convex mutations, the upload policy query,
+// i18n, toasts, and the image compressor — none of which are relevant to the
+// dedup/cap accounting under test, so they are stubbed. The real file-type
+// helpers and constants (CHAT_MAX_FILE_COUNT, etc.) are kept; only the async
+// byte-sniffing `detectMediaMime` is forced to "not media" so a plain File
+// resolves by extension without an arrayBuffer round-trip.
+// ---------------------------------------------------------------------------
+
+const generateUploadUrl = vi.fn().mockResolvedValue('https://upload.test/url');
+const saveFileMetadata = vi.fn().mockResolvedValue(undefined);
+
+vi.mock('./mutations', () => ({
+  useGenerateUploadUrl: () => ({ mutateAsync: generateUploadUrl }),
+}));
+
+vi.mock('@/app/hooks/use-convex-mutation', () => ({
+  useConvexMutation: () => ({ mutateAsync: saveFileMetadata }),
+}));
+
+vi.mock('@/app/features/settings/governance/hooks/queries', () => ({
+  useUploadPolicy: () => ({
+    policyEnabled: false,
+    maxFileSize: 100 * 1024 * 1024,
+    allowedTypes: [],
+    blockedExtensions: [],
+    allowedExtensions: [],
+  }),
+}));
+
+vi.mock('@/app/hooks/use-toast', () => ({ toast: vi.fn() }));
+
+vi.mock('@/lib/i18n/client', () => ({
+  useT: () => ({ t: (key: string) => key }),
+}));
+
+vi.mock('../utils/get-audio-duration', () => ({
+  getAudioDuration: vi.fn().mockResolvedValue(null),
+}));
+
+vi.mock('@/lib/utils/compress-image', () => ({ compressImage: vi.fn() }));
+
+vi.mock('@/lib/shared/file-types', async (importOriginal) => {
+  const actual =
+    await importOriginal<typeof import('@/lib/shared/file-types')>();
+  return { ...actual, detectMediaMime: vi.fn().mockResolvedValue(null) };
+});
+
+const compressImageMock = vi.mocked(compressImage);
+const toastMock = vi.mocked(toast);
+
+// A controllable POST/fetch: each call parks until the test resolves it, so a
+// batch can be held "in-flight" while a second batch races the cap/dedup gates.
+let pendingFetches: Array<() => void> = [];
+let storageCounter = 0;
+
+function installFetch() {
+  pendingFetches = [];
+  storageCounter = 0;
+  vi.stubGlobal(
+    'fetch',
+    vi.fn(
+      () =>
+        new Promise((resolve) => {
+          const storageId = `storage-${storageCounter++}`;
+          pendingFetches.push(() =>
+            resolve({
+              ok: true,
+              json: async () => ({ storageId }),
+            } as Response),
+          );
+        }),
+    ),
+  );
+}
+
+function resolveAllFetches() {
+  const settle = pendingFetches;
+  pendingFetches = [];
+  for (const done of settle) done();
+}
+
+function makeFile(name: string, size: number, type: string): File {
+  return new File([new Uint8Array(size)], name, { type });
+}
+
+// Let the synchronous gate logic and the early `await detectMediaMime` chain
+// run so in-flight reservations are registered before the next batch fires.
+async function flush() {
+  await act(async () => {
+    for (let i = 0; i < 5; i++) await Promise.resolve();
+  });
+}
+
+const config = { organizationId: 'org-1' };
+
+beforeEach(() => {
+  installFetch();
+  vi.stubGlobal('URL', {
+    ...URL,
+    createObjectURL: vi.fn(() => 'blob:preview'),
+    revokeObjectURL: vi.fn(),
+  });
+  compressImageMock.mockImplementation(async (file: File) => {
+    const compressed = new File(
+      [new Uint8Array(200)],
+      file.name.replace(/\.[^.]+$/, '.jpg'),
+      { type: 'image/jpeg' },
+    );
+    return {
+      file: compressed,
+      wasCompressed: true,
+      originalSize: file.size,
+      finalSize: compressed.size,
+    };
+  });
+});
+
+afterEach(() => {
+  vi.clearAllMocks();
+  vi.unstubAllGlobals();
+});
+
+describe('useConvexFileUpload — concurrent-batch cap & dedup', () => {
+  it('counts in-flight uploads against the 10-file cap across batches', async () => {
+    const { result } = renderHook(() => useConvexFileUpload(config));
+
+    // Batch A: 6 files, held in-flight (fetch parked).
+    const batchA = Array.from({ length: 6 }, (_, i) =>
+      makeFile(`a-${i}.txt`, 10, 'text/plain'),
+    );
+    let promiseA!: Promise<void>;
+    act(() => {
+      promiseA = result.current.uploadFiles(batchA);
+    });
+    await flush();
+
+    // 6 reservations now in-flight; only 4 slots remain of the 10-file cap.
+    const batchB = Array.from({ length: 6 }, (_, i) =>
+      makeFile(`b-${i}.txt`, 10, 'text/plain'),
+    );
+    let promiseB!: Promise<void>;
+    act(() => {
+      promiseB = result.current.uploadFiles(batchB);
+    });
+    await flush();
+
+    // Batch B should be capped to 4 and surface the too-many-files toast.
+    expect(
+      toastMock.mock.calls.some(([arg]) => arg?.title === 'tooManyFiles'),
+    ).toBe(true);
+
+    resolveAllFetches();
+    await act(async () => {
+      await Promise.all([promiseA, promiseB]);
+    });
+
+    // 6 from A + 4 from B = exactly the cap; never 12.
+    await waitFor(() => expect(result.current.attachments).toHaveLength(10));
+  });
+
+  it('dedupes a file already uploading in another batch', async () => {
+    const { result } = renderHook(() => useConvexFileUpload(config));
+
+    const fileA = makeFile('dup.txt', 42, 'text/plain');
+    let promiseA!: Promise<void>;
+    act(() => {
+      promiseA = result.current.uploadFiles([fileA]);
+    });
+    await flush();
+
+    // Re-attach the identical file while the first is still in-flight.
+    const fileB = makeFile('dup.txt', 42, 'text/plain');
+    let promiseB!: Promise<void>;
+    act(() => {
+      promiseB = result.current.uploadFiles([fileB]);
+    });
+    await flush();
+
+    expect(
+      toastMock.mock.calls.some(([arg]) => arg?.title === 'duplicateFile'),
+    ).toBe(true);
+
+    resolveAllFetches();
+    await act(async () => {
+      await Promise.all([promiseA, promiseB]);
+    });
+
+    await waitFor(() => expect(result.current.attachments).toHaveLength(1));
+  });
+
+  it('dedupes a re-attached image against its original (pre-compression) identity', async () => {
+    const { result } = renderHook(() => useConvexFileUpload(config));
+
+    // Upload an image; compression renames .png -> .jpg and shrinks the size.
+    const image = makeFile('photo.png', 1_500_000, 'image/png');
+    let promise1!: Promise<void>;
+    act(() => {
+      promise1 = result.current.uploadFiles([image]);
+    });
+    await flush();
+    resolveAllFetches();
+    await act(async () => {
+      await promise1;
+    });
+    await waitFor(() => expect(result.current.attachments).toHaveLength(1));
+
+    // The stored attachment carries the compressed name/size...
+    expect(result.current.attachments[0]?.fileName).toBe('photo.jpg');
+    expect(result.current.attachments[0]?.fileSize).toBe(200);
+    // ...but retains the original identity for dedup.
+    expect(result.current.attachments[0]?.originalFileName).toBe('photo.png');
+    expect(result.current.attachments[0]?.originalFileSize).toBe(1_500_000);
+
+    // Re-attaching the same original image is detected as a duplicate.
+    toastMock.mockClear();
+    const sameImage = makeFile('photo.png', 1_500_000, 'image/png');
+    await act(async () => {
+      await result.current.uploadFiles([sameImage]);
+    });
+
+    expect(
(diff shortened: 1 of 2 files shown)


larryro:
## Desk review — #2078 attachment cap & cross-batch dedup fix

**Verdict: NOT READY — one blocking item, and it is CI, not the code.** The implementation itself meets the review bar; the merge gate is currently red.

### What I verified
- **Tests (ran them):** `bunx vitest --run --config vitest.ui.config.ts app/features/chat/hooks/use-convex-file-upload.test.ts` → **3 passed / 3**. All three sub-defects are covered (10-file cap across concurrent batches; cross-batch dedup of an in-flight file; compressed-image original-identity dedup).
- **Concurrency correctness (the crux):** the gate block — dedup (≈L242), file-count (≈L273), total-size (≈L308) — and the slot reservation (≈L335) contain **no `await` between reading `pendingUploadsRef` and writing the reservations**. That span is synchronous, so it is atomic under JS's single-threaded loop: whichever overlapping batch reaches it first reserves before yielding, and the second batch then sees those reservations. The early `await detectMediaMime` / audio-duration awaits sit *before* the gate block and cannot interleave a second batch into the middle of it. **No double-spend race.** ✅
- **Dedup key consistency:** pending (`file.name:file.size`), committed (`originalFileName ?? fileName : originalFileSize ?? fileSize`), and incoming (`entry.file.name:entry.file.size`) all key on the *source* identity. The `?? fileName/fileSize` fallback keeps non-compressed files and legacy attachments correct. ✅
- **Total-size gate** mixes committed compressed sizes + pending/incoming *source* sizes — a conservative upper bound (compression only shrinks), so it cannot under-count. ✅
- **Reservation lifecycle:** reserved before any async work; released on success (≈L405) and idempotently in `finally` (≈L422). All early returns occur *before* reservations are made, so no orphaned entries. ✅

### Blocking
1. **CI is red.** `gh pr checks` reports **`Validate images` FAILING** — `sandbox-runtime 3114 MB exceeds 2800 MB budget`. This check is **unrelated to this diff** (the PR touches only a frontend TS hook + its test; it changes no Docker/image content), so it is almost certainly a pre-existing main/infra failure rather than something this PR introduced. **No code change to the reviewed logic is required to fix it.** But the merge bar requires green CI, so this must be resolved/rebased to green (or confirmed non-required) before merge. Every other check (Type check, Lint, Unit, UI, Format, Playwright, etc.) passes.

- **Cross-session dedup residual:** `use-persisted-attachments.ts` does not persist `originalFileName`/`originalFileSize`, so after a page refresh / thread switch a re-attached >1 MB image is not deduped. This is **not a regression** (pre-PR behavior was identical) — just a follow-up if full cross-session dedup is wanted.
- **Micro-window undercount:** the reservation is deleted (≈L405) one tick before `setAttachments` updates `attachmentsRef`; a third batch firing in that sub-render gap could under-count by one. Theoretical/negligible.
- **`fileId` = `name-size-Date.now()-index`** could collide only on identical name+size+ms+index across batches, and even then the dedupKey is identical — a transient under-count by one at worst. A counter/uuid would remove it.
- **Test gaps:** the concurrent total-size path and the upload-failure (reservation-release) path aren't directly asserted, though the implemented behavior is conservative/idempotent.

**NOT READY — CHANGES REQUIRED**
1. Get CI green: the `Validate images` check (`sandbox-runtime` image exceeds the 2800 MB budget) is currently failing. It is unrelated to this diff — resolve it on the infra/main side (or rebase once main is green); no change to `use-convex-file-upload.ts` is needed for correctness.

larryro:
## Desk review — PR #2147 (fixes #2078): chat attachment cap & cross-batch dedup

**Verdict: NOT READY — CHANGES REQUIRED (test coverage gaps).** The implementation is correct and genuinely resolves the reported issue; the blocking items below are **test additions only — do not change the hook's logic.**

### CI state
`gh pr checks` — **green.** 49 checks `pass` (Format, Lint, Type check, Unit, UI, Browser, Knip, Storybook, all 16 Playwright platform shards, Build *, Migrations, Opengrep, Trivy/Scan, Smoke test, etc.); the remaining checks (`Docs container test`, `Smoke test (fork PR)`, `Validate images (fork PR)`, `Web container test`) report `skipping` (not-applicable fork-only jobs), not failures. No failed or still-pending required checks.

### Test outcome
Ran the project's own UI suite against the changed file:
`bunx vitest --run --config vitest.ui.config.ts app/features/chat/hooks/use-convex-file-upload.test.ts`
→ **Test Files 1 passed (1); Tests 3 passed (3).** The three tests are real (a parked-`fetch` stub holds batch A in-flight while batch B races the gates) and meaningfully assert toast + final committed count, not tautologies.

### What the fix does, and why it is correct
- Adds `pendingUploadsRef` (a `Map<fileId,{dedupKey,size}>`) that reserves a slot for every accepted file **before** its upload starts, recording the **original** (pre-compression) dedup key + source size. The count, dedup, and total-size gates now fold these reservations in alongside committed `attachments`. Reservations are released on commit (line 405) and idempotently in `finally` (line 422).
- Stores `originalFileName`/`originalFileSize` on each attachment and keys dedup off those canonical source values on both sides, so re-attaching a >1MB image that was renamed/resized by compression is now caught.

I confirmed across five independent review lenses (correctness, robustness, elegance, tests, boundaries), each re-verified:
- **Concurrency is sound.** The read-check-reserve region (lines 242→342: dedup gate → count gate → total-size gate → `pendingUploadsRef.set`) contains **no `await`**, so it runs atomically; a second batch fired mid-upload reads the first batch's reservations before writing its own. All three caps (count, total-size, cross-batch dedup) hold. (Confirmed by direct grep of the await points.)
- **[23] genuinely fixed** for every traced case: PNG/JPEG compression (rename + shrink), images that are *not* compressed (`fileToUpload === file`, keys still match), and non-image attachments (`?? fileName`/`?? fileSize` fallback). `??` (not `||`) is the correct operator so a legitimate size of 0 isn't masked.
- **No reservation leak on the normal failure path** — the `finally` frees the slot on compression throw, `generateUploadUrl` reject, `!result.ok`, missing `storageId`, or `saveFileMetadata` reject.
- **`fileId` collisions** (`${name}-${size}-${Date.now()}-${index}`) cannot clobber live reservations, because two live reservations with identical name+size are prevented by the dedup gate itself.
- **Total-size accounting** uses source size for pending/incoming (upper bound) vs compressed size for committed — conservative (can only over-reject slightly), never under-counts. Documented.
- **Button-left-enabled** (`chat-input.tsx:326` still `attachDisabled = disabled || isLoading`) is acceptable: the issue explicitly allows "fold isUploading into attachDisabled **and/or** count in-flight files," and the in-flight accounting is authoritative.

### BLOCKING — required before merge (tests only; logic is correct)
1. **Total-size cross-batch cap is untested.** Issue #2078 names the **total-size cap** as one of the three concurrent-batch bypasses, and the PR claims to fix it, but no test exercises the `existingSize += pending.size` fold-in (`use-convex-file-upload.ts:312-314`). The largest file in the suite is 1.5 MB against a 200 MB `CHAT_MAX_TOTAL_SIZE`, so `if (existingSize + incomingSize > CHAT_MAX_TOTAL_SIZE)` never trips. Add a test to `use-convex-file-upload.test.ts` mirroring the count-cap test: hold a batch of large files in-flight (parked `fetch`), fire a second batch that would only exceed the total-size cap once the in-flight reservations are counted, and assert the `totalSizeExceeded` toast fires and the over-cap files are rejected. A regression that dropped the `pendingUploadsRef` sum (lines 312-314) must fail this test.
2. **Upload-failure reservation-release path is untested.** The fix's own comment (`use-convex-file-upload.ts:419-421`) calls the `finally` cleanup out as a distinct path, but the test's `fetch` stub only ever resolves `ok: true`, so failure is never exercised. If this cleanup regressed, a failed upload would leak a permanent phantom reservation that silently consumes a cap slot and total-size budget for the life of the composer. Add a test where one in-flight upload **fails** (resolve `fetch` with `ok: false`, or reject it / `saveFileMetadata`), then attach another file and assert the failed file's slot was freed — e.g. the cap/total-size budget is fully reclaimed and a subsequent attach up to the cap still succeeds.

### Non-blocking (optional polish — no logic change required to merge once the two tests land)
- `setUploadingFiles(...)` (`use-convex-file-upload.ts:346`) sits just **outside** the inner `try` (which begins at line 348). A synchronous throw there would orphan that file's reservation. React state setters with a function updater don't throw synchronously here, so it's not reachable — but moving line 346 inside the `try` would make the "every reservation is eventually deleted" invariant structurally guaranteed.
- The dedup key `${name}:${size}` is built inline at three sites (lines 245-247, 255, 338) that must stay in lockstep; a tiny `dedupKeyOf(name, size)` helper would remove the silent-breakage coupling.
- The success-path `delete` at line 405 shaves only a sub-render-tick window versus relying on the `finally` at 422 (since `setAttachments` commits asynchronously); harmless and well-commented, just slightly oversold by the comment.
- Optional: a within-single-batch dedup test, and replacing the fixed 5-tick `flush()` with a `waitFor` on `uploadingFiles.length`, would harden the suite against future added `await`s before the reservation.

---
**NOT READY — CHANGES REQUIRED**
1. Add a test in `use-convex-file-upload.test.ts` for the **total-size cap across concurrent batches** (in-flight large files counted via `pendingUploadsRef` into the total-size gate; assert `totalSizeExceeded` and rejection).
2. Add a test in `use-convex-file-upload.test.ts` for the **upload-failure path** freeing its `pendingUploadsRef` reservation (fail one in-flight upload, then assert its cap slot / total-size budget is reclaimed).

larryro:
## Desk Review — PR #2147 (issue #2078)

CI is **fully green** (all 41 checks pass, incl. Lint, Format, typecheck/Analyze, Build *, Playwright 1–16, Smoke). I ran the project's own UI suite for the changed hook:
`bunx vitest --run --config vitest.ui.config.ts app/features/chat/hooks/use-convex-file-upload.test.ts` → **5/5 pass**.

The fix is well-built and resolves most of the issue:
- ✅ Compressed-image dedup (`originalFileName`/`originalFileSize`) — correct; keys are consistent across stored / pending / incoming; fields are stripped before the server mutation in `use-send-message.ts`, so no leak.
- ✅ Cross-batch dedup and cap accounting *while a batch is still uploading* (`pendingUploadsRef`) — closes the large multi-second window that made the original 6+6→12 bypass trivial.
- ✅ Reservation release on success and failure (idempotent `finally`).

### ❌ BLOCKING — the concurrent-batch cap is still bypassable (commit→render handoff gap)

The three gates (count, total-size, dedup) read `pendingUploadsRef.current` **plus** `attachmentsRef.current`. But `attachmentsRef.current` is refreshed **only on React re-render** (`use-convex-file-upload.ts:107-108`), whereas the commit path **synchronously** deletes the pending reservation *before* that render lands:

```
405  pendingUploadsRef.current.delete(fileId);   // reservation gone immediately
406  setAttachments((prev) => [...prev, attachment]); // attachmentsRef updates only on the NEXT render
```

So there is a window — after a file’s reservation is deleted but before React re-renders — in which the committed file is in **neither** `pendingUploadsRef` **nor** `attachmentsRef.current`. A second batch whose gate evaluates in that window sees full capacity and over-reserves.

**I reproduced this deterministically** against the real hook + real React scheduler (a throwaway vitest test, not committed):
- REPRO: batch A (10 files) commits, then batch B (10 files) fires while React hasn’t re-rendered → **final attachments = 20**, double the `CHAT_MAX_FILE_COUNT` cap.
- CONTROL: identical sequence but with a macrotask gap so React re-renders between A and B → correctly capped at **10**.

This is the same class of defect the issue reports (“cap … bypassable by concurrent attach batches”) — the fix shrinks the window from seconds (in-flight) to the commit→render tick, but does not close it. The **total-size cap and cross-batch dedup share the identical gap** by the same mechanism (a re-attached duplicate committed-but-not-yet-rendered slips through; extra bytes slip through).

**Required change:** ensure a committed file is never invisible to the gates during the reservation→render handoff. Options (implementer’s choice):
1. Update a synchronous committed-attachments ref at commit time (e.g. set `attachmentsRef.current = [...attachmentsRef.current, attachment]` in the same block as the `delete`/`setAttachments` at lines 405-406), so the gates read the new file immediately rather than waiting for the render; or
2. Defer the `pendingUploadsRef` deletion until the attachment is actually reflected in `attachmentsRef.current` (e.g. release in an effect keyed on `attachments`), so pending and committed never have a gap.
Then add a regression test that commits batch A and fires batch B before a React re-render, asserting the cap (and total-size, and dedup) still hold.

### Non-blocking notes
- `use-convex-file-upload.ts:405` — the `delete(fileId)` here is redundant with the `finally` delete at line 422 (both run in the same synchronous block; no `await` between them). Harmless, but the success-path delete can be dropped once the fix above is in. Whatever you do, keep the gates gap-free.
- `:336` `fileId = \`${file.name}-${file.size}-${Date.now()}-${index}\`` — theoretical collision only if two batches with an identical file fire in the same millisecond; cross-batch dedup already prevents the same file from being reserved twice, so low risk. Not blocking.

The existing 5 tests are meaningful and faithful, but note they only exercise the **in-flight** window (batch A held mid-upload); none cover the **commit→render** window above, which is why they pass while the bypass remains.

larryro (commented):
## Desk review — READY TO MERGE ✅

Reviewed the full implementation (not just the diff) over two rounds: 5 parallel dimension reviews (correctness, robustness, elegance, tests, boundaries/resolution) followed by independent re-verification of every candidate blocker.

### CI & tests
- **CI: all checks green** (`gh pr checks` — Analyze, Build *, Lint, Format, Knip, Opengrep, Playwright ×16, Smoke test, etc. all `pass`).
- **Ran the project's own test once:** `bunx vitest --run --config vitest.ui.config.ts app/features/chat/hooks/use-convex-file-upload.test.ts` → **6 passed / 6**.

### Resolution — both reported defects fixed
- **[24] concurrent-batch cap/dedup/total-size bypass:** `pendingUploadsRef` reserves a slot (original dedup key + source size) *before* each upload starts and is folded into all three gates (count, dedup, total-size). Reservations are released on commit and idempotently in `finally` on failure — no leak, no permanent composer lockout. The gate-accounting approach is a complete fix on its own; it doesn't depend on disabling the attach button, and every interactive add-path (paste, file input, drop) funnels through `uploadFiles`.
- **[23] >1MB image re-attach not deduped:** stored attachments now carry `originalFileName`/`originalFileSize`, and the dedup key is computed from those canonical pre-compression values on the committed, pending, and incoming sides symmetrically.

### Verified, not a defect
- **commit→render race (lines 403–417):** the manual `attachmentsRef.current` mirror runs synchronously *before* the reservation `delete`, and `setAttachments` is enqueued in the same microtask — so the reduced `attachments` at any later render always contains the committed file. There is no render where the file is in neither `pendingUploadsRef` nor `attachmentsRef.current`. The comment's reasoning is sound; test #5 exercises exactly this window and passes.
- **Reservation lifecycle:** released on both success and failure paths; double-`delete` is harmless (idempotent). Early returns all precede the reservation loop, so nothing leaks.
- **Server contract:** `originalFileName`/`originalFileSize` are client-only composer state; `saveFileMetadata` still sends only the uploaded (compressed) `fileName`/`size`. No leak, no contract change.

### Non-blocking observations (optional follow-ups, not required to merge)
1. `fileId` = `name-size-Date.now()-index` is theoretically collidable across same-millisecond batches; the dedup gate (same name:size identity) shields the count/size gates from any exploitable collision today. `crypto.randomUUID()` would be more robust.
2. The `${name}:${size}` dedup-key literal appears in 3 places; a one-line helper would prevent the pending/incoming formats from drifting.
3. Test gaps (minor): no within-batch duplicate test (`[dup, dup]` in one call) and no test asserting the `?? fileName`/`?? fileSize` fallback for a non-image committed attachment. The `flush()` fixed 5-microtask-turn helper is slightly timing-fragile vs a `waitFor`.
4. Pre-existing & out of scope: `use-persisted-attachments.ts` restores attachments via `setAttachments` without re-running the gates. Untouched by this PR; data was gate-validated at original attach.

**Verdict: READY TO MERGE.**

Issue #1942, opened 2026-06-23
Improvement: Unify field height to 36px and replace display-only disabled inputs with borderless read-only fields


yannickmonney:
## Problem
Some disabled fields render at a different height than active inputs, and disabled inputs are used where the value is purely informational — looking heavy and shifting layout.

## Current behavior
- Inputs use `h-9` (36px): `packages/ui/src/components/forms/input.tsx:20`, `services/platform/app/components/ui/forms/input.tsx:25`, select trigger `select.tsx:58`.
- Disabled styling is `disabled:cursor-not-allowed disabled:opacity-50` (`input.tsx:23`) — there is **no** read-only visual variant.
- A borderless read-only display primitive already exists: `Field` (`services/platform/app/components/ui/forms/field.tsx`).

## Proposed change
1. Audit any control rendering at a non-36px height when disabled and normalize to `h-9`.
2. Add a `readOnly` visual variant: transparent background, **no border**, same `h-9` height + horizontal padding as an input so there is **no layout shift** when toggling. Extend the app `Input` wrapper or align `Field` to input metrics — it should behave like normal text but occupy the field's footprint.
3. Replace **display-only** disabled fields (values the user cannot edit in context) with the read-only variant. Keep `disabled` only where the control is *temporarily* unavailable (and pair with the disabled-reason tooltip issue).

## Acceptance criteria
- [ ] All form controls measure 36px tall in every state.
- [ ] Display-only values render borderless and text-like, with no layout shift vs. their editable state.
- [ ] A Storybook story / visual check demonstrates editable vs. read-only vs. disabled.

PR #2154, opened 2026-06-25
fix: dark-mode destructive contrast, empty-table hover, notification dot

AdeolaAdekoya:
## Summary

Three small, related UX/accessibility fixes for the platform, each as its own atomic commit.

- **WCAG AA on destructive fills (dark mode).** The `--destructive` token does double duty as error-*text* colour and as the *fill* behind white text on destructive buttons/badges. In dark mode it's lightened toward red-400 to stay legible as text, which drops white-on-fill contrast to **~2.75:1** (AA needs 4.5:1); light mode was borderline at **~3.7:1**. Pointed the four white-text-on-red fills at a fixed `red-600`/`red-700` — the shared destructive button variant (covers *Delete all chats* and every destructive button), the notification bell badge, the mobile tab-bar badge, and the chat-history "new response" badge. `red-600` clears AA against white in both themes. Error *text* keeps the token untouched (red-on-neutral is ~7:1).
- **Empty table rows no longer respond to hover.** Four governance editors suppressed the shared `TableBody` hover with a per-row `hover:bg-transparent`, which loses the CSS specificity fight against the `tbody` selector and highlighted anyway. Switched them to `data-no-hover`. Hover signals "this row is actionable" — an empty placeholder has nothing to act on. Real data rows keep hovering normally.
- **Notification unread dot → blue.** `sky-500` (#0ea5e9) reads as teal/cyan; switched to `blue-500` (#3b82f6) — the unambiguous unread-blue cue, which also clears the 3:1 non-text contrast bar in both themes.

## Test plan

- [ ] Dark mode: destructive buttons (*Delete all chats*), notification bell badge, mobile tab-bar badge, and chat "new response" badge are legible (white text on red, ≥4.5:1).
- [ ] Light mode: same destructive surfaces still look correct, not overly bold.
- [ ] Governance editors (Policies & Limits / budget, default-model, feature-flags, model-access): empty-state rows show no hover highlight; populated rows still highlight on hover.
- [ ] Notifications popover: unread dot is blue, not teal.

### DIFF
--- a/packages/ui/src/components/navigation/bottom-tab-bar.tsx
+++ b/packages/ui/src/components/navigation/bottom-tab-bar.tsx
@@ -115,7 +115,7 @@ function BottomTabBarButton({ item }: BottomTabBarButtonProps) {
           <>
             <span
               aria-hidden="true"
-              className="bg-destructive text-destructive-foreground absolute -top-0.5 -right-0.5 inline-flex h-4 min-w-4 items-center justify-center rounded-full px-1 text-[10px] leading-none font-semibold"
+              className="text-destructive-foreground absolute -top-0.5 -right-0.5 inline-flex h-4 min-w-4 items-center justify-center rounded-full bg-red-600 px-1 text-[10px] leading-none font-semibold"
             >
               {item.badge}
             </span>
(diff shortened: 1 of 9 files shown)


Israeltheminer:
### Search inputs lost their border/ring — could fold the fix in here

Every `SearchInput` now renders without its outline/ring until focused (e.g. the "Search integrations…" box).

**Cause:** #1942 (borderless read-only variant) made `Input` auto-resolve to the ring-less `readOnly` variant whenever the native `readOnly` attribute is set — `input.tsx:95`:
```ts
const resolvedVariant = variant ?? (props.readOnly ? 'readOnly' : undefined);
```
`SearchInput` sets `readOnly` until focus as an autofill-suppression trick (`search-input.tsx:65,115`) and passes no explicit variant, so it renders borderless until clicked into.

**Fix** — pin the variant in `search-input.tsx` so the autofill trick can't flip it:
```diff
           <Input
+            variant="default"
             className={cn('pl-10 max-w-70 h-9', ...)}
```
`resolvedVariant = variant ?? (...)`, so an explicit `variant="default"` short-circuits the auto-detection and keeps the ring in all states.

Affects **every** search box app-wide, not just integrations. Since this is already a UI/contrast polish PR, might be cleanest to land it here. @AdeolaAdekoya

PR #2176, opened 2026-06-27
fix(platform): require name in customer edit dialog for vendor parity

### LINKED ISSUE #2060: Bug: Clearing a customer's name in the edit dialog is a silent no-op but still shows success (and lacks the Vendor's required-name parity)
**Area / route**
Knowledge → Customers → Customer EDIT dialog. `customer-edit-dialog.tsx`, backend `convex/customers/update_customer.ts`. Vendor edit dialog is the correct reference. Two findings, one root cause (missing required-name validation), preserved below.

**Sites**

- **[74] Clearing a customer's name in the edit dialog is a silent no-op but still shows "Customer updated successfully"**
  - Evidence: customer-edit-dialog.tsx:55 (name: z.string().trim()) + customer-edit-dialog.tsx:102-109 (name: trimmedName || undefined; toast updateSuccess) + convex/customers/update_customer.ts:86-90 (Object.fromEntries(... value !== undefined) then ctx.db.patch).
- **[76] Customer edit dialog accepts an empty Name while the sibling Vendor edit dialog requires it (validation parity)**
  - Evidence: customer-edit-dialog.tsx:55 + Input at :142-149 (no required) vs vendor-edit-dialog.tsx:43-46 (.min(1)) + Input at :122-130 (required).

**Notes**
Fix: change the customer schema Name to `z.string().trim().min(1, tCommon('validation.required',{field:tCustomers('name')}))` and add `required` to the Input (matching vendor-edit-dialog.tsx:43-46/122-130); the `|| undefined` coalesce then becomes unnecessary and the misleading success toast no longer fires on a no-op. Distinct from #1991 (max-length/trim on create forms) — this is min/required parity between the Customer and Vendor EDIT dialogs.

---

larryro:
## Summary

Clearing a customer's name in the edit dialog was a silent no-op that still showed "Customer updated successfully". The schema only had `z.string().trim()` (no `min(1)`) and `onSubmit` coalesced the empty name to `undefined`, which the backend's `Object.fromEntries(... value !== undefined)` filter then dropped — so nothing changed but the success toast still fired. The sibling **Vendor** edit dialog already required its name; this brings the Customer dialog to parity.

## Changes
- `customer-edit-dialog.tsx`: name schema is now `z.string().trim().min(1, validation.required)`, the name `Input` is marked `required`, and the `name: trimmedName || undefined` coalesce is replaced with `name: data.name.trim()` — matching `vendor-edit-dialog.tsx`.
- `customer-edit-dialog.test.tsx`: replaced the now-invalid "submits when customer has no name" test with one asserting that clearing the name blocks the mutation, suppresses the success toast, and keeps the dialog open.

## Verification
- `bunx tsc --noEmit` — clean
- `bunx oxlint --type-aware` on changed files — 0 warnings/errors
- `vitest` (ui project) for the dialog test — 5/5 pass

Closes #2060

### DIFF
--- a/services/platform/app/features/customers/components/customer-edit-dialog.test.tsx
+++ b/services/platform/app/features/customers/components/customer-edit-dialog.test.tsx
@@ -54,36 +54,32 @@ describe('CustomerEditDialog', () => {
     expect(screen.getByDisplayValue('test@example.com')).toBeInTheDocument();
   });
 
-  it('submits when customer has no name', async () => {
+  it('blocks submission when the name is cleared', async () => {
     const onOpenChange = vi.fn();
     const { user } = render(
       <CustomerEditDialog
-        customer={makeCustomer({ name: undefined })}
+        customer={makeCustomer({ name: 'John' })}
         isOpen={true}
         onOpenChange={onOpenChange}
       />,
     );
 
-    const emailInput = screen.getByDisplayValue('test@example.com');
-    await user.clear(emailInput);
-    await user.type(emailInput, 'new@example.com');
+    const nameInput = screen.getByDisplayValue('John');
+    await user.clear(nameInput);
 
     const submitButton = screen.getByRole('button', { name: /save/i });
     await user.click(submitButton);
 
+    // Required-name validation must block the no-op update entirely:
+    // no mutation fires, no misleading success toast, and the dialog
+    // stays open so the user can correct the empty name.
     await waitFor(() => {
-      expect(mockMutateAsync).toHaveBeenCalledWith({
-        customerId: 'customer-1',
-        name: undefined,
-        email: 'new@example.com',
-        locale: 'en',
-        status: 'active',
-      });
+      expect(mockMutateAsync).not.toHaveBeenCalled();
     });
-
-    await waitFor(() => {
-      expect(onOpenChange).toHaveBeenCalledWith(false);
-    });
+    expect(mockToast).not.toHaveBeenCalledWith(
+      expect.objectContaining({ variant: 'success' }),
+    );
+    expect(onOpenChange).not.toHaveBeenCalledWith(false);
   });
 
   it('submits with name when customer has a name', async () => {
--- a/services/platform/app/features/customers/components/customer-edit-dialog.tsx
+++ b/services/platform/app/features/customers/components/customer-edit-dialog.tsx
@@ -52,7 +52,13 @@ export function CustomerEditDialog({
   const formSchema = useMemo(
     () =>
       z.object({
-        name: z.string().trim(),
+        name: z
+          .string()
+          .trim()
+          .min(
+            1,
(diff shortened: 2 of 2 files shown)


larryro:
## Desk Review — PR #2176 (issue #2060)

**Verdict: READY TO MERGE.**

The fix correctly resolves the reported bug: clearing a customer's name in the edit dialog is no longer a silent no-op that shows success, and the Customer dialog now has the Vendor's required-name parity.

### CI state
All checks **green** (~40 checks: Build platform/convex/db/proxy/sandbox/…, Lint, Format, Knip, Lint commits, Migrations, Opengrep, Analyze, Browser, all 16 Playwright (platform) shards, Playwright web/docs). Confirmed via `gh pr checks` — no failing or pending checks.

### Tests (run by the reviewer)
- `bunx vitest --run --config vitest.ui.config.ts app/features/customers/components/customer-edit-dialog.test.tsx` → **5/5 pass**.
- `bunx tsc --noEmit` → clean (exit 0).
- `bunx oxlint` (standard + `--type-aware`) on the two changed files → 0 warnings / 0 errors.

### What the fix does (verified)
- `customer-edit-dialog.tsx`: name schema is now `z.string().trim().min(1, validation.required)`, the name `Input` is `required`, and `onSubmit` sends `name: data.name.trim()` instead of the old `trimmedName || undefined`. Root cause confirmed in `convex/customers/update_customer.ts:86-88`: it strips `undefined` fields, so the old coalesce silently dropped the name while still firing the success toast. With validation now blocking empty/whitespace names, the bug path can no longer be reached from the dialog.
- The new test asserts the three required behaviours soundly: no mutation fires, no success toast, dialog stays open. The happy-path test correctly expects `name: 'Jane'` (no longer coalesced to undefined).
- Parity with `vendor-edit-dialog.tsx` is achieved (identical schema/Input wiring); the removed `trimmedName` leaves no dead code.

### Candidate findings investigated and dismissed (round-2 verified)
1. **de-CH.json lacks `validation.required` / `customers.name`** — NOT a defect. `packages/ui/src/i18n/init.ts` sets `fallbackLng['de-CH'] = ['de','en']`; regional bundles are intentionally sparse and fall back per-key to `de`/`en`. The Vendor dialog already uses these exact keys in production, confirming the chain works.
2. **Happy-path test doesn't assert `onOpenChange(false)`** — pre-existing (test unchanged by this PR); minor, non-blocking.

### Non-blocking observations (optional future hardening, out of scope for #2060)
- Backend `update_customer.ts` does not reject an empty/whitespace `name`; the REST API (`rest_api.ts` patchCustomer) and agent write tool could still set `name: ""`. The issue is scoped to the edit dialog, which is correctly fixed; a backend trim+guard would close these other entry points.
- A customer whose stored name is empty/undefined can no longer be edited without supplying a name — this is the intended required-name parity behaviour, not a regression.

**READY TO MERGE.**

PR #2177, opened 2026-06-27
fix(platform): validate budget rules in editor so dead rules can't be saved (#2061)

### LINKED ISSUE #2061: Bug: Budget rule editor saves a silently dead rule (scope with no target, or zero limits) that never enforces; dialog lacks inline field-validation parity
**Area / route**
Settings → Governance → Budgets rule dialog. `budget-editor.tsx`; schema `lib/shared/schemas/governance.ts`; enforcement `convex/governance/budget_enforcement.ts`. Two related findings preserved below.

**Sites**

- **[112] Budget rule editor saves a silently dead rule (user/team/role scope with no target, or any rule with zero limits) that never enforces**
  - Evidence: budget-editor.tsx:125-132 handleSubmit (no validation, no required-field check). Schema makes everything optional: lib/shared/schemas/governance.ts:171-179 (scopeId/maxTokens/maxCostCents/maxRequests all .optional()). Enforcement no-match: convex/governance/budget_enforcement.ts:93-105 (user/team/role rules require scopeId match).
- **[113] Budget rule dialog lacks the inline field-validation parity of the other bounded governance editors (no schema / errorMessage wiring)**
  - Evidence: budget-editor.tsx:276-296 (warningThreshold Input has min=0 max=100 but no errorMessage/validation); schema bound is server-side only: lib/shared/schemas/governance.ts:178 warningThresholdPercent .min(0).max(100). Contrast password-policy-editor.tsx:69-89 + 197 which surface errors.minLength inline.

**Notes**
Minimal fix client-side in RuleDialog: disable Confirm (or show inline errors) until scope is default/org or a scopeId is selected for user/team/role, AND >=1 of maxTokens/maxCostCents/maxRequests is set; wire errorMessage on each numeric Input. A defense-in-depth `.refine()` on budgetRuleSchema would also catch programmatic writes (but both client parse and saveGovernancePolicy share the schema — confirm no seeded limit-less placeholder rules first). Distinct from runtime-enforcement issues #1456/#1452/#1450 and confirm-dialog #1419; this is the create-time validation gap.

---

larryro:
## Summary

The budget rule editor (Settings → Governance → Budgets) could persist **silently dead rules** that never enforce:
- a `user`/`team`/`role` scope saved with **no target** (`scopeId`) — the enforcer requires a `scopeId` match, so it never applies;
- a rule with **no positive limit** (all of `maxTokens`/`maxCostCents`/`maxRequests` unset/zero) — the enforcer resolves no cap.

The dialog also lacked the inline field-validation parity of the other bounded governance editors (e.g. password policy).

## Changes
- Add `validateBudgetRule()` in `budget-editor.tsx` mirroring the constraints the enforcer relies on.
- Block save in the rule dialog when invalid and surface inline errors:
  - **target required** for user/team/role scopes,
  - **at least one limit required**,
  - per-field bounds (non-negative limits, warning threshold 0–100), wired via `errorMessage` on each numeric `Input` and `error` on the target selectors.
- Errors reveal on touch / submit attempt so a freshly-opened dialog isn't pre-filled with red.
- Add i18n keys (en/de/fr) and unit tests covering the dead-rule guard and the valid-save path.

## Verification
- `bunx tsc --noEmit` — clean
- `bunx oxlint --type-aware` (changed files) — 0 warnings, 0 errors
- `bunx vitest --config vitest.ui.config.ts budget-editor.test.tsx` — 11 passed

Closes #2061

### DIFF
--- a/services/platform/app/features/settings/governance/components/budget-editor.tsx
+++ b/services/platform/app/features/settings/governance/components/budget-editor.tsx
@@ -81,8 +81,11 @@ function scopeNeedsTarget(scope: BudgetRule['scope']): boolean {
   return scope === 'user' || scope === 'team' || scope === 'role';
 }
 
-/** A rule only enforces if at least one positive limit is set. With every
- *  limit unset (or zero) the enforcer resolves no cap and the rule is dead. */
+/** A rule only enforces meaningfully if at least one *positive* limit is set.
+ *  With every limit unset the enforcer resolves no cap and the rule is dead.
+ *  A limit of `0` is worse than unset: the enforcer treats it as the strictest
+ *  cap (`projectedCost >= 0` etc. is always true) and blocks every request — so
+ *  a set-but-non-positive limit is rejected per-field in `validateBudgetRule`. */
 function hasUsableLimit(rule: BudgetRule): boolean {
   return (
     (rule.maxTokens != null && rule.maxTokens > 0) ||
@@ -113,13 +116,18 @@ function validateBudgetRule(rule: BudgetRule, t: TFunction): BudgetRuleErrors {
     errors.scopeId = t('budgets.targetRequired');
   }
 
-  if (rule.maxTokens != null && rule.maxTokens < 0) {
+  // A *set* limit must be positive. `0` (typed directly — `"0"` is truthy so it
+  // survives the `onChange` guard — or a sub-cent cost that rounds down to `0`
+  // cents) is not "no limit": the enforcer reads it as the strictest possible
+  // cap and blocks every request. Reject `<= 0` to mirror `hasUsableLimit`'s
+  // `> 0` premise so a per-field zero can't slip past as a usable limit.
+  if (rule.maxTokens != null && rule.maxTokens <= 0) {
     errors.maxTokens = t('budgets.invalidTokenLimit');
   }
-  if (rule.maxCostCents != null && rule.maxCostCents < 0) {
+  if (rule.maxCostCents != null && rule.maxCostCents <= 0) {
     errors.maxCostCents = t('budgets.invalidCostLimit');
   }
-  if (rule.maxRequests != null && rule.maxRequests < 0) {
+  if (rule.maxRequests != null && rule.maxRequests <= 0) {
     errors.maxRequests = t('budgets.invalidMaxRequests');
   }
   if (
--- a/services/platform/app/features/settings/governance/components/budget-editor.tsx
+++ b/services/platform/app/features/settings/governance/components/budget-editor.tsx
@@ -14,7 +14,8 @@ import {
   TableRow,
 } from '@tale/ui/table';
 import { Text } from '@tale/ui/text';
-import { Pencil, Plus, Trash2, Wallet } from 'lucide-react';
+import type { TFunction } from 'i18next';
+import { Pencil, Plus, Trash2, Wallet, XCircle } from 'lucide-react';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { ConfirmDialog } from '@/app/components/ui/dialog/confirm-dialog';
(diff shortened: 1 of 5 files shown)


larryro:
## Desk review — PR #2177 (issue #2061)

The fix correctly closes the two dead-rule shapes the issue names (target-required for user/team/role scopes; at-least-one-positive-limit), wires inline errors with good touch/submit-attempt UX, and adds en/de/fr keys. **CI is fully green** (all required checks pass) and the unit suite passes (`bunx vitest run --config vitest.ui.config.ts budget-editor.test.tsx` → 11 passed, run locally on this branch). But a confirmed correctness gap and a test gap block merge.

### 🔴 Blocking 1 — A per-field limit of `0` slips through and creates a "block-everything" rule
`budget-editor.tsx` `validateBudgetRule` (L116-134) only rejects `< 0` per field, while `hasUsableLimit` (L86-92) requires `> 0`. The cross-field check is satisfied by **any** positive field, so a rule like `{ maxTokens: 1000000, maxCostCents: 0 }` passes validation and saves.

This `0` is reachable through plausible input:
- The numeric `onChange` guards use `e.target.value ? … : undefined`. **`"0"` is truthy in JS**, so typing `0` in the token/cost/requests field stores the number `0` (not `undefined`).
- A positive sub-cent cost (e.g. `0.001`) rounds via `Math.round(Number(v) * 100)` to `0` cents.

The enforcer (`convex/governance/budget_enforcement.ts` `checkRuleAgainstUsage`, L312-345) treats `0` as the strictest cap — `projectedCost >= 0` / `usage.totalTokens >= 0` / `projectedRequests >= 0` are **always true** — so the rule blocks *every* request for that scope. `resolveField` returns `0` (not "no cap") because `0 != null`. The table renders it as `$0.00`. This is the exact "surprising enforcement" class the PR set out to prevent, and the doc comment at L84-85 ("with every limit unset (or zero) the enforcer resolves no cap") is false for the per-field-zero case.

Verified by reading the enforcer and by executing the editor's converters + `hasUsableLimit` + the enforcer comparison.

**Fix:** in `validateBudgetRule`, reject a non-positive value for each limit field that is set (treat `<= 0` as invalid, mirroring `hasUsableLimit`'s `> 0` premise) — or normalize `0` to `undefined` in the `onChange` handlers so it counts as "unset" and the at-least-one-limit rule then applies. Apply to `maxTokens`, `maxCostCents`, and `maxRequests`, and correct the L84-85 comment.

### 🔴 Blocking 2 — The headline branch (#2061's named case) has no test
Only two tests were added: the no-limit block and a default-scope save. The **primary** fix — a `user`/`team`/`role` scope with **no target** being blocked — is untested in *both* directions (blocked-when-empty and saves-when-a-target-is-picked). A regression that inverted `scopeNeedsTarget` would pass the suite.

**Fix:** add (a) a negative test — select `user`/`team`/`role` scope, leave the target empty, confirm → asserts the target-required error and that the dialog stays open / nothing saves; (b) a positive test — scoped rule with a target + a limit saves; and (c) a test that a limit field set to `0` is rejected (covers Blocking 1).

### Non-blocking notes (not required to merge)
- **Server-side defense-in-depth.** `budgetRuleSchema` has no `.refine()`, so the server still accepts a dead/zero rule from a programmatic write. The issue explicitly frames this as optional ("confirm no seeded limit-less placeholder rules first"), so it's a reasonable follow-up — but note Blocking 1 means a `0`-limit rule is currently persisted server-side too. Negative limits are already rejected server-side by `z.number().nonnegative()`, so the client negative checks are correctly defense/UX-only.
- **Editing a legacy dead rule** is now blocked from re-saving until fixed (`handleSubmit` re-validates the draft regardless of origin). Likely the intended spirit of #2061; flagging as a behavior change for awareness.
- **a11y:** the cross-field `FieldError` is not associated to its control via `aria-describedby` (the `Select`/`SearchableSelect` only get a boolean `error`). `role="alert"` still announces it; minor. The `role="alert"` + `aria-live="polite"` pairing is copied verbatim from `Input`.
- **NaN** input is caught by the server zod (`nonnegative()` rejects NaN) → error toast, not a silent save; an inline `Number.isFinite` guard would improve UX.
- **Nits:** `handleSubmit` re-calls `validateBudgetRule` instead of reusing the memoized `errors`; limit-field membership is enumerated in two places (`updateDraft` and `hasUsableLimit`).

---
**NOT READY — CHANGES REQUIRED** (fix Blocking 1 and Blocking 2 above).

services/platform/app/features/settings/governance/components/budget-editor.tsx
```diff
@@ -73,6 +74,68 @@ function emptyRule(): BudgetRule {
...
+
+  if (rule.maxTokens != null && rule.maxTokens < 0) {
+    errors.maxTokens = t('budgets.invalidTokenLimit');
+  }
+  if (rule.maxCostCents != null && rule.maxCostCents < 0) {
+    errors.maxCostCents = t('budgets.invalidCostLimit');
+  }
+  if (rule.maxRequests != null && rule.maxRequests < 0) {
+    errors.maxRequests = t('budgets.invalidMaxRequests');
+  }
```
larryro, line 132: Per-field zero slips through: these checks only reject `< 0`, but a limit field set to `0` is valid here while `hasUsableLimit` (L86-92) requires `> 0`. Because `"0"` is truthy, typing `0` stores the number `0` (and a sub-cent cost rounds to `0`), so e.g. `{maxTokens:1000000, maxCostCents:0}` passes validation and saves. The enforcer then treats `0` as the strictest cap (`projectedCost >= 0` is always true -> blocks every request). Reject `<= 0` per limit field when set (or normalize 0->undefined in onChange), and fix the L84-85 comment which is false for a per-field zero.

larryro:
## Desk Review — #2061 (budget editor dead-rule validation)

**Verdict: READY TO MERGE.**

### CI & verification (run on branch `tale/xs7c1pge4axepx5xjedydyd26s89b9n1`)
- **CI:** 46 checks **pass**, 5 conditional **skips** (fork-only / docs-only jobs), **0 failed, 0 pending** → green.
- **Tests:** `bunx vitest --run --config vitest.ui.config.ts budget-editor.test.tsx` → **14 passed**.
- **Typecheck:** `bunx tsc --noEmit` → clean (exit 0).
- **Lint:** `bunx oxlint --type-aware` on the two changed source files → 0 warnings, 0 errors.

### What the fix does well
- **Both halves of the issue are resolved at create-time in the editor.** `validateBudgetRule` blocks (a) user/team/role scopes with no `scopeId` (target-required) and (b) rules with no positive limit (`hasUsableLimit`), mirroring the constraints the enforcer in `budget_enforcement.ts` relies on. Save is gated by a *fresh* `validateBudgetRule(draft, t)` call in `handleSubmit` (not the stale memo), so the guard can't be fooled by a last-keystroke race.
- **Per-field `<= 0` rejection is a correct, important catch beyond the literal ask.** A `0` limit is *not* "no limit": `resolveField` treats `0` as a real value and `checkRuleAgainstUsage` uses `usage >= limit`, so `maxTokens: 0` makes `totalTokens >= 0` always true and blocks every request. Rejecting set-but-non-positive limits is right.
- **Inline parity** matches the repo convention (`errorMessage` on `Input`, `error` on `Select`/`SearchableSelect`); the `FieldError` component for the cross-field target/limit messages is justified (no per-field `Input` owns them) and styled/announced consistently (`role="alert"`).
- **i18n:** 6 new keys added in en/de/fr, naming consistent, no hardcoded strings left.
- **Errors reveal on touch / submit-attempt**, so a freshly-opened dialog isn't pre-filled with red; dialog state resets correctly on open.

### Non-blocking notes (do not block this merge; worth a follow-up)
1. **Server-side defense-in-depth not added.** `budgetRuleSchema` still has every field `.optional()`/`nonnegative()` with no `.refine()`, so a dead rule remains persistable programmatically (direct mutation / config import) bypassing the editor. The issue itself framed `.refine()` as optional defense-in-depth ("would also catch programmatic writes") and warned to confirm no seeded limit-less placeholder rules first — so client-only is an acceptable scoping of #2061. Recommend a follow-up `.refine()` (scope→target, ≥1 positive limit) shared by client parse and `saveGovernancePolicy`.
2. **Pre-existing cascade (not introduced here, outside the issue's two cases):** `updateDraft` only clears `scopeId` when switching to `default`/`org`. Switching *between* target scopes (e.g. user→team) keeps the old `scopeId`; the new `SearchableSelect` shows empty but the draft retains e.g. a user-id under a `team` rule, which passes validation (truthy `scopeId`) and saves a rule that never matches at enforcement — a dead rule of a different flavor. Cheap hardening: clear `scopeId` on any scope change, or validate `scopeId` membership against the target options.
3. **Test gaps (core branches are covered; these are nice-to-have):** only the `role` scope is exercised for target-required (`user`/`team` go through `SearchableSelect` and are untested); `warningThresholdPercent` out-of-range, negative limits, and the sub-cent `maxCostCents` rounding-to-0 path have validation code but no test. The five #2061 headline branches (no-limit, default valid-save, scoped no-target, scoped valid-save, per-field zero) are all tested and pass.

None of the above blocks merge: the reported issue is genuinely fixed, the change is minimal and convention-following, tests pass, and CI is green.

**READY TO MERGE**

PR #2182, opened 2026-06-27
fix: scope getByStorageIds to caller's org to close cross-tenant leak (#2027)

### LINKED ISSUE #2027: Bug: getByStorageIds missing RLS — allows cross-tenant file metadata and transcript access
## Area

`services/platform/convex/file_metadata/queries.ts` — public Convex query `getByStorageIds` (line 66).

## Steps to reproduce

1. Sign in as user A (member of org A).
2. Obtain a `storageId` (`_storage` document ID) that belongs to a file uploaded by org B.
3. Call `api.fileMetadata.queries.getByStorageIds` with that storageId.
4. Receive the file's metadata row — including `fileName`, `contentType`, `transcript`, `ragStatus`, `ragError`, and processing metadata — with no org membership check.

## Expected

`getByStorageIds` should only return metadata for files belonging to an organization the caller is a member of. Lookups for out-of-tenant storage IDs should return no results.

## Actual

Any authenticated user can retrieve file metadata for any storage ID on the instance. The handler calls `getAuthUserIdentity` (confirms login) but then performs an unguarded `by_storageId` index lookup with no organizationId filter and no `getOrganizationMember` call.

The contrast: `getByDocumentId` (line 29 in the same file) accepts an explicit `organizationId` argument and scopes its query to `by_organizationId_and_documentId`. `getByStorageIds` takes no `organizationId` and performs no per-row org check.

## Evidence

- `services/platform/convex/file_metadata/queries.ts:66-152` — `getByStorageIds` handler
- `services/platform/convex/file_metadata/queries.ts:29-64` — `getByDocumentId` as the correctly-scoped reference

## Notes

**Severity: cross-tenant data exposure.** The response includes `transcript` (full audio/video transcription text), `fileName`, and RAG processing details. In a multi-tenant deployment, a user knowing or guessing a storage ID (e.g. from a Convex storage URL or API response) can retrieve file content metadata and transcripts belonging to another organization.

Fix: add an `organizationId` argument and a `getOrganizationMember` check (matching `getByDocumentId`), then filter results to only include rows where `meta.organizationId === args.organizationId`.

larryro:
## Summary

`fileMetadata.queries.getByStorageIds` confirmed login via `getAuthUserIdentity` but then performed an unguarded `by_storageId` index lookup with **no organization scoping**. Any authenticated user could pass a foreign `_storage` id and read another tenant's `fileName`, `contentType`, `transcript`, and RAG/processing metadata — cross-tenant data exposure (issue #2027).

## Fix

- Add a required `organizationId` argument to `getByStorageIds` (mirrors `getByDocumentId`).
- Verify the caller is a member of that org via `getOrganizationMember`, soft-failing to `[]` on non-membership to preserve the reactive-subscription contract (a non-member must not throw and break the chip UI).
- Filter every returned row by `meta.organizationId === args.organizationId`, so a foreign-tenant row is dropped even when the caller is a member of *some* org.
- Thread `organizationId` through all four call sites (`source-cards`, `file-displays`/`FileAttachmentDisplay`, `use-file-indexing-status`, `use-file-transcription-status`) and their parents (`chat-interface`, `use-assistant-chat`, `message-bubble`, `task-attachments` → `task-modal`); each `useQuery` skips until an org id is present.

## Tests

Added `convex/file_metadata/queries.test.ts` covering: unauthenticated → `[]`, non-member → `[]`, foreign-org row filtered out, in-tenant rows returned.

Verified locally: `tsc --noEmit`, `oxlint --type-aware`, server vitest (file_metadata suite + new test), and `migrations:check` all green.

Closes #2027

### DIFF
--- a/services/platform/app/features/automations/hooks/use-assistant-chat.ts
+++ b/services/platform/app/features/automations/hooks/use-assistant-chat.ts
@@ -77,13 +77,15 @@ export function useAssistantChat({
     removeAttachment,
     clearAttachments,
   } = useConvexFileUpload({ organizationId });
-  const { isIndexing, statusMap: indexingStatuses } =
-    useFileIndexingStatus(attachments);
+  const { isIndexing, statusMap: indexingStatuses } = useFileIndexingStatus(
+    attachments,
+    organizationId,
+  );
   const {
     isTranscribing,
     isQueryLoading: isTranscriptionQueryLoading,
     statusMap: transcriptionStatuses,
-  } = useFileTranscriptionStatus(attachments);
+  } = useFileTranscriptionStatus(attachments, organizationId);
   const [inputValue, setInputValue] = useState('');
   const [isPending, setIsPending] = useState(false);
   const [threadId, setThreadId] = useState<string | null>(null);
--- a/services/platform/app/features/chat/components/chat-interface.tsx
+++ b/services/platform/app/features/chat/components/chat-interface.tsx
@@ -228,15 +228,17 @@ export function ChatInterface({
     disableIndexing: isExternalAgentThread,
   });
 
-  const { isIndexing, statusMap: indexingStatuses } =
-    useFileIndexingStatus(attachments);
+  const { isIndexing, statusMap: indexingStatuses } = useFileIndexingStatus(
+    attachments,
+    organizationId,
+  );
 
   const {
     isTranscribing,
     isQueryLoading: isTranscriptionQueryLoading,
     statusMap: transcriptionStatuses,
     hasFailedAudioJobs,
-  } = useFileTranscriptionStatus(attachments);
+  } = useFileTranscriptionStatus(attachments, organizationId);
 
   const {
     jobs: videoLinkJobs,
--- a/services/platform/app/features/chat/components/message-bubble.tsx
+++ b/services/platform/app/features/chat/components/message-bubble.tsx
@@ -888,6 +888,7 @@ function MessageBubbleComponent({
                 <FileAttachmentDisplay
                   key={attachment.fileId}
                   attachment={attachment}
+                  organizationId={organizationId}
                   onImageClick={
                     galleryIdx >= 0 ? () => openGallery(galleryIdx) : undefined
                   }
--- a/services/platform/app/features/chat/components/message-bubble/file-displays.tsx
+++ b/services/platform/app/features/chat/components/message-bubble/file-displays.tsx
@@ -173,9 +173,11 @@ export function FileTypeIcon({
 
 export const FileAttachmentDisplay = memo(function FileAttachmentDisplay({
   attachment,
+  organizationId,
   onImageClick,
 }: {
   attachment: FileAttachment;
+  organizationId?: string;
   onImageClick?: () => void;
 }) {
   const { t } = useT('chat');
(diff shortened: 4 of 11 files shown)


larryro:
## Desk Review — #2182 (closes #2027): scope `getByStorageIds` to caller's org

**Verdict: READY TO MERGE.**

Reviewed via two rounds (6 breadth subagents across correctness / robustness / call-site threading / tests / elegance / boundaries, then independent confirmation of every candidate finding).

### Evidence
- **CI:** all required checks **green** (Type check, Lint, Unit, UI, Build platform/convex, 16 Playwright platform shards, Migrations, Opengrep, Format, Knip, etc.). Only fork-PR / docs-container / Trivy jobs skipped, as expected.
- **Tests (run locally):** `bunx vitest --run --project server convex/file_metadata/queries.test.ts` → **4 passed**; full `file_metadata` suite → **99 passed**.

### Correctness of the security fix — confirmed sound
`getByStorageIds` (convex/file_metadata/queries.ts:67-178) now closes all three cross-tenant vectors:
1. Authenticated **non-member** → `getOrganizationMember` throws `UnauthorizedError` → caught → `[]`.
2. Member of org A passing a **foreign storageId** (row owned by org B) → per-row filter `meta.organizationId !== args.organizationId` (line 151) drops it.
3. Member of org A passing **organizationId = org B** (not their org) → membership gate throws → `[]`.

The per-row filter is the load-bearing gate (the `by_storageId` index is global — schema.ts:137 — with no `(organizationId, storageId)` composite). `fileMetadata.organizationId` is a required `v.string()`, so the strict `!==` filter cannot be bypassed by a missing/undefined org on a row. Empty/whitespace `organizationId` fails the membership gate → `[]`. All four frontend callers were updated to thread `organizationId` and `skip` until it is present; no internal callers remain.

### Non-blocking notes (follow-ups, not required to merge)
1. **Broad catch vs. repo convention.** The membership `try/catch` (lines 132-138) swallows *all* errors → `[]`, whereas the dominant pattern (`tasks/queries.ts:244-249`, `sandbox/session_queries_public.ts:134-139,263-270`) narrows on `instanceof UnauthorizedError` and re-throws. It fails closed (no leak) and satisfies the no-empty-catch rule, and `tts/queries.ts:259-271` is a precedent for fail-closed-on-any-error — but narrowing would let transient Better-Auth/infra errors surface/retry instead of silently rendering an empty chip. Consistency nit.
2. **Test hardening.** Consider adding cases for a row with `organizationId === undefined` (proves the filter fails closed on malformed data) and the disabled-member branch. Current coverage is adequate; these are regression guards.
3. **Elegance.** `FileAttachmentDisplay`'s `organizationId?: string` is optional while the two status hooks require it; making it required would turn the silent `skip` into a compile-time guarantee. An `(organizationId, storageId)` composite index would encode tenant scope in the index rather than a droppable `if`.
4. **Out of scope (file a separate issue):** `files/queries.ts` `getFileUrl`/`getFileUrls` (lines 8-55) have the same shape as #2027 — authenticated-only, no org scoping, returning signed download URLs for arbitrary `_storage` ids. Not touched by this PR.

A candidate regression (discussion-thread-view / shared-chat-view render `MessageBubble` without `organizationId`) was investigated and **dismissed**: neither view ever mounts a media `FileAttachmentDisplay` (discussion messages don't carry the enriched attachment marker; the shared message shape has no `attachments` field), so no transcript chip exists to break there.

**READY TO MERGE** — the bar is met, the fix fully resolves #2027 with defense-in-depth, tests pass, and CI is green. The items above are optional follow-ups.

PR #2185, opened 2026-06-27
fix(convex): add thread RLS to approval-by-thread queries (#2025)

### LINKED ISSUE #2025: Bug: getPendingIntegrationApprovalsForThread, getWorkflowCreationApprovalsForThread, and getHumanInputRequestsForThread missing RLS — allows cross-tenant approval data access
## Area

`services/platform/convex/approvals/queries.ts` — three public Convex queries: `getPendingIntegrationApprovalsForThread` (line 213), `getWorkflowCreationApprovalsForThread` (line 242), and `getHumanInputRequestsForThread` (line 271).

## Steps to reproduce

1. Sign in as user A (member of org A).
2. Obtain (or guess) a `threadId` belonging to org B.
3. Call `api.approvals.queries.getPendingIntegrationApprovalsForThread` with that threadId.
4. Receive all pending integration-operation approval rows for org B's thread — no org membership check is performed.

Same applies to `getWorkflowCreationApprovalsForThread` and `getHumanInputRequestsForThread`.

## Expected

Only members of the org that owns the thread should be able to read its approval rows. A call with an out-of-tenant `threadId` should return `[]`.

## Actual

Any authenticated user (from any org on the same instance) can read all approval rows for any thread by knowing its `threadId`. The handlers check `getAuthUserIdentity` (user is logged in) but never call `getOrganizationMember`, `canAccessThread`, or any equivalent RLS check.

The contrast is explicit in the same file: `listResolvedHumanInputRequestsByThread` (line 174) correctly calls `getOrganizationMember` AND then cross-checks `approval.organizationId !== args.organizationId` before returning rows. The three failing handlers omit both checks.

## Evidence

- `services/platform/convex/approvals/queries.ts:213-298` — all three handlers
- `services/platform/convex/approvals/queries.ts:174-211` — `listResolvedHumanInputRequestsByThread` as the correctly-gated reference

## Notes

**Severity: cross-tenant data exposure.** The `approvalItemValidator` includes `metadata: v.any()`, `resourceId`, `wfExecutionId`, `stepSlug`, and `organizationId` — sufficient to expose integration operation targets, workflow identifiers, and human-input prompt text from another tenant's active agent runs.

Fix: add `canAccessThread(ctx, args.threadId, authUser)` (or `getOrganizationMember` + an `organizationId` arg parity with the resolved handler) to all three handlers before the table scan.

larryro:
## Summary

`getPendingIntegrationApprovalsForThread`, `getWorkflowCreationApprovalsForThread`, and `getHumanInputRequestsForThread` in `services/platform/convex/approvals/queries.ts` only checked that the caller was authenticated — they never verified org membership. Any logged-in user could read another tenant's pending integration-operation, workflow-creation, or human-input-request approval rows (including `metadata`, `resourceId`, `wfExecutionId`, `stepSlug`, and `organizationId`) just by knowing/guessing a `threadId`.

## Fix

Each of the three handlers now calls `canAccessThread(ctx, args.threadId, authUser)` before scanning the `approvals` table and returns `[]` when the thread is not accessible to the caller — mirroring the established RLS pattern used by `listResolvedHumanInputRequestsByThread` and the thread-scoped queries elsewhere in the codebase (e.g. `thread_files`, `user_memories`). As defence in depth, each row is also dropped when `approval.organizationId` diverges from the thread's org.

These queries take no `organizationId` arg and have no in-repo callers to update, so the threadId-based `canAccessThread` gate is the least-disruptive correct fix.

## Tests

Added `services/platform/convex/approvals/queries_thread_rls.test.ts` (15 cases, 5 per handler): unauthenticated → `[]`, out-of-tenant/inaccessible threadId → `[]` (and `canAccessThread` invoked), accessible thread returns matching rows, cross-org rows dropped, and other resourceTypes ignored.

- `vitest --run --project server` (approvals): 36 passed
- `tsc --noEmit`: clean

Closes #2025

### DIFF
--- a/services/platform/convex/approvals/queries.ts
+++ b/services/platform/convex/approvals/queries.ts
@@ -4,6 +4,7 @@ import { v } from 'convex/values';
 import { query } from '../_generated/server';
 import { DEFAULT_COUNT_CAP } from '../lib/helpers/count_items_in_org';
 import { getAuthUserIdentity, getOrganizationMember } from '../lib/rls';
+import { canAccessThread } from '../lib/rls/auth/can_access_thread';
 import { UnauthorizedError } from '../lib/rls/errors';
 import * as ApprovalsHelpers from './helpers';
 import { listApprovalsPaginated as listApprovalsPaginatedHelper } from './list_approvals_paginated';
@@ -222,6 +223,14 @@ export const getPendingIntegrationApprovalsForThread = query({
       return [];
     }
 
+    // RLS: only members of the thread's org may read its approval rows. A
+    // guessed/out-of-tenant threadId resolves to no accessible thread and
+    // returns []. Capture the thread's org to cross-check each row below.
+    const thread = await canAccessThread(ctx, args.threadId, authUser);
+    if (!thread) {
+      return [];
+    }
+
     const approvals = [];
     for await (const approval of ctx.db
       .query('approvals')
@@ -230,6 +239,10 @@ export const getPendingIntegrationApprovalsForThread = query({
         continue;
       }
       if (args.messageId && approval.messageId !== args.messageId) {
+        continue;
+      }
+      // Defence in depth: drop any row whose org diverges from the thread's.
+      if (approval.organizationId !== thread.organizationId) {
         continue;
       }
       approvals.push(approval);
@@ -251,6 +264,14 @@ export const getWorkflowCreationApprovalsForThread = query({
       return [];
     }
 
+    // RLS: only members of the thread's org may read its approval rows. A
+    // guessed/out-of-tenant threadId resolves to no accessible thread and
+    // returns []. Capture the thread's org to cross-check each row below.
+    const thread = await canAccessThread(ctx, args.threadId, authUser);
+    if (!thread) {
+      return [];
+    }
+
     const approvals = [];
     for await (const approval of ctx.db
       .query('approvals')
@@ -259,6 +280,10 @@ export const getWorkflowCreationApprovalsForThread = query({
         continue;
       }
       if (args.messageId && approval.messageId !== args.messageId) {
+        continue;
+      }
+      // Defence in depth: drop any row whose org diverges from the thread's.
+      if (approval.organizationId !== thread.organizationId) {
         continue;
       }
       approvals.push(approval);
@@ -280,6 +305,14 @@ export const getHumanInputRequestsForThread = query({
       return [];
     }
 
+    // RLS: only members of the thread's org may read its approval rows. A
+    // guessed/out-of-tenant threadId resolves to no accessible thread and
+    // returns []. Capture the thread's org to cross-check each row below.
+    const thread = await canAccessThread(ctx, args.threadId, authUser);
+    if (!thread) {
+      return [];
+    }
+
     const approvals = [];
     for await (const approval of ctx.db
       .query('approvals')
@@ -288,6 +321,10 @@ export const getHumanInputRequestsForThread = query({
         continue;
       }
       if (args.messageId && approval.messageId !== args.messageId) {
+        continue;
+      }
+      // Defence in depth: drop any row whose org diverges from the thread's.
+      if (approval.organizationId !== thread.organizationId) {
         continue;
       }
       approvals.push(approval);
--- a/services/platform/convex/approvals/queries.ts
+++ b/services/platform/convex/approvals/queries.ts
@@ -242,7 +242,12 @@ export const getPendingIntegrationApprovalsForThread = query({
         continue;
       }
       // Defence in depth: drop any row whose org diverges from the thread's.
-      if (approval.organizationId !== thread.organizationId) {
+      // Org-less threads (canAccessThread returns no organizationId) have
+      // nothing to diverge from, so skip the check rather than drop every row.
+      if (
+        thread.organizationId &&
+        approval.organizationId !== thread.organizationId
+      ) {
         continue;
       }
       approvals.push(approval);
@@ -283,7 +288,12 @@ export const getWorkflowCreationApprovalsForThread = query({
         continue;
       }
       // Defence in depth: drop any row whose org diverges from the thread's.
-      if (approval.organizationId !== thread.organizationId) {
+      // Org-less threads (canAccessThread returns no organizationId) have
+      // nothing to diverge from, so skip the check rather than drop every row.
+      if (
+        thread.organizationId &&
+        approval.organizationId !== thread.organizationId
+      ) {
         continue;
       }
       approvals.push(approval);
@@ -324,7 +334,12 @@ export const getHumanInputRequestsForThread = query({
         continue;
       }
       // Defence in depth: drop any row whose org diverges from the thread's.
-      if (approval.organizationId !== thread.organizationId) {
+      // Org-less threads (canAccessThread returns no organizationId) have
+      // nothing to diverge from, so skip the check rather than drop every row.
+      if (
+        thread.organizationId &&
+        approval.organizationId !== thread.organizationId
+      ) {
         continue;
       }
       approvals.push(approval);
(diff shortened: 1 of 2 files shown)


larryro:
## Desk review — RLS fix for approval-by-thread queries (#2025)

**Verdict: NOT READY — one change required.** The core security fix is correct and complete, but a defence-in-depth check is mis-implemented on a reachable branch and can silently drop a legitimate user's rows.

### CI & tests (observed)
- **CI: green** — all required checks pass (`gh pr checks`): Analyze, Build platform/convex, Lint, Format, Knip, Opengrep, Migrations, and all 16 Playwright shards passed; only `Docs container test` is `skipping` (expected).
- **Tests: pass** — ran `bunx vitest --run --project server approvals` in `services/platform` → **36 passed**; new file `approvals/queries_thread_rls.test.ts` alone → **15 passed**.

### What's right
- The three handlers now gate on `canAccessThread(ctx, args.threadId, authUser)` and return `[]` when it resolves to no accessible thread — this correctly closes the reported cross-tenant leak: a non-member guessing a `threadId` can no longer read another org's approvals. Verified `canAccessThread` reliably returns `null` for a non-owner / non-member / non-shared thread.
- No sibling leak left behind: every other approval-reading export in `convex/approvals/queries.ts` is gated by `getOrganizationMember(organizationId)` or `canAccessThread`. The audit of all `approvals`-table read paths came back clean.
- Tests follow the repo's established handler-mock pattern (matches `thread_files/queries.test.ts`), cover all three handlers, and would fail if the gate or the cross-org filter were reverted.

### BLOCKING — defence-in-depth check breaks for org-less threads
`services/platform/convex/approvals/queries.ts:245`, `:286`, `:327`

```js
if (approval.organizationId !== thread.organizationId) continue;
```

`canAccessThread` intentionally returns the thread metadata for an **org-less thread** to its owner (`lib/rls/auth/can_access_thread.ts:67` — "org-less thread: owner access, no org to check"), so `thread.organizationId` is `undefined` on that branch. The `approvals` schema makes `organizationId` a required string (`approvals/schema.ts:7`), so for an org-less thread the comparison is always `"<org>" !== undefined` → **true** → **every** approval row is dropped and the handler returns `[]`.

This is a behavior change from `main` (which returned all matching rows with no org check) and contradicts the line's own comment ("drop any row whose org **diverges**"): on an org-less thread nothing diverges, yet everything is dropped. Org-less threads are a real, supported state (e.g. `createChatThread(ctx, userId, undefined, 'general')` in `openai_compat/internal_mutations.ts`). The bug fails *safe* (it over-restricts, never leaks), but it is a latent correctness regression on a reachable branch.

**Fix (all three sites):** guard the org-less case so the check only drops genuinely divergent rows:
```js
if (thread.organizationId && approval.organizationId !== thread.organizationId) continue;
```
Add a test case: accessible org-less thread (`canAccessThread` → `{ threadId, organizationId: undefined }`) returns its matching rows instead of `[]`.

### Non-blocking notes (not required for merge)
- The three handlers call `canAccessThread` without the optional `expectedOrgId`, so active-org coherence isn't enforced — a user who belongs to both org A and org B and has switched to B could still read org-A thread approvals via these queries. This is NOT the reported bug (the caller is a legitimate member of the thread's org, not a cross-tenant non-member) and the queries take no `organizationId` arg, consistent with sibling thread-scoped queries. Worth a follow-up, not a blocker.
- Two RLS patterns now coexist in this file (`getOrganizationMember(organizationId)` in `listResolvedHumanInputRequestsByThread` vs `canAccessThread(threadId)` in the three new handlers). Both are correct; just a consistency note.

services/platform/convex/approvals/queries.ts
```diff
@@ -232,6 +241,10 @@ export const getPendingIntegrationApprovalsForThread = query({
       if (args.messageId && approval.messageId !== args.messageId) {
         continue;
       }
+      // Defence in depth: drop any row whose org diverges from the thread's.
+      if (approval.organizationId !== thread.organizationId) {
```
larryro: BLOCKING (org-less thread regression): `thread.organizationId` is `undefined` for an org-less thread that `canAccessThread` returns to its owner (see `lib/rls/auth/can_access_thread.ts:67`). Since `approvals.organizationId` is always a string (`approvals/schema.ts:7`), this comparison is always true on that branch and drops EVERY row → handler returns `[]`. Guard it: `if (thread.organizationId && approval.organizationId !== thread.organizationId) continue;`

services/platform/convex/approvals/queries.ts
```diff
@@ -261,6 +282,10 @@ export const getWorkflowCreationApprovalsForThread = query({
...
```
larryro: BLOCKING (org-less thread regression): `thread.organizationId` is `undefined` for an org-less thread that `canAccessThread` returns to its owner (see `lib/rls/auth/can_access_thread.ts:67`). Since `approvals.organizationId` is always a string (`approvals/schema.ts:7`), this comparison is always true on that branch and drops EVERY row → handler returns `[]`. Guard it: `if (thread.organizationId && approval.organizationId !== thread.organizationId) continue;`

services/platform/convex/approvals/queries.ts
```diff
@@ -290,6 +323,10 @@ export const getHumanInputRequestsForThread = query({
...
```
larryro: BLOCKING (org-less thread regression): `thread.organizationId` is `undefined` for an org-less thread that `canAccessThread` returns to its owner (see `lib/rls/auth/can_access_thread.ts:67`). Since `approvals.organizationId` is always a string (`approvals/schema.ts:7`), this comparison is always true on that branch and drops EVERY row → handler returns `[]`. Guard it: `if (thread.organizationId && approval.organizationId !== thread.organizationId) continue;`

larryro:
## Desk Review — PR #2185 (Closes #2025): thread RLS on approval-by-thread queries

**Verdict: READY TO MERGE.**

### CI
All checks **green** (`gh pr checks` — every job `pass`; the only non-pass entries are legitimately `skipping` fork-PR/container/docs jobs that don't run on this PR).

### Tests (run locally on this branch)
- `bunx vitest --run --project server approvals/queries_thread_rls.test.ts` → **18 passed** (6 cases × 3 handlers).
- `bunx vitest --run --project server approvals/` → **4 files, 39 passed** (no regressions).

### What the fix does
Each of `getPendingIntegrationApprovalsForThread`, `getWorkflowCreationApprovalsForThread`, `getHumanInputRequestsForThread` now calls `canAccessThread(ctx, args.threadId, authUser)` after the auth gate and returns `[]` when the thread is not accessible, plus a per-row defence-in-depth drop of any `approval.organizationId` that diverges from the thread's org (skipped for org-less threads).

### Review findings (two rounds, breadth + confirmation)
- **Correctness ✓** — All branches walked: unauthenticated → `[]` (no thread lookup); guessed/out-of-tenant threadId → `canAccessThread` null → `[]`; accessible org thread → only same-org rows; messageId filter unaffected.
- **Security ✓** — Closes the cross-tenant hole. No unfixed sibling: the only other public approval-by-thread query (`listResolvedHumanInputRequestsByThread`) is already org-gated; `getApprovalsForThread` and the other `by_threadId` approval reads are `internalQuery`/`internalMutation` (not client-reachable). The org-less skip path is **safe** — `canAccessThread` grants org-less threads only to their owner (owner branch requires `metadata.userId === authUser.userId`; both non-owner branches require `metadata.organizationId`), so no other tenant can reach those rows.
- **Robustness ✓** — Read-only queries; `canAccessThread` returns `null` (does not throw) on denial; the extra `threadMetadata`/membership read runs in parallel via `Promise.all` and adds no scaling hazard atop the existing `by_threadId` scan.
- **Elegance ✓ (nit)** — The RLS block + org guard are triplicated verbatim across the three handlers. Consistent with the codebase's inline-guard style (thread_files, user_memories), so acceptable; a shared helper would be a minor improvement, not required.
- **Tests ✓ (minor gap)** — Coverage is solid across every RLS branch including org-less. The `messageId` filter branch (`if (args.messageId && approval.messageId !== args.messageId)`) is not exercised by any test. Non-blocking, but worth a follow-up case.

### Non-blocking follow-ups (do not gate merge)
1. Add a test exercising the `messageId` argument (match + mismatch) for at least one handler.
2. Optionally extract the repeated RLS/org-guard block into a small shared helper.

**READY TO MERGE.**

PR #2186, opened 2026-06-27
fix(platform): count in-flight uploads against chat attachment cap (#2026)

### LINKED ISSUE #2026: Bug: Chat attachment 10-file cap can be bypassed when rapid sequential uploads overlap in-flight
## Area

## Steps to reproduce

1. Start a new chat thread.
2. Rapidly drop two batches of 6 files each in quick succession (before the first batch finishes uploading).
3. The slot-count check runs at the start of \`uploadFiles\` and reads \`attachmentsRef.current.length\`, which only reflects *settled* uploads — not in-flight ones.
4. Both calls see \`length = 0\` and each approve their full 6-file batch.
5. Both batches upload in parallel; the chat ends up with 12 attached files, exceeding the 10-file cap.

## Expected

Each subsequent \`uploadFiles\` call should account for in-flight uploads and reject overflow files before they are uploaded.

## Actual

The slot check at line 230 reads \`attachmentsRef.current.length\`, which is backed by React state updated only when each individual upload *completes* (line 335: \`setAttachments((prev) => [...prev, attachment])\`). Files still uploading are tracked in \`uploadingFiles\` (string IDs) but their count is never subtracted from the available slots. Two rapid drops (or paste + drop) before any upload settles both see the stale count and both pass the guard, resulting in up to 2 × 10 = 20 uploads.

## Evidence

- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:230-231\` — slot check reads \`attachmentsRef.current.length\` (settled only)
- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:283\` — in-flight IDs added to \`uploadingFiles\` string array, never counted against slots
- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:335\` — \`setAttachments\` called only after successful upload (too late for the guard)
- \`services/platform/lib/shared/file-types.ts:521\` — \`CHAT_MAX_FILE_COUNT = 10\`

## Notes

Fix: add an \`inFlightCountRef\` that is incremented before upload starts and decremented on settle, then use \`attachmentsRef.current.length + inFlightCountRef.current\` as the slot-check denominator. Alternatively, gate the slot check under an async lock/semaphore. The deduplication and total-size checks have the same stale-ref exposure.

larryro:
## Summary

The chat attachment 10-file cap could be bypassed when two upload batches overlapped in-flight. The slot-count, deduplication, and total-size guards in `use-convex-file-upload.ts` all read `attachmentsRef.current`, which only reflects *settled* uploads. Two rapid `uploadFiles` calls (e.g. paste + drop, or two quick drops of 6 files each) before the first batch finished both saw a stale count and each admitted their full batch — ending with up to 12 (or 2×10 = 20) attachments.

## Fix

- Added an `inFlightRef` tracking admitted-but-unsettled uploads (`id`, `key`, `size`).
- The slot, dedup, and total-size guards now count in-flight files in addition to settled attachments.
- Each accepted file reserves a slot **synchronously**, before the first upload `await` yields control — since JS is single-threaded and there is no `await` between the guards and the reservation, a second overlapping call cannot interleave until the reservations are registered, closing the race.
- Reservations are released in the upload `finally` (on success *or* failure).

## Tests

Extended `use-convex-file-upload.cap.test.ts` with two regression cases:
- Two overlapping 6-file batches attach exactly 10 files and fire a `tooManyFiles` toast.
- A duplicate dropped while the first copy is still uploading is rejected as a duplicate.

`bunx vitest` (4/4 pass), `oxlint --type-aware` (0 warnings/errors), and `tsc --noEmit` (clean) all green.

Closes #2026

### DIFF
--- a/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
+++ b/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
@@ -4,6 +4,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { useUploadPolicy } from '@/app/features/settings/governance/hooks/queries';
 import { toast } from '@/app/hooks/use-toast';
+import { CHAT_MAX_FILE_COUNT } from '@/lib/shared/file-types';
 
 import { useConvexFileUpload } from './use-convex-file-upload';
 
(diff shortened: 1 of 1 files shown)


larryro:
## Desk Review — #2186 (fixes #2026): in-flight upload cap

**Verdict: NOT READY — changes required.** The core race reported in #2026 is genuinely closed (the two new regression tests fail on `origin/main` and pass here), but the fix introduces a new correctness defect in its own reservation bookkeeping, and leaves one confirmed residual window — both can still push the chat past `CHAT_MAX_FILE_COUNT` (10) under realistic input.

### CI & tests (observed)
- **CI: GREEN.** All required checks pass (`gh pr checks` — Lint, Format, Knip, Build platform, Typecheck-via-build, all 16 Playwright shards, Analyze, etc.). Only `Trivy`/`Opengrep OSS` are conditional skips, not failures.
- **Tests: PASS.** `cd services/platform && bunx vitest --run app/features/chat/hooks/use-convex-file-upload.cap.test.ts` → **4/4 passing**. Lint on the two changed files: `bunx oxlint --type-aware …` → 0 warnings / 0 errors.
- The two new regression tests are meaningful (not vacuous): the overlap test would yield 12 attachments on `origin/main`; the in-flight-dedup test would admit the duplicate. Both correctly assert the fixed behavior.

### What the fix gets right
- `inFlightRef` is reserved **synchronously**, with no `await` between the dedup/slot/size guards (lines 234–311) and the reservation registration (line 324), so two overlapping `uploadFiles` calls cannot both pass the guards before the first registers — the reported race is closed.
- All three guards the issue named (dedup, slot count, total size) now include in-flight files.
- Reservations are released in `finally` on both success and failure paths.

---

### BLOCKING 1 — Reservation `id` can collide; release-by-id then removes the wrong/extra entry

`use-convex-file-upload.ts:320`
```js
id: `${file.name}-${Date.now()}-${i}`,
```
The `id` omits `file.size`, but the dedup `key` (line 321) is `name:size`, so **two files with the same name and different size are both admitted** into `inFlightRef`. When two overlapping batches each carry such a file at the same index `i` in the same millisecond, their `id`s are identical. The release filter at `use-convex-file-upload.ts:397-399`:
```js
inFlightRef.current = inFlightRef.current.filter((r) => r.id !== fileId);
```
then removes **both** entries when the first upload settles. Net effect (independently confirmed): a **transient under-count** of in-flight files from first-settle until second-settle — *not* a permanent leak — during which a concurrent batch can over-admit and push attachments to **11–12**, re-opening the exact cap-bypass this PR fixes. The same `id` is used for `setUploadingFiles` (line 329 / 396), so `uploadingFiles`/`isUploading` is also corrupted (a still-uploading file is shown as complete).

**Required fix:** make the reservation identity collision-free. Simplest: include size in the id (`` `${file.name}-${file.size}-${Date.now()}-${i}` ``), or release by the dedup `key` (which is already unique within `inFlightRef`, since same-name-same-size is deduped away) by threading `key` into the `finally`.

### BLOCKING 2 — Count gap when a reservation is released before its attachment is observable

On success, `setAttachments(...)` (line 382) is an **async** React state update, but the `finally` removes the reservation from `inFlightRef` **synchronously** (lines 397–399). `attachmentsRef.current` is only refreshed on re-render (line 99). So between the synchronous release and React's commit there is a window where the settled file is in **neither** `attachmentsRef` **nor** `inFlightRef`. A concurrent `uploadFiles` whose pre-guard awaits (`detectMediaMime` loop ~120, audio-duration `Promise.all` ~165) resolve in that window will under-count by 1 at the slot guard (lines 260–263) and admit one extra file → up to **11** attachments. Independently confirmed. This is narrower than the originally reported bug (which required both batches to run *before the first settled* — that path is now closed); it is a residual window introduced by mixing a synchronous ref with async state.

**Required fix:** close the handoff so a file is continuously accounted for. E.g. track a synchronous reserved/settled count and only drop a reservation once the corresponding attachment is reflected in the count the guards read — rather than releasing the reservation while the settled attachment is still only a pending `setState`. A single accounting change can address both BLOCKING 1 and BLOCKING 2.

---

### Non-blocking notes (worth addressing, not gating)
1. **Size accounting inconsistency for images.** Reservation/guards use the *original* `file.size`, but settled attachments store the *compressed* size (`fileToUpload.size`, line 364/376) for images. The total-size guard therefore mixes pre- and post-compression sizes. Pre-existing dedup quirk, but now also affects the in-flight size sum. Consider documenting or normalizing.
2. **Test coverage gaps.** The new tests don't exercise (a) the upload-**failure** path releasing a reservation (fetch is always mocked `ok:true`), nor (b) the **total-size** overlap branch (files are 1 MB, far under the 200 MB cap). Both use the same `inFlightRef` pattern as the covered count guard, so risk is low, but a failure-release test and a size-overlap test would lock in the branches. A regression test for the same-name/different-size collision (BLOCKING 1) should be added with the fix.
3. **Parallel bookkeeping.** `uploadingFiles` (state) and `inFlightRef` (ref) now track the same lifecycle with two mechanisms keyed by the same id — kept in sync by hand. Acceptable given the ref's synchronicity is load-bearing, but a one-line note on why both exist would help future readers.

**Verdict: NOT READY — CHANGES REQUIRED** (fix BLOCKING 1 and BLOCKING 2 above; both can be resolved by a single collision-free, gap-free in-flight accounting change, plus a regression test for the same-name/different-size case).

larryro:
## Desk Review — #2186 (fixes #2026: in-flight chat attachment cap bypass)

**Verdict: ✅ READY TO MERGE.**

### CI & test evidence
- **CI: all green.** `gh pr checks` shows every required check passing (Analyze JS/TS, Build platform, Unit, UI, Type check, Lint, Format, Knip, Opengrep, Smoke test, all 16 Playwright platform shards, etc.). Only `skipping`/fork-only jobs are skipped — none failing or pending.
- **Tests: pass — I ran them.** `bunx vitest --run use-convex-file-upload.cap` → **6/6 passed** (2 pre-existing #2048 + 4 new #2026). (The PR description says "4/4"; the file actually has 6 — the 2 #2048 tests still pass, no regression.)
- `bunx tsc --noEmit` → clean (exit 0). `bunx oxlint --type-aware` on both changed files → 0 warnings / 0 errors.

### What I verified (two-round review, 5 breadth agents + confirmation pass)
The fix adds an `inFlightRef` of admitted-but-unsettled reservations and counts it in **all three** guards — slot count (`use-convex-file-upload.ts:288-291`), dedup (`:262-267`), and total-size (`:320-325`). Reservations are registered **synchronously before the first `await`** (`:347-353`), so a second overlapping `uploadFiles` call cannot interleave through the critical section — this genuinely closes the race in #2026.

- **No undercount gap.** On success the reservation is marked `settled` (`:411-418`) and only dropped by the `useEffect([attachments])` (`:106-110`) *after* the `setAttachments` commit lands in `attachmentsRef` — so an accepted file is always counted in `inFlightRef` or `attachmentsRef`, never neither. On failure the reservation is released immediately in `catch` (`:431-434`).
- **Collision-free ids.** `reservationSeqRef` monotonic counter replaces the old `Date.now()`-based id, which could collide across same-millisecond batches and drop both reservations on release. Test at `cap.test.ts:206` pins `Date.now` to prove the old scheme would collide and the cap still holds.
- **Real-world trigger covered.** Chat composer fires fire-and-forget `void uploadFiles(...)` from paste / picker / drop (`chat-input.tsx:787,834,843`), which can overlap — exactly the scenario the fix guards.

### Confirmed NON-blocking (independently re-verified, do not block merge)
1. **`clearAttachments` (`:513-522`) does not clear `inFlightRef`.** Verified this is *pre-existing* (in-flight uploads re-add via async `setAttachments` on `main` too) and *self-healing* (reservations release as uploads settle) — conservative, never bypasses the cap. Out of scope for #2026.
2. **Client-side-only cap.** `saveFileMetadata` enforces governance policy server-side but not `CHAT_MAX_FILE_COUNT`/`CHAT_MAX_TOTAL_SIZE`; acceptable for this issue's scope (a UX cap), not a regression.
3. Ref declarations (`:128-136`) sit *after* the `useEffect` that closes over them (`:106`) — verified benign (effect callback runs post-mount, no TDZ); purely a readability nit.

### Optional follow-ups (not required for merge)
- Add a regression test for the **total-size** guard under overlap (two batches whose combined size exceeds `CHAT_MAX_TOTAL_SIZE` while the first is in-flight). The slot/dedup branches of the same `inFlightRef` mechanism are tested; the size branch is a trivially-correct analogue but currently has no dedicated test.
- Consider moving the ref declarations above the `useEffect`, and trimming the ~28-line comment block down toward the density of the existing `retryInFlightRef` pattern in the same file.

The change is correct across all branches, robust, minimal, follows the file's own conventions, and is well-tested on the core race paths. **READY TO MERGE.**

PR #2187, opened 2026-06-27
fix(platform): run total-size check before slot-overflow toast (#2029)

### LINKED ISSUE #2029: Bug: Chat — misleading 'files not added' toast fires before total-size check, implying partial upload that never happens
## Area

## Steps to reproduce

1. Add ~8 large files (~20 MB each) to the chat composer (≈160 MB total).
2. Attempt to add 5 more files at once that together exceed 40 MB.
3. The slot cap (10 files) trims the batch to 2 files and shows: **"2 files were not added. You can attach up to 10 files per message."**
4. The user sees this and expects the 2 accepted files to upload.
5. Immediately after, the total-size check fires (160 MB existing + 40 MB incoming > 200 MB) and shows: **"Total attachment size cannot exceed 200 MB. Remove some files and try again."**
6. The function returns early — none of the 2 accepted files actually upload.

## Expected

Either:
- The slot-cap toast should not fire when the total-size check will also reject (checks should be reordered, or the batch-trim and total-size check should happen together before showing any toast).
- Or: the total-size toast should clarify that all files in the current batch were rejected, not just the oversized ones.

## Actual

Two toasts fire in rapid succession: the first misleads the user into thinking 2 files were accepted and queued, and the second then contradicts that with a blanket rejection. No files upload.

## Evidence

- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:249-258\` — slot-overflow toast fires unconditionally when \`acceptedFiles.length < deduped.length\`
- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:269-278\` — total-size check fires *after* the slot toast, returns early without uploading
- \`services/platform/lib/shared/file-types.ts:524\` — \`CHAT_MAX_TOTAL_SIZE = 200 MB\`

## Notes

Fix: run the total-size check before emitting the slot-overflow toast, so the slot-overflow toast only fires when the trimmed batch will actually be uploaded. Alternatively, merge both validations into a single pass that determines the final accepted set before showing any toast.

larryro:
## Summary

Fixes the misleading double-toast in the chat composer described in #2029.

When a batch of files overflowed both the **slot cap** (10 files/message) and the **total-size cap** (200 MB), the slot-overflow toast ("N files were not added…") fired *before* the total-size check. The user was led to believe the trimmed batch had been accepted, only for the total-size check to then reject the whole batch and return early — two contradictory toasts and **zero uploads**.

## Change

In `use-convex-file-upload.ts`, the **total-size check now runs before** the slot-overflow toast. The slot-overflow toast only fires once we know the trimmed batch will actually be uploaded:

- If the trimmed batch still exceeds the total cap → only the `totalSizeExceeded` toast fires, nothing uploads.
- If the trimmed batch fits under the total cap → the `tooManyFiles` slot-overflow toast fires and the accepted files upload, exactly as before.

No behavioral change to which files upload; only the ordering/emission of the toasts.

## Tests

Added two regression tests to `use-convex-file-upload.cap.test.ts`:
1. Slot overflow + total-size overflow → only the total-size toast fires, no uploads.
2. Slot overflow but within total cap → slot-overflow toast still fires, accepted files upload.

All 4 tests in the file pass; lint clean on the changed files.

Closes #2029

### DIFF
--- a/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
+++ b/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
@@ -346,3 +346,87 @@ describe('useConvexFileUpload — in-flight slot cap (#2026)', () => {
     await waitFor(() => expect(result.current.attachments).toHaveLength(1));
   });
 });
+
+// ---------------------------------------------------------------------------
+// Regression coverage for #2029: when a batch overflows the slot cap AND the
+// trimmed batch would still push the running total past CHAT_MAX_TOTAL_SIZE,
+// the total-size check must run *before* the slot-overflow toast. Otherwise
+// the user sees a misleading "N files were not added" toast (implying the
+// trimmed batch was accepted) immediately followed by a blanket total-size
+// rejection — two contradictory toasts and zero uploads.
+// ---------------------------------------------------------------------------
+describe('useConvexFileUpload — slot-overflow vs total-size ordering (#2029)', () => {
+  it('shows only the total-size toast (not the slot-overflow toast) and uploads nothing when the trimmed batch still exceeds the total cap', async () => {
+    const { result } = renderHook(() => useConvexFileUpload(config));
+
+    // Pre-fill 8 of the 10 slots with 21MB files → 168MB used, 2 slots left.
+    await act(async () => {
+      await result.current.uploadFiles(
+        Array.from({ length: 8 }, (_, i) =>
+          makeAudioFile(`existing-${i}.mp3`, 21 * MB),
+        ),
+      );
+    });
+    await waitFor(() => expect(result.current.attachments).toHaveLength(8));
+    toastMock.mockClear();
+    saveFileMetadata.mockClear();
+
+    // Add 5 more 21MB files: the slot cap trims the batch to 2 (2 slots left),
+    // but 168MB + 42MB = 210MB > 200MB total cap, so the whole batch is
+    // rejected by the total-size check.
+    await act(async () => {
+      await result.current.uploadFiles(
+        Array.from({ length: 5 }, (_, i) =>
+          makeAudioFile(`incoming-${i}.mp3`, 21 * MB),
+        ),
+      );
+    });
+
+    // Nothing new was uploaded.
+    expect(result.current.attachments).toHaveLength(8);
+    expect(saveFileMetadata).not.toHaveBeenCalled();
+
+    // The total-size toast fired...
+    expect(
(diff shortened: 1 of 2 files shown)


larryro:
## Desk review — #2029 fix (slot-overflow vs total-size toast ordering)

**Verdict: READY TO MERGE.**

### What was reviewed
The reorder in `services/platform/app/features/chat/hooks/use-convex-file-upload.ts`: the chat total-size check (which `return`s early without uploading) now runs **before** the `tooManyFiles` slot-overflow toast, plus the new regression tests in `use-convex-file-upload.cap.test.ts`.

### Method
Two-round review (5 breadth subagents across correctness / robustness / elegance / tests / boundaries, then independent confirmation of every candidate finding).

### CI & tests
- **CI: all required checks green** (`gh pr checks` — Unit, UI, Type check, Lint, Format, Build *, all 16 Playwright platform shards, etc. pass; only fork-PR/Trivy/container jobs skipped, as expected).
- **Tests: ran `bunx vitest --run --project client app/features/chat/hooks/use-convex-file-upload.cap.test.ts` → 4 passed, 0 failed** (re-run in confirmation round, same result).

### Findings
- **Correctness:** Clean reorder. Total-size check sums over `acceptedFiles` (the slot-trimmed set), which is exactly what would upload — correct. The contradictory double-toast is eliminated: when the trimmed batch exceeds the cap, only `totalSizeExceeded` fires; when it fits, `tooManyFiles` still fires and the trimmed batch uploads. No success/failure inversion. Exactly one copy of the moved toast block (the `slotsAvailable <= 0` toast at L247 is a distinct, mutually-exclusive case).
- **Tests:** Both branches covered. The regression test's `tooManyFiles == false` assertion would have **failed** on the pre-fix ordering, so it genuinely pins the fix; the second test guards against over-correcting (suppressing the legitimate slot toast).
- **Elegance:** Minimal, focused; the explanatory comment matches this file's heavily-commented style and references #2029.
- **Boundaries/resolution:** Implements the issue's recommended approach verbatim. Strict `>` against `CHAT_MAX_TOTAL_SIZE` (200 MB) is correct inclusive-cap semantics. Client-side UX only — no security boundary crossed. No other call-sites needed changing (all chat composers consume this one hook).

### Non-blocking note (out of scope)
There is a **pre-existing** TOCTOU in `uploadFiles`: it reads `attachmentsRef.current` for slot/size/dedup before any `setAttachments`, so two rapid concurrent calls could each pass the caps and over-commit. Confirmed this exists on `origin/main` and is **not introduced or worsened** by this PR (the reorder adds no new state mutation or await). Worth a separate issue; does not block this merge.

**READY TO MERGE.**

PR #2188, opened 2026-06-27
fix(platform): use compressed image sizes in chat total-attachment-size check (#2031)

### LINKED ISSUE #2031: Bug: Chat total-attachment-size check uses pre-compression image sizes, causing false rejections for image batches
## Area

## Steps to reproduce

1. Open a chat composer.
2. Attach 8 images that together total 150 MB raw (but would compress to ~40 MB via \`compressImage\`).
3. Attempt to add 2 more images that are 60 MB raw combined.
4. Total size check: \`150 MB (existing, compressed) + 60 MB (incoming, raw) = 210 MB > 200 MB\` — **rejected**.
5. In reality the actual stored sizes would be ≈40 MB + ≈15 MB ≈ 55 MB — well within the 200 MB cap.

## Expected

The total-size check should compare against the compressed (actual uploaded) size of incoming images, or at minimum be consistent with the stored size values used by \`existingSize\`.

## Actual

- **\`incomingSize\`** at line 266 sums \`file.size\` — the **original** file size, before \`compressImage\` runs.
- **\`existingSize\`** at line 262 sums \`att.fileSize\` from \`attachmentsRef.current\`, which was populated from \`fileToUpload.size\` (post-compression) at line 329.
- The comparison mixes compressed existing sizes with raw incoming sizes, systematically over-counting incoming images and producing false rejections for image-heavy batches.

## Evidence

- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:262-268\` — \`existingSize\` uses stored (compressed) sizes; \`incomingSize\` uses raw sizes
- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:285-291\` — image compression runs inside the upload promise, AFTER the total-size check has already decided
- \`services/platform/app/features/chat/hooks/use-convex-file-upload.ts:329\` — compressed \`fileToUpload.size\` stored in attachment state
- \`services/platform/lib/shared/file-types.ts:524\` — \`CHAT_MAX_TOTAL_SIZE = 200 MB\`

## Notes

Fix: either move compression before the size check (so raw compressed sizes are used for all calculations), or remove pre-upload compression and handle it elsewhere, or accept the over-estimate as a conservative guard (but document the trade-off). The inconsistency between \`existingSize\` (compressed) and \`incomingSize\` (raw) is the core issue; the two should use the same basis.

larryro:
## Summary

The chat composer's total-attachment-size check mixed two different size bases: `existingSize` summed the **compressed** sizes of already-attached files (stored after `compressImage` runs), while `incomingSize` summed the **raw** `file.size` of incoming images — because compression ran *inside* the upload promise, *after* the check had already decided. This systematically over-counted image-heavy batches and produced false `totalSizeExceeded` rejections (e.g. 240 MB raw images that compress to a few MB).

## Fix

Move image compression to run **before** the total-size check, so both `existingSize` and `incomingSize` use the same post-compression basis (the sizes that are actually stored/uploaded). The upload promise now reuses the already-compressed file instead of compressing again. A compression failure falls back to the original file so one bad image doesn't sink the whole batch.

## Tests

Added regression coverage in `use-convex-file-upload.cap.test.ts`:
- accepts an image batch that exceeds the 200 MB cap raw but compresses under it
- still rejects when even the compressed total exceeds the cap

Verified with `vitest --project client` (4 passed), `oxlint --type-aware` (0 errors), and `tsc --noEmit` (clean).

Closes #2031

### DIFF
--- a/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
+++ b/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
@@ -4,7 +4,8 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { useUploadPolicy } from '@/app/features/settings/governance/hooks/queries';
 import { toast } from '@/app/hooks/use-toast';
-import { CHAT_MAX_FILE_COUNT } from '@/lib/shared/file-types';
+import { CHAT_MAX_FILE_COUNT, detectMediaMime } from '@/lib/shared/file-types';
+import { compressImage } from '@/lib/utils/compress-image';
 
 import { useConvexFileUpload } from './use-convex-file-upload';
 
@@ -62,12 +63,20 @@ vi.mock('@/lib/shared/file-types', async (importOriginal) => {
 
 const toastMock = vi.mocked(toast);
 const useUploadPolicyMock = vi.mocked(useUploadPolicy);
+const detectMediaMimeMock = vi.mocked(detectMediaMime);
+const compressImageMock = vi.mocked(compressImage);
 
 const MB = 1024 * 1024;
 
 // Override `size` so the test doesn't allocate the simulated byte count.
 function makeAudioFile(name: string, size: number): File {
   const file = new File([new Uint8Array(0)], name, { type: 'audio/mpeg' });
+  Object.defineProperty(file, 'size', { value: size });
+  return file;
+}
+
+function makeImageFile(name: string, size: number): File {
+  const file = new File([new Uint8Array(0)], name, { type: 'image/png' });
   Object.defineProperty(file, 'size', { value: size });
   return file;
 }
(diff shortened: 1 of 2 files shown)


larryro:
## Desk review — #2031 total-attachment-size uses pre-compression image sizes

**Verdict: READY TO MERGE.**

### What was verified
- **CI:** all checks green (`gh pr checks` — every job `pass`; the handful of `skipping` jobs are fork-only/conditional, expected).
- **Tests:** ran the project's own suite locally —
  `bunx vitest --run --project client app/features/chat/hooks/use-convex-file-upload.cap.test.ts` → **4 passed (4)**.
  (The `.cap.test.ts` file lives in the `client` jsdom project, not the default `bun test` server/pii projects.)

### Correctness — fix resolves the issue
The bug was the mixed basis in `useConvexFileUpload.uploadFiles`: `incomingSize` summed raw `file.size` while `existingSize` summed the already-compressed `att.fileSize`, over-counting image batches. The fix compresses images **before** the total-size check (`preparedFiles`), so both `existingSize` and `incomingSize` (`use-convex-file-upload.ts:304-311`) now use post-compression sizes — the same basis stored in attachment state (`fileSize: fileToUpload.size`). The downstream upload/metadata/preview paths all consistently use `fileToUpload`. Branch walk (image-success / image-failure-fallback / non-image passthrough / empty / mixed batch) is sound.

The new accept-test is a **true regression test**: 3×80MB raw = 240MB would be rejected by the old `incomingSize` (raw) path but is accepted now that it compresses to 3MB. It fails on `main`, passes on this branch.

### Non-blocking observations (do NOT block merge; reasonable follow-ups)
1. **Untested error branch** — the new `try/catch` compression fallback (`use-convex-file-upload.ts:291-296`) is never exercised by a test (no `mockRejectedValue`). It's a simple defensive fallback that still counts raw size against the per-file and total caps, so no limit is bypassed — but a `compressImage` rejection test would close the coverage gap.
2. **Behavior change on compression failure** — previously a compression throw was caught by the per-file upload `catch` (→ `uploadFailed` toast, file skipped); now it falls back to uploading the **raw** file. This is deliberate and documented ("a single compression failure doesn't sink the whole batch") and arguably an improvement; flagging only so it's a conscious choice.
3. **Out of scope (pre-existing, not introduced here):** the per-file cap (`:143`) and the dedup key (`:228`) still use raw `file.size`; there is no server-side `CHAT_MAX_TOTAL_SIZE` enforcement (client is the sole gate). None of these are part of #2031.

The change is minimal, focused, follows repo conventions (`console.warn`, no empty catch), and leaves no dead code.

**READY TO MERGE.**

Issue #2024, opened 2026-06-23
Bug: Automations — /automations/metrics?period=90 crashes (same SearchParamError as #1987 on agents)


yannickmonney:
## Area

Automations → Metrics tab — `services/platform/app/routes/dashboard/\$id/automations/metrics.tsx`

## Steps to reproduce

1. Navigate directly to `/dashboard/{orgId}/automations/metrics?period=90` (or `?period=30`, `?period=7`) — e.g. share/bookmark the URL while viewing the 90-day window.
2. The page renders the full-page error boundary instead of the metrics.

Selecting the period via the in-app combobox works (it navigates with a string), but opening a shared/bookmarked URL fails because TanStack Router parses `90` as a JSON number while the schema expects the string `'90'`.

## Expected

The page loads the automations metrics for the requested period; an out-of-range value falls back to the default (30 days) rather than crashing.

## Actual

`SearchParamError` — `invalid_value, values: ['7', '30', '90'], path: [period]` — causes the error boundary to render.

## Evidence

`services/platform/app/routes/dashboard/$id/automations/metrics.tsx:14-16`
```ts
const searchSchema = z.object({
  period: z.enum(['7', '30', '90']).optional(),
});
```

The schema uses a string enum without `.catch()` or `.coerce`. When a URL contains `?period=90`, the router parses the value as the number `90`, which fails the string enum — identical root cause to #1987 on the agents metrics page.

## Notes

- This is a separate instance of the same bug class as #1987 (agents metrics). Both routes need the same fix: coerce + catch, e.g. `z.coerce.string().pipe(z.enum(['7', '30', '90'])).catch('30').optional()`.
- `services/platform/app/routes/dashboard/$id/agents/metrics.tsx` (filed as #1987) and this file share the identical pattern.

Issue #1987, opened 2026-06-23
Bug: A shareable /agents/metrics?period=90 URL crashes the page (error boundary) — search-param enum is string, router parses number


yannickmonney:
### What happened?
Opening the workforce metrics page with an explicit period query param — e.g. a bookmarked/shared `/dashboard/{org}/agents/metrics?period=90` — renders the full-page "Something went wrong" error boundary. Selecting the period from the in-app combobox works (it navigates with a string), but a typed/shared URL crashes because TanStack Router parses `90` as a JSON number while the schema expects the string `'90'`.

### Steps to reproduce
1. Open `/dashboard/{org}/agents/metrics?period=90` directly (or `?period=30`).
2. Page renders the error boundary instead of the metrics.

### Expected behavior
The page loads the metrics for the requested period; an out-of-range value falls back to a default rather than crashing.

### Affected service
Platform (web app)

### Version or commit
`cd13268b8`

### Browser and OS
Headless Chrome (Playwright) on macOS

### Relevant log output / evidence
**Verified live**: main content = "Agents … Something went wrong / An error occurred while loading this page." Console: `SearchParamError … invalid_value, values:['7','30','90'], path:[period]`.

**Root cause.** `services/platform/app/routes/dashboard/$id/agents/metrics.tsx:6-7` — `searchSchema = z.object({ period: z.enum(['7','30','90']).optional() })` with `validateSearch: searchSchema`. The router supplies the parsed number `90`, which fails the string enum.

**Fix.** Coerce + fallback, e.g. `period: z.coerce.number().pipe(z.union([z.literal(7),z.literal(30),z.literal(90)])).catch(30).optional()` (or accept string|number and `.catch`).

---
*Filed by an automated deep manual-QA sweep (wave 2) on a local hermetic stack (commit `cd13268b8`, mode-A mock LLM, headless Chrome). Reproduced before filing.*

PR #2190, opened 2026-06-27
fix(platform): coerce project metrics period search param (#2033)

### LINKED ISSUE #2033: Bug: /projects/$projectId/metrics?period=90 crashes — same JSON-parsed-number SearchParamError as #1987 and #2024
## Area

Platform — project metrics route at `/dashboard/$id/projects/$projectId/metrics`.

## Steps to reproduce

1. Open any project's metrics page normally (e.g. navigate from the Tasks view → Metrics link).
2. Once there the URL has `?period=%2230%22` (JSON-encoded string `"30"`).
3. Now copy the URL, replace `%2230%22` with `30` (bare integer) and open it — or navigate directly to `/dashboard/{org}/projects/{projectId}/metrics?period=90`.
4. The page hits the error boundary.

## Expected

The page renders with the 90-day period selected (or falls back to the default if 90 is not in the allowed set).

## Actual

`validateSearch` receives `period: 90` (number — TanStack Router's default `parseSearch` calls `JSON.parse("90")` → `90`). `z.enum(['7','30','90'])` expects strings and throws a `SearchParamError`, which bubbles to the error boundary.

## Evidence

`services/platform/app/routes/dashboard/$id/projects/$projectId/metrics.tsx:22`:
```ts
const searchSchema = z.object({
  period: z.enum(['7', '30', '90']).optional(),
});
```

Root cause confirmed in `node_modules/@tanstack/router-core/dist/cjs/searchParams.cjs`:
```js
var defaultParseSearch = parseSearchWith(JSON.parse);
// "90" → JSON.parse("90") = 90 (number)
```

## Notes

Identical root cause to #1987 (agents metrics) and #2024 (automations metrics). Fix pattern: replace `z.enum(['7','30','90'])` with `z.string().optional()` validated to `PeriodDays` in the component, or use `z.coerce.string()` before the enum check.

larryro:
## Summary

A shared or bookmarked `/dashboard/{org}/projects/{projectId}/metrics?period=90` URL was parsed by TanStack Router's default `parseSearch` (which runs `JSON.parse`) as the JSON **number** `90`. The route's `z.enum(['7','30','90'])` expects strings, so it threw a `SearchParamError` and the page fell to the error boundary.

This applies the same fix already shipped for the agents metrics page (#1987) and automations metrics page (#2024): coerce the value to a string, validate it against the enum, and fall back to the default 30-day window for any out-of-range value so a shared/bookmarked URL never crashes.

## Changes
- `metrics.tsx`: `period` now uses `z.coerce.string().pipe(z.enum(['7','30','90'])).catch('30').optional()`. The schema is exported for testing.
- `metrics.search.test.ts`: regression coverage mirroring #2024 — numeric coercion, string pass-through, out-of-range fallback, omitted-undefined.

## Verification
- `vitest --project server` on the new test: 4 passed
- `tsc --noEmit`: clean

Closes #2033

### DIFF
--- a/services/platform/app/routes/dashboard/$id/projects/$projectId/metrics.search.test.ts
+++ b/services/platform/app/routes/dashboard/$id/projects/$projectId/metrics.search.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from 'vitest';
+
+import { searchSchema } from './metrics';
+
+// Regression coverage for #2033: a shared/bookmarked
+// `/projects/$projectId/metrics?period=90` URL is parsed by the router as the
+// JSON number 90, which must not crash the route via SearchParamError.
+describe('project metrics searchSchema', () => {
+  it('coerces numeric period values to the string enum', () => {
+    expect(searchSchema.parse({ period: 90 }).period).toBe('90');
+    expect(searchSchema.parse({ period: 30 }).period).toBe('30');
+    expect(searchSchema.parse({ period: 7 }).period).toBe('7');
+  });
+
+  it('accepts string period values', () => {
+    expect(searchSchema.parse({ period: '90' }).period).toBe('90');
+  });
+
+  it('falls back to the default window for out-of-range values', () => {
+    expect(searchSchema.parse({ period: 999 }).period).toBe('30');
+    expect(searchSchema.parse({ period: 'nonsense' }).period).toBe('30');
+  });
+
+  it('leaves an omitted period undefined', () => {
+    expect(searchSchema.parse({}).period).toBeUndefined();
+  });
+});
--- a/services/platform/app/routes/dashboard/$id/projects/$projectId/metrics.tsx
+++ b/services/platform/app/routes/dashboard/$id/projects/$projectId/metrics.tsx
@@ -9,8 +9,16 @@ import {
 } from '@/app/features/tasks/components/project-metrics-page';
 import { seo } from '@/lib/utils/seo';
 
-const searchSchema = z.object({
-  period: z.enum(['7', '30', '90']).optional(),
+export const searchSchema = z.object({
+  // The router parses a bare `?period=90` as the JSON number 90, which fails a
+  // plain string enum and crashes the page (issue #2033). Coerce to a string
+  // first, then fall back to the default window for any out-of-range value so a
+  // shared/bookmarked URL never renders the error boundary.
+  period: z.coerce
+    .string()
+    .pipe(z.enum(['7', '30', '90']))
+    .catch('30')
+    .optional(),
 });
 
 export const Route = createFileRoute(


larryro:
## Desk Review — PR #2190 (fixes #2033)

**Verdict: READY TO MERGE.**

Reviewed the whole implementation (not just the diff) over two rounds: 5 breadth subagents across the five review dimensions, then independent re-verification of every load-bearing claim.

### CI
All required checks **green** — Type check ✅, Lint ✅, Format ✅, Knip ✅, Unit ✅, Build platform ✅, all 16 Playwright (platform) shards ✅, Opengrep ✅, Smoke test ✅. The only non-`pass` jobs are conditional skips (Trivy, fork-PR smoke/validate, docs/web container tests) — none are failures or pending.

### Tests (run locally, single time)
- `bunx vitest --run --project server metrics.search.test.ts` → **4 passed**.
- Additional edge sweep I ran (null, boolean, object, array, empty string, negative, float, explicit-undefined) → all 11 pass; every JSON-parseable input either maps to a valid enum value or falls back to `'30'`. **No input can throw `SearchParamError` anymore.**

### Findings (confirmed)
- **Correctness ✅** — `z.coerce.string().pipe(z.enum(['7','30','90'])).catch('30').optional()` resolves bare numeric `?period=7|30|90` to the matching string; out-of-range/invalid → `'30'`; omitted → `undefined`. The consumer ternary (`metrics.tsx:41`) maps all four possible outputs (`'7'|'30'|'90'|undefined`) exhaustively to `PeriodDays`. `handleChangePeriod` writes back only valid enum strings. Fixes the reported root cause, not a narrower symptom.
- **Robustness ✅** — `.optional()` short-circuits `undefined` before coercion (verified: `{}` → `undefined`, not `'30'`); `.catch('30')` swallows every enum failure so the route never falls to the error boundary. Search-param values come from `JSON.parse`, so no Symbol/BigInt can reach `z.coerce.string()` to make it throw. `useCallback` deps are complete; `replace: true` is appropriate.
- **Boundaries ✅** — value is doubly constrained: frontend enum (7/30/90) plus backend clamp `windowKeys(args.days ?? 30)` → `Math.min(Math.max(days,1), 90)` at `convex/task_metrics/queries.ts:48`. Server-side access control (`getProjectTaskMetrics`) is unchanged. `period` is the only search param on this route.
- **Tests ✅** — regression test directly covers the bug (numeric `90` → `'90'`), plus out-of-range, string pass-through, and omitted cases; file matches the `server` vitest project. Minor untested inputs (null/boolean/float) are behaviorally safe (empirically verified above) — not blocking.
- **Elegance ✅** — minimal, focused 3-line schema change with an accurate explanatory comment; exporting `searchSchema` is solely to enable the test.

### Non-blocking note
The PR description states this fix was "already shipped for the agents metrics page (#1987) and automations metrics page (#2024)." That precedent does **not** exist in the codebase: both `dashboard/$id/agents/metrics.tsx` and `dashboard/$id/automations/metrics.tsx` still use the vulnerable `z.enum(['7','30','90']).optional()`, and this PR's file is the only exported `searchSchema`. This does not affect #2033's fix (which is self-contained and correct), but those two routes carry the identical latent bug and warrant their own follow-up under #1987/#2024.

**READY TO MERGE.**

PR #2192, opened 2026-06-27
fix: guard owner role and duplicate membership in createMember/addMember (#2035)

### LINKED ISSUE #2035: Bug: Public createMember / addMember accept role:'owner' with no owner guard — an admin can self-escalate to owner
## Area / route
Auth / RBAC — member-role authorization. Backend public Convex mutations:
- `api.users.mutations.createMember`
- `api.members.mutations.addMember`

## Steps to reproduce
As a non-owner org **admin** (role `'admin'`), call either public mutation directly (bypassing the UI), e.g. via the Convex client or HTTP:

1. `api.users.mutations.createMember({ organizationId, email, password, role: 'owner' })` — creates a brand-new account whose membership role is `owner`.
2. `api.members.mutations.addMember({ organizationId, userId: <own or any userId>, role: 'owner' })` — `addMember` does not check for an existing membership, so an admin can re-add their own `userId` with `role: 'owner'`, minting a second owner membership row for themselves.

Both calls pass arg validation because the role validators include the `'owner'` literal, and the handlers only gate on `isAdmin(callerRole)` — which is true for `admin` as well as `owner`.

## Expected
Creating or adding a member with `role: 'owner'` must be rejected, exactly as `updateMemberRole` does (`members/mutations.ts:284-286`: "The owner role cannot be assigned manually"). Owner should be reachable only through the owner-gated `transferOwnership` flow (`members/mutations.ts:432`, caller must already be `owner`). The role argument from a non-owner caller should never be honored as `owner`.

## Actual
Neither create path has an owner guard:
- `users/create_member.ts:127` and `:217` compute `role = (args.role ?? 'member').toLowerCase()` and write it straight through to the Better Auth `member` row.
- `members/mutations.ts:80` (`addMember`) does the same: `const role = (args.role ?? 'member').toLowerCase();` written through, with no existing-membership check.

The forged `owner` membership row is privilege-bearing downstream: `two_factor/mutations.ts:87,178`, `governance/dsar_policy.ts:210` (`callerIsOwner`), `auth.ts:263`, and `lib/rls/helpers/access_control.ts:242` all grant elevated capability based on the `owner` role. `upsertMemberMirror` (`members/mirror_sync.ts:41-74`) keys on `memberId`, so a second membership produces a second mirror row that the RLS read cache will surface. Net effect: a non-owner admin escalates to owner, bypassing the owner-only `transferOwnership` invariant.

The `AddMemberDialog` Zod schema (`member-add-dialog.tsx:75`) restricts the role select to `['disabled','admin','developer','editor','member']`, hiding `owner` — but that is a UI-only restriction. The public Convex mutation is the real authorization boundary and it is unguarded.

## Evidence (file:line)
- `services/platform/convex/users/validators.ts:9-16` — `roleValidator` includes `v.literal('owner')`
- `services/platform/convex/members/validators.ts:13-20` — `memberRoleValidator` includes `v.literal('owner')`
- `services/platform/convex/users/mutations.ts:58-77` — `createMember` is a public `mutation` with `role: v.optional(roleValidator)`
- `services/platform/convex/users/create_member.ts:71-73` — gate is only `isAdmin(callerRole)`
- `services/platform/convex/users/create_member.ts:127` and `:217` — `role = (args.role ?? 'member').toLowerCase()` written through, no owner check
- `services/platform/convex/members/mutations.ts:45,80` — `addMember` accepts `memberRoleValidator`, writes role through, no owner check, no existing-membership check
- `services/platform/convex/lib/rls/helpers/role_helpers.ts:8-11` — `isAdmin()` returns true for both `admin` and `owner`
- Contrast (the guard that should also apply): `services/platform/convex/members/mutations.ts:280-286` (`updateMemberRole` rejects `owner`) and `:432` (`transferOwnership` is owner-only)
- Downstream owner power: `services/platform/convex/two_factor/mutations.ts:87,178`; `services/platform/convex/governance/dsar_policy.ts:210`; `services/platform/convex/auth.ts:263`; `services/platform/convex/lib/rls/helpers/access_control.ts:242`

## Notes
- Fix: in both `createMember` (both branches at `create_member.ts:127` and `:217`) and `addMember` (`members/mutations.ts:80`), reject `role.toLowerCase() === 'owner'` before the adapter create, mirroring `updateMemberRole`'s existing rejection. Consider dropping `'owner'` from `roleValidator`/`memberRoleValidator` for the create/add args (or introduce a create-specific validator without the `owner` literal) so the boundary is enforced at arg-validation time too.
- Secondary hardening: `addMember` should also reject adding a `userId` that already has a membership in the org (the `createMember` existing-user branch already does this at `create_member.ts:122`), to prevent duplicate/second membership rows regardless of role.
- The `internal_mutations.ts` / `add_member_internal.ts` paths were not part of this report; this finding concerns only the public `createMember` and `addMember` mutations.

---

larryro:
## Summary

Closes #2035.

The public Convex mutations `api.users.mutations.createMember` and `api.members.mutations.addMember` accepted `role: 'owner'` and gated only on `isAdmin(callerRole)` — which is true for `admin` as well as `owner`. This let a non-owner admin self-escalate to `owner`, bypassing the owner-only `transferOwnership` invariant. `addMember` additionally never checked for an existing membership, so an admin could mint a second (owner) membership row for any `userId`, including their own.

## Changes

- **`users/create_member.ts`** — reject `role.toLowerCase() === 'owner'` right after the admin check, before any account/membership is written. Mirrors `updateMemberRole`'s existing guard ("The owner role cannot be assigned manually").
- **`members/mutations.ts` (`addMember`)** — reject `role === 'owner'` with the same message, and reject adding a `userId` that already has a membership in the org ("User is already a member of this organization"), mirroring `createMember`'s existing-user branch. This prevents duplicate/second membership rows regardless of role.
- Owner remains reachable only through the owner-gated `transferOwnership` flow.

## Tests

- `members/mutations.test.ts` — new cases: `addMember` rejects the owner role, and rejects adding an already-member user; updated the happy-path test for the new existing-membership lookup.
- `users/create_member.test.ts` (new) — `createMember` rejects the owner role for an admin caller and writes nothing; plus unauthenticated / non-admin guards.

## Verification

- `vitest --run --project server` (members + users mutation tests) — green (26 tests)
- `tsc --noEmit` — clean
- `oxlint --type-aware` on changed files — 0 warnings / 0 errors

### DIFF
--- a/services/platform/convex/members/mutations.test.ts
+++ b/services/platform/convex/members/mutations.test.ts
@@ -555,6 +555,42 @@ describe('addMember handler', () => {
     );
   });
 
+  it('rejects assigning the owner role', async () => {
+    mockGetAuthUser.mockResolvedValue(AUTH_USER);
+    const ctx = createMockCtx();
+    // caller is admin
+    ctx.runQuery.mockResolvedValueOnce({
+      page: [{ _id: 'm_caller', role: 'admin' }],
+    });
+    const handler = await getHandler();
+
+    await expect(
+      handler(ctx, { ...defaultArgs, role: 'owner' }),
+    ).rejects.toThrow('The owner role cannot be assigned manually');
+  });
+
+  it('rejects adding a user who is already a member', async () => {
+    mockGetAuthUser.mockResolvedValue(AUTH_USER);
+    const ctx = createMockCtx();
+    // caller is admin
+    ctx.runQuery.mockResolvedValueOnce({
+      page: [{ _id: 'm_caller', role: 'admin' }],
+    });
+    // target user lookup
+    ctx.runQuery.mockResolvedValueOnce({
+      page: [{ _id: 'user_target', email: 'target@example.com' }],
+    });
+    // existing-membership lookup returns a row
+    ctx.runQuery.mockResolvedValueOnce({
+      page: [{ _id: 'm_existing', role: 'member' }],
+    });
+    const handler = await getHandler();
+
+    await expect(handler(ctx, defaultArgs)).rejects.toThrow(
+      'User is already a member of this organization',
+    );
+  });
+
   it('allows owner to add members', async () => {
     mockGetAuthUser.mockResolvedValue(AUTH_USER);
     const ctx = createMockCtx();
@@ -566,6 +602,8 @@ describe('addMember handler', () => {
     ctx.runQuery.mockResolvedValueOnce({
       page: [{ _id: 'user_target', email: 'target@example.com' }],
     });
+    // existing-membership lookup — none, so the add proceeds
+    ctx.runQuery.mockResolvedValueOnce({ page: [] });
     // create
     ctx.runMutation.mockResolvedValueOnce({ _id: 'm_new' });
     const handler = await getHandler();
--- a/services/platform/convex/members/mutations.ts
+++ b/services/platform/convex/members/mutations.ts
@@ -69,6 +69,14 @@ export const addMember = mutation({
       throw new Error('Only admins can add members');
     }
 
+    // The owner role can only be assigned through the owner-gated
+    // transferOwnership flow — reject it here so a non-owner admin cannot
+    // self-escalate by passing role: 'owner' (mirrors updateMemberRole).
+    const role = (args.role ?? 'member').toLowerCase();
+    if (role === 'owner') {
+      throw new Error('The owner role cannot be assigned manually');
+    }
+
     const targetUser = findOneUser(
       await ctx.runQuery(components.betterAuth.adapter.findMany, {
         model: 'user',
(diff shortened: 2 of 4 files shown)


larryro:
## Desk Review — #2035 owner-role self-escalation guard

**Verdict: READY TO MERGE.**

Reviewed branch `tale/xs7921x194c3a41eb73bdns62n89e5t2` (PR #2192) by a two-round method (5 breadth subagents across correctness / robustness / elegance / tests / boundaries, then independent re-verification of every blocking candidate).

### CI & tests
- `gh pr checks` — **all checks green** (Analyze, Build ×10, Lint, Format, Knip, Opengrep, Migrations, Playwright platform 1–16/web/docs, Scan, etc.). No red, no pending.
- Ran the project's own suite: `bunx vitest --run --project server` → **542 files, 6495 tests passed**. The two changed test files alone → 26 passed. No regressions.

### What's correct
- **`users/create_member.ts`** — owner guard placed once after the `isAdmin` check (`if ((args.role ?? '').toLowerCase() === 'owner') throw`), *before* both the existing-user and new-user write branches and before `auth.api.signUpEmail`. Verified no member-write path can reach `role:'owner'`. Test asserts the guard fires with zero `runMutation` calls.
- **`members/mutations.ts` (`addMember`)** — same owner guard, plus a new existing-membership check (`organizationId`+`userId`) mirroring `createMember`, preventing duplicate/second membership rows. Guards fire before any side effect.
- Both messages match the existing `updateMemberRole` guard verbatim ("The owner role cannot be assigned manually"). Owner remains reachable only through the owner-gated `transferOwnership`.
- Case bypass not possible: validators only accept the exact lowercase literal `'owner'`; `.toLowerCase()` is defensive.
- **No sibling bypass.** Swept `convex/` for other public mutations writing member rows with an attacker-controlled role: none. `team_members.mutations.addMember` writes `teamMember` rows with no role field and requires existing org membership; `reset_owner` is an admin-CLI internalMutation. The two public org-member-writing paths named in the issue are now guarded.
- Robustness: Convex's serializable OCC makes the read-then-create existence check race-free (same pattern used elsewhere in the repo). Mirror-sync lifecycle unchanged; no leak introduced.

### Notes (non-blocking)
- The issue's optional "Notes" suggestion to also drop `'owner'` from `roleValidator`/`memberRoleValidator` was not taken; runtime-only guarding is sufficient and correct here, since those validators are shared with `updateMemberRole`/`transferOwnership`/read paths that legitimately need `'owner'`.
- `createMember`'s pre-existing existing-user / new-user branches remain without dedicated unit tests, but that is pre-existing code untouched by this PR (the diff only adds the guard); the new behavior introduced here is fully tested. Not a blocker.

Clean, minimal, well-targeted fix that genuinely closes the reported escalation. Approving.

PR #2195, opened 2026-06-27
fix(platform): guard trusted-headers authenticate endpoint against open redirect (#2037)

### LINKED ISSUE #2037: Bug: Open redirect on the trusted-headers authenticate endpoint (?redirect= reflected into navigation with no origin validation)
**Area / route**
Auth (trusted-headers / reverse-proxy SSO). Endpoint: `GET /api/trusted-headers/authenticate` → `convex/trusted_headers_auth/authenticate_handler.ts`. Only active on deployments with `TRUSTED_HEADERS_ENABLED=true`.

**Steps to reproduce**
1. Deploy with `TRUSTED_HEADERS_ENABLED=true` (behind an authenticating reverse proxy: Authelia/Authentik/oauth2-proxy).
2. Deliver to a victim a link to either:
   - `/log-in?redirectTo=https://evil.com` (the login page URL-encodes `redirectTo` and forwards it as `?redirect=` to the authenticate endpoint — see log-in.tsx:81-82, fired automatically by the useEffect at lines 84-92), or
   - the endpoint directly: `/api/trusted-headers/authenticate?redirect=https://evil.com`.
3. The proxy has already authenticated the victim, so the endpoint mints a session and returns the success HTML.
4. The victim's browser is redirected to `https://evil.com`.

`//evil.com` (protocol-relative) works identically.

**Expected**
`redirect` should be validated as a same-origin / relative-path target (e.g. require it to begin with a single `/` and not `//`, or resolve against the deployment origin and reject any other origin) before being used as a navigation target. External absolute URLs should be rejected and replaced with the default `${basePath}/dashboard`.

**Actual**
The redirect target is taken verbatim from the query string and reflected into both navigation sinks with only XSS-escaping applied:
- `redirectTo = url.searchParams.get('redirect') || `${basePath}/dashboard`` (no validation).
- Reflected into `<meta http-equiv="refresh" content="0;url=${escapeHtmlAttr(redirectTo)}">` and `window.location.href = '${escapeJs(redirectTo)}'`.
- `escapeHtmlAttr`/`escapeJs` only escape HTML-attribute / JS-string metacharacters (`&"'<>`, backslash, control chars). They do NOT check the scheme/origin, so `https://evil.com` and `//evil.com` pass through unchanged and become the navigation target.

Because the trusted-headers login flow auto-fires on page load, an attacker can use a trusted, legitimately-authenticating domain as an open redirector for phishing (victim sees a real login on the trusted domain, then lands on the attacker site).

- `services/platform/convex/trusted_headers_auth/authenticate_handler.ts:70-71` — `redirect` taken verbatim from the query string.
- `:190` — reflected into `<meta http-equiv="refresh" ...>`.
- `:195` — reflected into `window.location.href`.
- `:230-237` (`escapeHtmlAttr`) and `:239-250` (`escapeJs`) — XSS escaping only, no origin/scheme/path validation.
- `services/platform/app/routes/_auth/log-in.tsx:81-82` — login page URL-encodes `redirectTo` and forwards it as `?redirect=`; `:84-92` auto-fires the redirect on a trusted-headers deployment.
- `services/platform/convex/http.ts:931-935` and `convex/trusted_headers_auth/http_handlers.ts:1-7` — route + wrapper add no validation.

**Notes**
- No redirect-validation helper exists anywhere in the codebase (repo-wide grep for `isSafeRedirect`/`sameOrigin`/`allowlist`/`validateRedirect` found nothing applicable).
- The analogous SSO post-authentication handler `convex/enterprise_sso/login/set_session_handler.ts:63,68` hard-codes the target to `${basePath}/dashboard` and never reflects a user-supplied param — the trusted-headers handler is the inconsistent one. A minimal fix is to validate `redirect` to a relative same-origin path (reject anything not matching `^/(?!/)`), falling back to `${basePath}/dashboard`. The login-page forward should apply the same allowlist defensively.
- Priority medium because exposure is limited to deployments running with `TRUSTED_HEADERS_ENABLED=true`.

---

larryro:
## Summary

Closes #2037 — open redirect on the trusted-headers authenticate endpoint.

The runtime fix for this issue already landed on `main` as part of #2118: a shared `sanitizeInternalRedirect` / `isSafeInternalPath` helper (`lib/shared/utils/safe-redirect.ts`) is now applied at **both** navigation sinks that the issue called out:

- `convex/trusted_headers_auth/authenticate_handler.ts` — the `redirect` query param is passed through `sanitizeInternalRedirect(..., `${basePath}/dashboard`)` before being reflected into the `<meta http-equiv="refresh">` and `window.location.href` sinks.
- `app/routes/_auth/log-in.tsx` — the login page forwards only a validated same-origin path as `?redirect=` (defence in depth).

`isSafeInternalPath` requires a single leading `/`, rejects protocol-relative (`//evil.com`) and backslash (`/\evil.com`) host-smuggling, rejects absolute/`scheme:` URLs, and re-resolves against a fixed origin as a belt-and-braces check. Anything unsafe falls back to `${basePath}/dashboard`.

What was still missing was a regression test pinned to the **actual vulnerable endpoint** (the existing tests only cover the helper in isolation). This PR adds that:

- `convex/trusted_headers_auth/authenticate_handler.test.ts` drives `trustedHeadersAuthenticateHandler` end-to-end with hostile `redirect` values (`https://evil.com`, `http://evil.com/phish`, `//evil.com`, `/\evil.com`, `javascript:alert(1)`, bare `dashboard`) and asserts the malicious host/scheme never reaches either navigation sink and that both fall back to `/dashboard`. It also asserts a safe same-origin path (`/log-in/team-a?tab=members`) is preserved and that a missing param falls back to the dashboard.

This formally links and closes the issue, and locks in the fix so a future refactor of the handler can't silently reintroduce the open redirect.

## Tests

- `bunx vitest --run --project server convex/trusted_headers_auth/authenticate_handler.test.ts` → 8 passing.
- `bunx vitest --run --project server lib/shared/utils/safe-redirect.test.ts` → 5 passing.
- `bunx oxlint --type-aware` clean on the new file.

### DIFF
--- a/services/platform/convex/trusted_headers_auth/authenticate_handler.test.ts
+++ b/services/platform/convex/trusted_headers_auth/authenticate_handler.test.ts
@@ -1,0 +1,107 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { ActionCtx } from '../_generated/server';
+import { trustedHeadersAuthenticateHandler } from './authenticate_handler';
+
+// The handler signs a cookie and asks Better Auth for a JWT cookie; neither is
+// relevant to the redirect-target hardening under test, so stub both.
+vi.mock('../enterprise_sso/sign_cookie_value', () => ({
+  signCookieValue: vi.fn().mockResolvedValue('signed-token'),
+}));
+
+vi.mock('../auth', () => ({
+  createAuth: vi.fn(() => ({
+    handler: vi.fn().mockResolvedValue(new Response(null, { status: 200 })),
+  })),
+}));
+
+function createCtx(): ActionCtx {
+  return {
+    runMutation: vi.fn().mockResolvedValue({
+      sessionToken: 'session-token',
+      userId: 'user_1',
+      organizationId: 'org_1',
+      shouldClearOldSession: false,
+      trustedHeadersChanged: false,
+    }),
+  } as unknown as ActionCtx;
+}
+
+function authRequest(redirect: string | null): Request {
+  const url = new URL(
+    'https://app.example.com/api/trusted-headers/authenticate',
+  );
+  if (redirect !== null) {
+    url.searchParams.set('redirect', redirect);
+  }
+  return new Request(url.toString(), {
+    headers: { 'Remote-Email': 'user@example.com' },
+  });
+}
+
+describe('trustedHeadersAuthenticateHandler — open-redirect hardening (#2037)', () => {
+  beforeEach(() => {
+    process.env.TRUSTED_HEADERS_ENABLED = 'true';
+    process.env.BETTER_AUTH_SECRET = 'test-secret';
+    delete process.env.BASE_PATH;
+  });
+
+  afterEach(() => {
+    delete process.env.TRUSTED_HEADERS_ENABLED;
+    delete process.env.BETTER_AUTH_SECRET;
+    vi.clearAllMocks();
+  });
+
+  // Each entry is a user-controlled `redirect` value that must NOT survive into
+  // the navigation sinks; the handler must fall back to the dashboard.
+  const hostileRedirects = [
+    'https://evil.com',
+    'http://evil.com/phish',
+    '//evil.com',
+    '/\\evil.com',
+    'javascript:alert(1)',
+    'dashboard',
+  ];
+
+  it.each(hostileRedirects)(
+    'neutralises hostile redirect %s and falls back to the dashboard',
+    async (redirect) => {
+      const res = await trustedHeadersAuthenticateHandler(
+        createCtx(),
+        authRequest(redirect),
+      );
+      const html = await res.text();
+
+      expect(res.status).toBe(200);
+      // The malicious host/scheme must never reach either navigation sink.
+      expect(html).not.toContain('evil.com');
+      expect(html).not.toContain('javascript:');
+      // Both sinks point at the safe default instead.
+      expect(html).toContain('0;url=/dashboard');
+      expect(html).toContain("window.location.href = '/dashboard'");
+    },
+  );
+
+  it('preserves a safe same-origin redirect path', async () => {
+    const res = await trustedHeadersAuthenticateHandler(
+      createCtx(),
+      authRequest('/log-in/team-a?tab=members'),
+    );
+    const html = await res.text();
+
+    expect(html).toContain('0;url=/log-in/team-a?tab=members');
+    expect(html).toContain(
+      "window.location.href = '/log-in/team-a?tab=members'",
+    );
+  });
+
+  it('falls back to the dashboard when no redirect is supplied', async () => {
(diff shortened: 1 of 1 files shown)


larryro:
## Desk review — #2195 (resolves #2037, open redirect on trusted-headers authenticate)

**Verdict: READY TO MERGE.**

### What this PR contains
Exactly one file: `services/platform/convex/trusted_headers_auth/authenticate_handler.test.ts` (+107). The production guard (`sanitizeInternalRedirect`) already landed on `main` via #2118 and is applied at both navigation sinks in `authenticate_handler.ts` (meta-refresh ~L195 and `window.location.href` ~L200, both fed from the sanitized `redirectTo` at L73-76) **and** defensively in the login-page forward (`app/routes/_auth/log-in.tsx` L85-88). So this PR's contribution is the **regression test** that locks the fix in. After merge, `main` carries both the working guard and its test, and the endpoint is genuinely protected.

### CI
All green: **46 pass / 5 skipping / 0 fail / 0 pending**; PR `MERGEABLE` / `CLEAN` (2 commits behind `main`, no conflict). Skips are fork-PR / container jobs not applicable here.

### Tests I ran (bun + vitest, `server` project)
- `authenticate_handler.test.ts` — **8 passed**
- `lib/shared/utils/safe-redirect.test.ts` — **5 passed**
- login-page feature tests (`client` project) — **8 passed**

### Verification of the underlying fix (adversarial, two rounds)
I exercised the real `sanitizeInternalRedirect` against the full vector set. Genuinely hostile inputs are rejected to the `/dashboard` fallback: `https://evil.com`, `http://evil.com/phish`, `//evil.com`, `/\evil.com`, `javascript:…`, bare `dashboard`, leading-whitespace/control-char `//` smuggling, `////evil.com`, `/\t/evil.com`. Safe same-origin paths (incl. query/fragment, `/@evil.com`, `/.evil.com`) are preserved. I confirmed `//evil.com`, `/\evil.com`, `https://evil.com` flip origin to `evil.com` (correctly rejected), while `/%2f%2fevil.com` / `/%5Cevil.com` resolve to the **same** origin (harmless paths, correctly preserved — browsers don't decode `%2f`/`%5c` during origin resolution; not an open redirect). The test is a genuine regression test: against the pre-fix verbatim handler the hostile cases reflect `evil.com`/`javascript:` into the HTML, which fails the `not.toContain(...)` + fallback assertions.

### Non-blocking notes
- Test assertions are substring checks against the HTML sinks; they are tight enough to fail without the guard, but a structural/regex check would be marginally more robust. Not required.
- No coverage for `BASE_PATH` set or the error paths, but those are outside the open-redirect scope of #2037.
- The commit message reads "fix(...)" though the diff is test-only (the fix pre-landed in #2118). Cosmetic; the issue is fully resolved in the post-merge state.

**READY TO MERGE.**

PR #2196, opened 2026-06-27
fix(platform): clear conversations search filter when box emptied after ?search= seed (#1992)

### LINKED ISSUE #1992: Bug: Conversations — clearing the search box does nothing when the lane was opened via a ?search= URL param (filter stuck)
### What happened?
Loading a conversations lane with a `?search=` param seeds and applies the filter correctly, but the user can no longer clear it: emptying the search box leaves the list filtered, because the filter falls back to the URL param on every render.

### Steps to reproduce
1. Open `/dashboard/{org}/conversations/open?search=Refund` (filters to matching rows).
2. Clear the search box to empty.
3. The list stays filtered; hidden conversations never reappear.

### Expected behavior
Clearing the search box removes the filter and shows all conversations in the lane.

### Affected service
Platform (web app)

### Version or commit
`cd13268b8`

### Browser and OS

**Root cause.** `services/platform/app/features/conversations/components/conversations.tsx:91,113` — `searchQuery` is seeded from `initialSearch`, but the applied term is `const searchTerm = searchQuery || initialSearch;`. When the user clears the box, `'' || initialSearch` falls back to the URL param, so the filter sticks.

**Fix.** Use `searchQuery` alone as the source of truth after the initial seed (don't OR-fallback to `initialSearch` on every render); sync the URL param from state instead.

---

larryro:
## Summary

Fixes #1992.

When a conversations lane was opened via a `?search=` URL param, clearing the search box did nothing — the list stayed filtered.

**Root cause.** In `services/platform/app/features/conversations/components/conversations.tsx`, `searchQuery` was seeded from the `initialSearch` prop, but the applied term was `const searchTerm = searchQuery || initialSearch;`. Once the user emptied the box, `'' || initialSearch` fell back to the stale URL param on every render, so the filter stuck.

## Fix

- `searchQuery` is now the single source of truth for the filter (seeded once from the URL param). The applied filter uses `searchQuery` alone — no OR-fallback to `initialSearch`.
- The `?search=` URL param is now kept in sync **from** state via `useNavigate` (mirrors the existing pattern in `customers-table.tsx`), using `replace: true` so typing doesn't spam history. Clearing the box drops the param.

## Tests

- Added `conversations.test.tsx`: seeds the filter from `search="Refund"`, asserts only the matching row shows, then clears the box and asserts every row reappears (fails against the old `|| initialSearch` code).
- `bunx tsc --noEmit` — green
- `bunx oxlint --type-aware` on changed files — green
- `bunx vitest --run --config vitest.ui.config.ts` for the new test — green

Closes #1992

### DIFF
--- a/services/platform/app/features/conversations/components/conversations.test.tsx
+++ b/services/platform/app/features/conversations/components/conversations.test.tsx
@@ -1,0 +1,116 @@
+import type { UsePaginatedQueryResult } from 'convex/react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import type { ConversationItem } from '@/convex/conversations/types';
+import { render, screen } from '@/tests/utils/render';
+
+import { Conversations } from './conversations';
+
+// ---------------------------------------------------------------------------
+// Regression coverage for #1992: a lane opened via `?search=` seeds the filter
+// from the URL param, but clearing the search box must actually clear the
+// filter (state is the source of truth — it must NOT fall back to the stale URL
+// param on every render).
+// ---------------------------------------------------------------------------
+
+// `useNavigate` is called to keep the URL in sync with the search state; in the
+// jsdom test there is no router, so a no-op navigate is enough.
+const navigateSpy = vi.fn();
+vi.mock('@tanstack/react-router', () => ({
+  useNavigate: () => navigateSpy,
+}));
+
+// The bulk-actions hook reaches for convex mutations; stub it out so the list
+// renders without a backend (mirrors customers-table.test.tsx).
+vi.mock('../hooks/use-bulk-actions', () => ({
+  useBulkActions: () => ({
+    isBulkProcessing: false,
+    bulkSendDialog: { isOpen: false, isSending: false },
+    openBulkSendDialog: vi.fn(),
+    closeBulkSendDialog: vi.fn(),
+    handleSendMessages: vi.fn(),
+    handleBulkResolve: vi.fn(),
+    handleBulkReopen: vi.fn(),
+    handleBulkSpam: vi.fn(),
+    handleBulkArchive: vi.fn(),
+    handleBulkUnarchive: vi.fn(),
+  }),
+}));
+
+// The reading pane pulls in convex queries/mutations; it is irrelevant to the
+// list-filtering behaviour under test, so stub it to a marker.
+vi.mock('./conversation-panel', () => ({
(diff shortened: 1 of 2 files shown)


larryro:
**Desk review — Issue #1992 (clearing the search box does nothing when opened via `?search=`)**

✅ **READY TO MERGE.** The fix is correct, minimal, well-tested, and CI is green.

### What was verified
- **Root cause fixed.** The filter no longer falls back to the stale URL param: `searchQuery` is now the single source of truth (`conversations.tsx:136`, deps cleaned at `:152`), seeded once from `initialSearch` (`:98`). Clearing the box sets `searchQuery=''` → filter clears. `handleSearchChange` (`:103-117`) keeps the URL in sync via `navigate({ replace: true })`, preserving `priority` through `...prev`.
- **No remount hazard.** The route `key` is `${organizationId}-${status}` (excludes `search`), so the per-keystroke URL update does not remount the component and the `useState` seed runs once — exactly what the fix relies on. Search filtering is client-side (the paginated query at `$status.tsx:92-97` is not keyed on `search`), so per-keystroke `navigate` does **not** thrash the backend query.
- **Regression genuinely guarded (verified empirically).** Restoring the old `searchQuery || initialSearch` logic locally makes the new test fail at the "all rows return after clear" assertion; with the fix it passes. Ran `bunx vitest --run --project client app/features/conversations/components/conversations.test.tsx` → **1 passed**.
- **Security/boundaries clean.** `filterByTextSearch` is a pure `.toLowerCase().includes()` substring match — no RegExp/eval, so no regex-injection or ReDoS, and it trims (whitespace-only → returns all items). Route validates `search` via `z.string().optional()`.
- **CI:** all checks **pass** (the `skipping` jobs are path-filtered conditionals, not failures).

### Non-blocking suggestions (optional, do not block merge)
1. `conversations.tsx:111` — a whitespace-only entry writes `?search=%20%20` to the URL (the filter still correctly ignores it). Consider `value.trim().length > 0 ? value : undefined` for a cleaner URL.
2. The test mocks `navigateSpy` but never asserts on it, so the URL-sync half of the fix is unasserted. The core regression (filter clears) is fully covered and a URL-sync break would not reproduce #1992, so this is a nice-to-have, not a blocker. Adding a `expect(navigateSpy).toHaveBeenCalledWith(...search: undefined...)` would harden it.
3. Per-keystroke `navigate` is undebounced; other search inputs in the repo use `useDebounce`. Harmless here (client-side filter, `replace:true`), but worth noting for consistency.

PR #2200, opened 2026-06-27
fix: treat missing retention config as empty state on governance load (#1997)

### LINKED ISSUE #1997: Improvement: Governance "Policies & Limits" / "Content & Models" log RETENTION_CONFIG_MISSING ConvexErrors on normal load
### What should be improved?
Opening the governance Policies & Limits / Content & Models pages logs three `ConvexError(RETENTION_CONFIG_MISSING)` to the console on a normal load (e.g. a fresh org with no retention config set).

### Current behavior
A not-yet-configured retention setting is surfaced as a thrown ConvexError at load time rather than a handled "not configured" state.

### Proposed change
Treat a missing retention config as an expected empty state (return null / default) instead of throwing, so normal page loads are console-clean.

### Affected service
Platform (web app)

Resilience / console-hygiene refinement.

---

larryro:
## Summary

Opening Governance **Policies & Limits** on a fresh org (no `retention.json` installed yet) logged `ConvexError(RETENTION_CONFIG_MISSING)` on every normal page load. The read-only retention actions threw on a not-yet-configured retention setting instead of treating it as an expected empty state.

This makes the **read** paths return a graceful empty state, while the **write** path still throws (you cannot save against bounds the operator has not installed):

- `getRetentionBoundsAction` (read, drives `useRetentionBounds`) now returns `{ bounds: [], retentionDisabled }` when the org has no retention config. The editor already falls back to schema-level defaults for categories without a bound row, so rendering is unaffected — just console-clean.
- `getPendingBoundsProposal` (read, drives the proposal banner) now returns `null` when the config is missing — there is nothing to propose until the operator installs the config. It bails out before reading applied bounds.
- `upsertRetentionPolicyAction`, `applyBoundsProposal`, `rejectBoundsProposal`, and `seedInitialBoundsInternal` are unchanged — write/apply paths still surface `RETENTION_CONFIG_MISSING`.

## Tests

Added `convex/governance/retention_config_missing.test.ts` pinning the behaviour:
- read paths return empty/null without throwing (and don't query applied bounds),
- `retentionDisabled` still reflects the env flag,
- the write path still throws a `ConvexError(RETENTION_CONFIG_MISSING)`.

Verified locally: `vitest --project server` (retention suites, 66+5 passing), `oxlint --type-aware`, and `tsc --noEmit` all green.

Closes #1997

### DIFF
--- a/services/platform/convex/governance/retention_actions.ts
+++ b/services/platform/convex/governance/retention_actions.ts
@@ -116,10 +116,14 @@ export const getRetentionBoundsAction = action({
     const orgSlug = await resolveOrgSlug(ctx, args.organizationId);
     const orgConfig = await loadOrgRetentionConfig(ctx, orgSlug);
     if (!orgConfig) {
-      throw new ConvexError({
-        code: 'RETENTION_CONFIG_MISSING',
-        message: `Retention config not yet installed. Copy builtin-configs/governance/retention.json to $TALE_CONFIG_DIR/${orgSlug}/governance/retention.json then reload.`,
-      });
+      // A not-yet-installed retention config is an expected empty state on
+      // normal page loads (e.g. a fresh org), not an error. Return empty
+      // bounds so the editor falls back to schema-level defaults instead of
+      // logging a RETENTION_CONFIG_MISSING ConvexError on every load.
+      return {
+        bounds: [],
+        retentionDisabled: isRetentionDisabled(),
+      };
     }
 
     let bounds;
--- a/services/platform/convex/governance/retention_bounds_proposal.ts
+++ b/services/platform/convex/governance/retention_bounds_proposal.ts
@@ -303,7 +303,26 @@ export const getPendingBoundsProposal = action({
     });
 
     const orgSlug = await resolveOrgSlug(ctx, args.organizationId);
-    const proposedBounds = await computeEffectiveAppliedBounds(ctx, orgSlug);
+    // A not-yet-installed retention config is an expected empty state on
+    // normal page loads (e.g. a fresh org), not an error: there is no
+    // proposal to surface until the operator installs the config. Return
+    // null (banner stays silent) instead of throwing a
+    // RETENTION_CONFIG_MISSING ConvexError on every load. The write paths
+    // (apply/reject/seed) still throw — you cannot act on bounds that the
+    // operator has not installed yet.
+    let proposedBounds: AppliedBoundsByCategory;
+    try {
+      proposedBounds = await computeEffectiveAppliedBounds(ctx, orgSlug);
+    } catch (err) {
+      if (
+        err instanceof ConvexError &&
+        isRecord(err.data) &&
+        err.data.code === 'RETENTION_CONFIG_MISSING'
+      ) {
+        return null;
+      }
+      throw err;
+    }
     const proposedHash = await hashAppliedBounds(proposedBounds);
 
     const applied: {
--- a/services/platform/convex/governance/retention_config_missing.test.ts
+++ b/services/platform/convex/governance/retention_config_missing.test.ts
@@ -1,0 +1,172 @@
+import { ConvexError } from 'convex/values';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Regression coverage for #1997: opening Governance "Policies & Limits"
+// on a fresh org (no retention.json installed yet) used to log three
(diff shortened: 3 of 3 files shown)


larryro:
## Desk Review — PR #2200 (fixes #1997)

**VERDICT: READY TO MERGE** (one non-blocking recommendation noted below.)

`fix: treat missing retention config as empty state on governance load (#1997)`
Branch `tale/xs7az7hxpypemyb84n1bm9sabx89fb2e` @ `b485c51`. Single canonical OPEN PR for #1997 — no ambiguity.

### CI state
All checks **GREEN** (`gh pr checks` → every job `pass`; a few fork-only jobs `skipping`; none failed, none pending). `mergeable_state: clean`.

### Tests (run once locally)
- `bunx vitest --run --project server retention` → **6 files / 101 tests passing**.
- New file `convex/governance/retention_config_missing.test.ts` → **5/5 passing**.
- Underlying logic happy-path is already covered by `retention_floors.test.ts` (env-tightening behavior with a present config + the missing-category throw).

### What the fix does (verified by tracing every branch)
The reported issue is a **fresh org with no `retention.json`** (config file absent → `loadOrgRetentionConfig` returns `null`). Both load-time READ paths now stay console-clean:
- `getRetentionBoundsAction` (`retention_actions.ts:118`): `if (!orgConfig) return { bounds: [], retentionDisabled: isRetentionDisabled() }` — never calls `applyEnvTighteningAll`. ✔ silent.
- `getPendingBoundsProposal` (`retention_bounds_proposal.ts:313-325`): `computeEffectiveAppliedBounds` throws `RETENTION_CONFIG_MISSING`; the new try/catch returns `null`. ✔ silent.

Confirmed correct:
- **WRITE/apply paths still throw** `RETENTION_CONFIG_MISSING`: `upsertRetentionPolicyAction` (`retention_actions.ts:191`), `applyBoundsProposal` / `rejectBoundsProposal` (via un-caught `computeEffectiveAppliedBounds`), `seedInitialBoundsInternal` skips gracefully (migration helper). No way to save against bounds the operator hasn't installed.
- **No auth bypass** — both early returns happen *after* `getAuthUserIdentity` + `verifyOrgMember`.
- **`ConvexError.data` guard is correct** — `new ConvexError({code,...})` stores the payload on `.data`; the `err instanceof ConvexError && isRecord(err.data) && err.data.code === 'RETENTION_CONFIG_MISSING'` check matches the throw, and mirrors the pre-existing pattern in `seedInitialBoundsInternal`.
- **Return values conform** to the Convex `returns` validators (empty `bounds` array; `null` union).
- **Frontend handles empty bounds** — `useRetentionBounds` does `result.data?.bounds ?? []` → categories fall back to schema-level defaults; the proposal banner renders nothing on `null`. No NaN / broken `<input min/max>`.

### Non-blocking recommendation (not a merge blocker)
**Read-path asymmetry on a *partial* config file.** The retention schema permits partial files (each category is `.optional()`). If a config file is *present but missing a category*, `applyEnvTighteningAll` throws `RetentionConfigMissingError` → `RETENTION_CONFIG_MISSING`. In that (abnormal) case the two read paths diverge: `getRetentionBoundsAction` still throws/surfaces it, while `getPendingBoundsProposal`'s broad `code === 'RETENTION_CONFIG_MISSING'` catch swallows it and returns `null`.

This does **not** block merge because: (a) the reported #1997 scenario is the *absent-file* case, which both paths handle correctly; (b) orgs are seeded with a complete built-in catalog at creation, so a partial file is an abnormal operator state, not normal load; (c) the swallow mirrors the established `seedInitialBoundsInternal` convention and "nothing to propose when the config is incomplete" is defensible; (d) even then the bounds action still surfaces the misconfig on the same page. Optional hardening for a follow-up: have `getPendingBoundsProposal` only treat the *absent-file* case as the empty state (mirroring `getRetentionBoundsAction`'s `!orgConfig` early return) so a genuinely broken partial file isn't silently hidden — and add a test for the partial-file branch.

### Method
Round 1: 5 parallel review subagents (correctness / robustness / elegance / tests / boundaries) sized to this small focused PR. Round 2: independently re-verified every claimed blocker against the source — the schema (partial files valid → mode-2 reachable), `retention_floors.test.ts` (happy path covered → "no happy-path test" downgraded to non-blocking), the mode-1 trace, and `ConvexError.data`. No finding survived as a blocker.

**READY TO MERGE.**

PR #2201, opened 2026-06-27
fix: knowledge_entries mutations throw ConvexError instead of raw Error (#2000)

### LINKED ISSUE #2000: Bug: knowledge_entries mutations throw raw Error instead of ConvexError — opaque Server Errors on client
## Area
`services/platform/convex/knowledge_entries/` — public mutations (`createKnowledgeEntry`, `updateKnowledgeEntry`, `deleteKnowledgeEntry`) and their shared helper (`validateTopicAndContent`).

## Steps to reproduce
1. Open the Knowledge Entries UI.
2. Try to create a knowledge entry with an empty topic (or content exceeding the max length), or with a topic that already exists.
3. Observe the browser console — a raw "Server Error" is surfaced instead of a structured, readable error.

Alternatively:
- Edit a superseded (non-active) entry.
- Delete a knowledge entry that no longer exists.

## Expected
The client receives a `ConvexError` with a structured `code` field (following the pattern used in `tasks/mutations.ts` and `branding/file_actions.ts`) so the UI can display a user-readable message.

## Actual
`throw new Error(...)` is used throughout, which Convex surfaces as an opaque "Server Error" in the browser console; the message is lost to the client.

## Evidence
`services/platform/convex/knowledge_entries/helpers.ts:46` — `throw new Error('Topic is required')`
`services/platform/convex/knowledge_entries/helpers.ts:48` — `throw new Error(\`Topic exceeds ${TOPIC_MAX_LENGTH} characters\`)`
`services/platform/convex/knowledge_entries/helpers.ts:50` — `throw new Error('Content is required')`
`services/platform/convex/knowledge_entries/helpers.ts:52` — `throw new Error(\`Content exceeds ${CONTENT_MAX_LENGTH} characters\`)`
`services/platform/convex/knowledge_entries/mutations.ts:34` — `throw new Error('Unauthenticated')` (createKnowledgeEntry)
`services/platform/convex/knowledge_entries/mutations.ts:53` — `throw new Error(\`A knowledge entry for the topic "..." already exists...\`)`
`services/platform/convex/knowledge_entries/mutations.ts:86` — `throw new Error('Unauthenticated')` (updateKnowledgeEntry)
`services/platform/convex/knowledge_entries/mutations.ts:90` — `throw new Error('Knowledge entry not found')`
`services/platform/convex/knowledge_entries/mutations.ts:93` — `throw new Error('Only the active version of an entry can be edited')`
`services/platform/convex/knowledge_entries/mutations.ts:116` — `throw new Error(\`A knowledge entry for the topic "..." already exists.\`)`
`services/platform/convex/knowledge_entries/mutations.ts:172` — `throw new Error('Unauthenticated')` (deleteKnowledgeEntry)
`services/platform/convex/knowledge_entries/mutations.ts:176` — `throw new Error('Knowledge entry not found')` (deleteKnowledgeEntry)

## Notes
Related: #1993 (same pattern in websites/customers/vendors). The fix is to import `ConvexError` from `"convex/values"` and replace each throw with `throw new ConvexError({ code: '...' })` matching the structured pattern used in `tasks/mutations.ts`. `validateTopicAndContent` in `helpers.ts` is called from both public mutations and internal mutations; switching it to `ConvexError` is safe since `ConvexError` propagates correctly through both paths.

Severity: medium — user-facing validation and auth errors silently become opaque "Server Error" in the console; the UI typically shows no feedback when these mutations reject.

larryro:
## What

`services/platform/convex/knowledge_entries/` public mutations and their shared `validateTopicAndContent` helper threw raw `Error(...)`, which Convex redacts to an opaque **"Server Error"** in production — the message never reaches the client, so the Knowledge Entries UI shows no feedback when a mutation rejects.

This replaces every raw `Error` throw in the public surface with a structured `ConvexError({ code: '...' })`, matching the pattern already used in `tasks/mutations.ts`, `branding/file_actions.ts`, and the existing `KNOWLEDGE_ENTRY_DUPLICATE` guard in this same file.

## Changes

`helpers.ts` — `validateTopicAndContent` now throws:
- `KNOWLEDGE_ENTRY_TOPIC_REQUIRED`
- `KNOWLEDGE_ENTRY_TOPIC_TOO_LONG` (carries `max`)
- `KNOWLEDGE_ENTRY_CONTENT_REQUIRED`
- `KNOWLEDGE_ENTRY_CONTENT_TOO_LONG` (carries `max`)

`mutations.ts` — create/update/delete now throw:
- `UNAUTHENTICATED` (matching the repo-wide convention)
- `KNOWLEDGE_ENTRY_NOT_FOUND`
- `KNOWLEDGE_ENTRY_NOT_ACTIVE`

The helper is shared by both public and internal mutations; `ConvexError` propagates correctly through both paths, so this is safe.

## Tests

- Extended `mutations.test.ts` with a `#2000` suite asserting the structured codes for unauthenticated / blank-topic / blank-content / not-found / not-active.
- Updated `helpers.test.ts` validation cases to assert the new `ConvexError` codes instead of the old raw message strings.
- `bunx vitest --run --project server knowledge_entries` → 25 passed.
- `bunx tsc --noEmit` → clean. `bunx oxlint --type-aware` → 0 warnings / 0 errors.

Closes #2000

### DIFF
--- a/services/platform/convex/knowledge_entries/helpers.test.ts
+++ b/services/platform/convex/knowledge_entries/helpers.test.ts
@@ -1,3 +1,4 @@
+import { ConvexError } from 'convex/values';
 import { describe, expect, it, vi } from 'vitest';
 
 import { normalizeTopicKey } from './constants';
@@ -127,28 +128,46 @@ describe('validateTopicAndContent', () => {
     expect(result.content).toBe('Open 9-5');
   });
 
+  // #2000: validation rejects with structured ConvexError codes (not raw
+  // `Error`), so the client receives a readable code instead of an opaque
+  // "Server Error".
+  function codeOf(fn: () => unknown): string | undefined {
+    try {
+      fn();
+    } catch (err) {
+      if (!(err instanceof ConvexError)) return undefined;
+      const data: unknown = err.data;
+      if (typeof data !== 'object' || data === null || !('code' in data)) {
+        return undefined;
+      }
+      const candidate: unknown = (data as { code: unknown }).code;
+      return typeof candidate === 'string' ? candidate : undefined;
+    }
+    return undefined;
+  }
+
   it('rejects empty topic', () => {
-    expect(() => validateTopicAndContent('   ', 'content')).toThrow(
-      'Topic is required',
+    expect(codeOf(() => validateTopicAndContent('   ', 'content'))).toBe(
+      'KNOWLEDGE_ENTRY_TOPIC_REQUIRED',
     );
   });
 
   it('rejects topic over the cap', () => {
-    expect(() => validateTopicAndContent('x'.repeat(121), 'content')).toThrow(
-      '120',
-    );
+    expect(
+      codeOf(() => validateTopicAndContent('x'.repeat(121), 'content')),
+    ).toBe('KNOWLEDGE_ENTRY_TOPIC_TOO_LONG');
   });
 
   it('rejects empty content', () => {
-    expect(() => validateTopicAndContent('topic', '   ')).toThrow(
-      'Content is required',
+    expect(codeOf(() => validateTopicAndContent('topic', '   '))).toBe(
+      'KNOWLEDGE_ENTRY_CONTENT_REQUIRED',
     );
   });
 
   it('rejects content over the cap', () => {
-    expect(() => validateTopicAndContent('topic', 'x'.repeat(8001))).toThrow(
-      '8000',
-    );
+    expect(
+      codeOf(() => validateTopicAndContent('topic', 'x'.repeat(8001))),
+    ).toBe('KNOWLEDGE_ENTRY_CONTENT_TOO_LONG');
   });
 });
 
--- a/services/platform/convex/knowledge_entries/helpers.ts
+++ b/services/platform/convex/knowledge_entries/helpers.ts
@@ -5,6 +5,8 @@
  * bundle (or their tests' mock surface).
  */
 
+import { ConvexError } from 'convex/values';
+
 import type { Doc, Id } from '../_generated/dataModel';
 import type { MutationCtx } from '../_generated/server';
 import {
@@ -43,13 +45,27 @@ export function validateTopicAndContent(
 ): { topic: string; topicKey: string; content: string } {
   const trimmedTopic = topic.trim();
   const trimmedContent = content.trim();
-  if (!trimmedTopic) throw new Error('Topic is required');
-  if (trimmedTopic.length > TOPIC_MAX_LENGTH) {
-    throw new Error(`Topic exceeds ${TOPIC_MAX_LENGTH} characters`);
+  // Structured ConvexError codes so the client surfaces a readable message
+  // instead of an opaque "Server Error" (raw `Error` messages are redacted by
+  // Convex in prod). Called from both public and internal mutations;
+  // ConvexError propagates correctly through both paths.
+  if (!trimmedTopic) {
+    throw new ConvexError({ code: 'KNOWLEDGE_ENTRY_TOPIC_REQUIRED' });
   }
-  if (!trimmedContent) throw new Error('Content is required');
+  if (trimmedTopic.length > TOPIC_MAX_LENGTH) {
+    throw new ConvexError({
+      code: 'KNOWLEDGE_ENTRY_TOPIC_TOO_LONG',
+      max: TOPIC_MAX_LENGTH,
+    });
+  }
+  if (!trimmedContent) {
+    throw new ConvexError({ code: 'KNOWLEDGE_ENTRY_CONTENT_REQUIRED' });
+  }
   if (trimmedContent.length > CONTENT_MAX_LENGTH) {
-    throw new Error(`Content exceeds ${CONTENT_MAX_LENGTH} characters`);
+    throw new ConvexError({
+      code: 'KNOWLEDGE_ENTRY_CONTENT_TOO_LONG',
+      max: CONTENT_MAX_LENGTH,
+    });
   }
   return {
     topic: trimmedTopic,
--- a/services/platform/convex/knowledge_entries/mutations.test.ts
+++ b/services/platform/convex/knowledge_entries/mutations.test.ts
@@ -29,7 +29,11 @@ vi.mock('../lib/rls/organization/get_organization_member', () => ({
 
 import { ConvexError } from 'convex/values';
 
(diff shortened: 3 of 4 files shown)


larryro:
## Desk Review — #2000 (ConvexError migration for `knowledge_entries`)

**Verdict: NOT READY — one change required (test coverage gap).** The implementation itself is correct and complete; the only gap is in the dedicated `#2000` regression suite.

### What I verified
- **Tests:** `bunx vitest --run knowledge_entries` (in `services/platform`) → **25/25 passing**.
- **CI:** `gh pr checks` → **46 pass, 5 skipping (fork-PR / unrelated containers), 0 fail, 0 pending** — green.

### Correctness — complete ✅
Every in-scope raw `throw new Error(...)` is now `throw new ConvexError({ code })`:
- `helpers.ts` `validateTopicAndContent`: `KNOWLEDGE_ENTRY_TOPIC_REQUIRED`, `…_TOPIC_TOO_LONG` (+`max`), `…_CONTENT_REQUIRED`, `…_CONTENT_TOO_LONG` (+`max`).
- `mutations.ts`: `UNAUTHENTICATED` (create/update/delete), `KNOWLEDGE_ENTRY_NOT_FOUND` (update/delete), `KNOWLEDGE_ENTRY_NOT_ACTIVE` (update), `KNOWLEDGE_ENTRY_DUPLICATE` (create + update-rename). No raw `throw new Error` remains in the module. Codes match the `tasks/mutations.ts` convention (bare `UNAUTHENTICATED`, domain-prefixed otherwise).

### Robustness — safe ✅
- `ConvexError extends Error`, so the layout error-boundary's `error.message` "Unauthenticated" check still works, and `internal_actions.ts`'s `instanceof Error` catch still works.
- Client already consumes codes via `convexErrorCode()`; `KNOWLEDGE_ENTRY_DUPLICATE` is handled in the add/edit dialogs. Other new codes are not yet branched on the client (expected — separate follow-up per the issue).
- All throws occur before any DB write → no partial-write risk. `topic`/`max` payloads are the user's own authorized org data → safe to expose.

### ⚠️ Blocking: missing regression tests for two converted branches
The new `knowledge entry structured errors (#2000)` suite covers `createKnowledgeEntry`'s `UNAUTHENTICATED` branch but **not** the two other `UNAUTHENTICATED` branches this PR converts:
- `updateKnowledgeEntry` — `mutations.ts:90`
- `deleteKnowledgeEntry` — `mutations.ts:177`

Both are correct (one-line copies of the tested create branch), but since this is the dedicated regression suite locking in #2000's behavior, those two error branches should be covered. Add a test for each that sets `ctx.auth.getUserIdentity` to resolve `null` and asserts the thrown code is `UNAUTHENTICATED` (mirror the existing create test at `mutations.test.ts:150`).

### Non-blocking note (pre-existing, out of scope)
`updateKnowledgeEntry`/`deleteKnowledgeEntry` throw `NOT_FOUND`/`NOT_ACTIVE` before `getOrganizationMember`, so a caller could probe entry existence across orgs. This is unchanged by this PR (only the error *type* changed) and matches the `tasks/mutations.ts` pattern — flagging for awareness only, not for this PR.

larryro:
## Desk Review — PR #2201 (Closes #2000)

**Verdict: READY TO MERGE.**

Reviewed the whole implementation (not just the diff) on branch `tale/xs7btmj8hyhss74khbgpcdjqw989fdvr`, including the internal callers of the shared helper. Two-round review (breadth across the 5 review dimensions, then independent confirmation of every point that could block).

### CI
All required checks **green** — Type check, Lint, Format, Knip, Unit, Build convex/platform, Smoke test, all 16 Playwright (platform) shards, UI, Storybook, Analyze, Opengrep. The only non-green entries are normal skips (`Trivy`, fork-PR-only jobs). No red, none pending.

### Tests
Ran the project's own suite once:
`bunx vitest --run --project server knowledge_entries` → **27 passed (2 files)**.

### What was verified
- **Correctness across all branches** — every raw `throw new Error(...)` in the public surface (`createKnowledgeEntry`, `updateKnowledgeEntry`, `deleteKnowledgeEntry`, and the shared `validateTopicAndContent`) is converted to `ConvexError({ code })`. Each branch carries the correct code (`UNAUTHENTICATED`, `KNOWLEDGE_ENTRY_{TOPIC,CONTENT}_{REQUIRED,TOO_LONG}`, `KNOWLEDGE_ENTRY_NOT_FOUND`, `KNOWLEDGE_ENTRY_NOT_ACTIVE`), consistent with the pre-existing `KNOWLEDGE_ENTRY_DUPLICATE` guard. No half-converted path.
- **Shared helper is safe for internal callers** — `validateTopicAndContent` is also called by `internal_mutations.ts` (`createKnowledgeWriteApproval`, `applyKnowledgeWrite`). Neither matches on the error message; `ConvexError` is an `Error` subclass and propagates identically. The one `try/catch` in `internal_actions.ts:164` uses `error.message` only for logging/recording, so the type switch is harmless there.
- **Out-of-scope raw `Error`s left intact correctly** — remaining `throw new Error(...)` in `internal_mutations.ts` / `internal_actions.ts` are internal invariants, not client-facing, and outside issue #2000's scope.
- **Robustness** — `ConvexError` imported correctly in both files; numeric `max` payload is serializable (matches established repo usage); all throws occur before any `ctx.db` write/scheduler call, so no partial writes and no new ordering/concurrency hazard.
- **Elegance** — minimal, focused; code naming and the locally-defined `codeOf` test helper match repo precedent (e.g. `tasks/mutations.ts`, `approvals/error_codes.test.ts`). No dead code.
- **Resolution** — genuinely fixes the reported symptom: the client now receives structured codes instead of an opaque "Server Error". The add/edit dialogs already consume `KNOWLEDGE_ENTRY_DUPLICATE`; the issue's "Expected" only requires the client to *receive* a structured code, which it now does.

### Non-blocking suggestions (optional, do not gate merge)
1. `KNOWLEDGE_ENTRY_TOPIC_TOO_LONG` / `KNOWLEDGE_ENTRY_CONTENT_TOO_LONG` are asserted at the helper level (`helpers.test.ts`) but not at the mutation level. Coverage is adequate since the helper is the single source of those throws, but a mutation-level case would round it out.
2. Follow-up (separate from #2000): the dialogs map only `KNOWLEDGE_ENTRY_DUPLICATE`; the other new codes fall to the generic error toast. Mapping `NOT_FOUND` / `NOT_ACTIVE` to specific messages would improve UX.

**READY TO MERGE.**

PR #2204, opened 2026-06-27
fix(platform): close data-quality and affordance gaps (#1999)

### LINKED ISSUE #1999: Improvement: Data-quality + affordance gaps — duplicate display names allowed (agents, products), no secret reveal/edit, document accept-hint omits TXT
### What should be improved?
A few low-severity data-quality and affordance gaps found in the deep pass.

### Current behavior
1. Duplicate display names are silently allowed for agents and products (only the slug is unique). 2. The project secret store offers only Add and Delete — no reveal or edit affordance. 3. The document-upload accept hint omits TXT, although TXT uploads succeed.

### Proposed change
1. Warn on (or prevent) duplicate display names where it causes ambiguity. 2. Add a reveal/edit affordance to project secrets. 3. Add TXT to the upload accept hint.

### Affected service
Platform (web app)

Data-quality / affordance refinements.

---

larryro:
## Summary

Resolves the three low-severity data-quality / affordance gaps from the deep manual-QA sweep in #1999.

### 1. Duplicate display names
- **Products** — a product's `name` IS its identity (there is no separate slug), so a duplicate name is genuine ambiguity. `assertUniqueProductName` now rejects a case-insensitive, trimmed duplicate at **create** and **update** (excluding the row being renamed). Surfaced as a field error in the edit dialog and a toast in the create wizard. Bulk import keeps its own path and is unaffected.
- **Agents** — the slug is already unique server-side, so a duplicate display name is only cosmetic. The create dialog now shows a **non-blocking warning** under the Display name field when the entered name already belongs to another agent.

### 2. Project secret affordances
- Added an **Edit** affordance to each secret: it reuses the existing `setProjectSecret` upsert to re-encrypt a new value (the name — the env-var key agents resolve — stays fixed).
- **Reveal is intentionally not added.** The secret store deliberately never returns plaintext to the client (schema docs + the user-facing *"Values are never shown after saving"* / *"never revealed in the UI"* strings). Edit covers the real gap (rotating/replacing a value) without weakening that guarantee.

### 3. Document upload accept hint
- The dropzone hint omitted TXT even though TXT uploads are accepted. Added `TXT` to `documents.upload.dropZoneDescription` in en/de/fr (docs already listed plain text).

## Tests / checks run
- `assert_unique_product_name.test.ts` (new): unique passes, case/trim collision throws, rename-self excluded, empty-name skip.
- `bun run --filter @tale/platform typecheck` — clean
- `bun run --filter @tale/platform lint` — 0 warnings / 0 errors
- `bun run format:check` — clean
- `bun run knip:check` — clean
- `bun run lint:sast` (Opengrep) — 0 findings
- Platform unit suite green for the touched areas (products / messages parity / secrets).

i18n keys added to en/de/fr (parity test green). No Convex schema change → no migration.

Closes #1999


larryro:
## Desk review — #1999 fix (data-quality + affordance gaps)

**Verdict: NOT READY — CHANGES REQUIRED** (test-coverage gap on new branching behavior).

### What I checked
- Branch `tale/xs7agjbdyrrc2ykz335wjzpaeh89fj8z` checked out from `main`.
- **CI: fully green.** All required checks pass (Build platform, Type check, Lint, Format, Knip, Opengrep, Unit, Storybook, UI, 16 Playwright platform shards, Playwright docs/web, Migrations, Smoke, Validate images, CodeQL). Remaining entries are `skipping` no-ops (fork-PR / Trivy / container tests).
- **Tests I ran locally (all pass):**
  - `bunx vitest --run assert_unique_product_name` → 5 passed
  - `bunx vitest --run --config vitest.ui.config.ts agent-create-dialog` → 7 passed
  - `bunx vitest --run --project server convex/products` → 11 passed
  - `bunx vitest --run --project server messages` → 37 passed
  - `bunx vitest --run --project server lib/i18n/messages` (i18n parity) → 23 passed
- Verified the TXT upload-hint is **accurate**: `text/plain` + `.txt` are in `DOCUMENT_UPLOAD_ALLOWED_TYPES`, `_ALLOWED_EXTENSIONS`, and `DOCUMENT_UPLOAD_ACCEPT` (`lib/shared/file-types.ts`). en/de/fr parity confirmed.
- Backend duplicate-name logic (`assert_unique_product_name.ts`) is correct: trim+lowercase normalize, `by_organizationId` index exists, exclude-self on rename works, empty-name skips. Wired into `createProduct` and `updateProduct` mutations.

### Blocking — must fix
1. **`project-secrets-tab.tsx`: the new Edit affordance has zero test coverage.** The component test (`project-secrets-tab.test.tsx`) covers only render / add / delete, and the e2e `projects-depth.spec.ts` ("create then delete an API-key secret") never clicks Edit. This is the most stateful change in the PR — a new `editingName` state machine threaded through `openEdit()` (seeds name+description, forces `type='custom'`), `isEditing` conditional rendering, the name field being disabled, the value field being `required`, the edit branch of `handleSave` (upsert with the fixed name), and `resetForm()` clearing `editingName` on close. Add a unit test (mirroring the existing add/delete tests) that: opens Edit on an existing row, asserts the name input is disabled and prefilled, submits a new value, asserts `setProjectSecret` is called with the original (fixed) name + new value, and asserts that after closing, the next Add dialog is NOT stuck in edit mode.

2. **Product duplicate-name error handling in the dialogs is untested.** Only the backend helper is tested in isolation; the dialog branches that translate a `DUPLICATE_PRODUCT_NAME` ConvexError into a toast (`product-create-dialog.tsx`) and into a `name` field error (`product-edit-dialog.tsx`) are not exercised by any test. Add a test that simulates the mutation rejecting with `new ConvexError({ code: 'DUPLICATE_PRODUCT_NAME' })` and asserts the create dialog shows `create.toast.duplicateName` and the edit dialog sets the `name` field error — so the user-facing half of the fix is actually verified.

### Non-blocking — recommended
- **`product-create-dialog.tsx:161` / `product-edit-dialog.tsx:154` — `instanceof ConvexError`:** matches the dominant repo convention (e.g. `task-dependencies.tsx`, `task-modal.tsx`), so this is not a regression. But the repo ships `app/features/prompts/lib/extract-error-code.ts` *specifically because* Vite chunk splitting can produce multiple `ConvexError` copies that break `instanceof`. Consider `extractErrorCode(err) === 'DUPLICATE_PRODUCT_NAME'` for robustness. Adding the tests in blocking item 2 will also surface this if it ever breaks.
- **Uniqueness is UI-only.** `createProduct`/`updateProduct` mutations enforce it, but the REST paths (`internal_mutations.ingestProduct` / `updateProducts`), bulk import, workflow engine, and agent product-write tool do **not** call `assertUniqueProductName`. A duplicate created via those paths can later block a legitimate UI rename. The PR documents this scoping; acceptable for #1999, but worth a follow-up note.
- **Secrets edit UX:** the value field is `required` in edit mode, so a user wanting to change only the description must re-enter the (hidden) value. Minor; acceptable given the never-reveal guarantee.
- Consider a `products.errors` i18n section + error-code registry mirroring the `projects/error_codes.ts` pattern, rather than an inline string code.

### Confirmed sound (no action)
- Agent duplicate display-name warning is non-blocking and tested (case-insensitive warn + no-warn-when-unique). Locale-resolved comparison and the `if (!rawAgents) return` guard are fine.
- i18n parity across en/de/fr is complete and the parity test passes.
- TXT hint is truthful.

**Verdict: NOT READY — CHANGES REQUIRED.** Add tests covering (1) the project-secret Edit flow and (2) the product duplicate-name error handling in the create/edit dialogs. The implementation itself reads as correct and CI is green, but two of the three core deliverables ship without tests for their new branches.

larryro:
## Desk Review — PR #2204 (issue #1999: data-quality + affordance gaps)

**VERDICT: NOT READY — CHANGES REQUIRED.** Two confirmed defects in the product duplicate-name UX (the central deliverable). CI is green and the existing suite passes, but two of the three features don't behave correctly in the real app, and the bugs are masked by tests that stub out the very hooks that cause them.

### CI state
`gh pr checks` — **all green**: every check is `pass` or `skipping`; none failed or pending. (Skipped: Docs/Web container tests, fork-PR smoke/validate, Trivy — all the fork-gated jobs. The substantive jobs — Build platform/convex, Type check, Lint, Unit, UI, Knip, Format, Migrations, Opengrep, CodeQL, all 16 Playwright shards — pass.)

### Tests I ran (this branch, after `bun install`)
- `bunx vitest --run --project server convex/products/assert_unique_product_name.test.ts` → **5 passed**
- `bunx vitest --run --config vitest.ui.config.ts` on the 4 changed dialog/tab specs → **17 passed**
- `bunx vitest --run lib/i18n/messages.test.ts` (i18n parity/usage) → **23 passed**
- A throwaway reproduction of the edit-dialog rollback (see blocker #1) → **reproduced the defect**, then deleted (no code left behind).

---

### BLOCKING

**1. Product EDIT dialog: the inline duplicate-name error does not work in production.**
`useUpdateProduct` defines an `optimisticUpdate` that patches `name` into the cached product (`app/features/products/hooks/mutations.ts:44-71`). The edit dialog runs `reset({...product})` on **every** change of the `product` prop (`product-edit-dialog.tsx:117-128`), and react-hook-form's `reset` clears field errors by default. Real sequence on a duplicate-name rename:
1. Submit → optimistic update sets cached `product.name` to the new value → `product` prop identity changes → `useEffect` fires `reset()`.
2. Server rejects `DUPLICATE_PRODUCT_NAME` → `onError` calls `setError('name', …)` (`:155-159`).
3. Optimistic update **rolls back** → `product.name` reverts → `product` prop identity changes again → `useEffect` fires `reset()` again, **clearing the field error and reverting the user's typed name to the original.**

Net effect in the app: the user sees **no error** and their input is silently discarded — the feature's core deliverable is defeated. I confirmed this empirically with a faithful repro (rerender with a reverted `product` prop wipes the `edit.toast.duplicateName` error and resets the input to "Original name"). The existing test (`product-edit-dialog.test.tsx`) passes only because it stubs `useUpdateProduct` as a bare `vi.fn()` — no optimistic update, no prop change — so it never exercises the rollback.
**Fix:** stop resetting the form on every `product` change while the dialog is open/dirty (e.g. snapshot the product at open time and reset only on the open transition), or otherwise make the duplicate-name error survive the optimistic rollback. Then add a test that drives the optimistic-update + rollback path (not a stubbed hook).

**2. Product CREATE dialog: two contradictory destructive toasts on error.**
`useCreateProduct` (`app/features/products/hooks/mutations.ts:13-15`) does **not** pass `errorToast: false`, so the shared `useConvexMutation` always fires a generic destructive toast (`toast.error.generic`, `app/hooks/use-convex-mutation.ts:55-66`) — and then the dialog's own `onError` fires the specific toast (`create.toast.duplicateName` / `create.toast.error`, `product-create-dialog.tsx:158-170`). On a duplicate name the user sees **both** "Something went wrong" and "A product with this name already exists." The sibling hooks `useUpdateProduct` (`:40-43`) and `useDeleteProduct` (`:21-24`) both correctly set `errorToast: false` with the comment "shows its own specific error toast"; create is the inconsistent outlier. The dialog tests don't catch it because they mock `../hooks/mutations` wholesale, bypassing the shared hook's toast.
**Fix:** add `errorToast: false` to `useCreateProduct` so only the dialog's specific toast shows.

---

### NON-BLOCKING (should address or explicitly confirm as out-of-scope)

3. **Bulk product import bypasses uniqueness.** `bulkCreateProducts` → `createProductWithTranslations` never calls `assertUniqueProductName` (`convex/products/bulk_create_products.ts`). A CSV import can create duplicate product names and collide with dialog-created ones. The PR scopes this out by design; please confirm that's intended, since it leaves the "prevent duplicate products" requirement only half-enforced.
4. **Agent rename has no duplicate warning.** The display-name warning exists only in the *create* dialog; the agent edit page (`routes/dashboard/$id/agents/$agentId/index.tsx`) lets you rename to a duplicate display name with no warning. The issue names agents broadly; this addresses only creation. (Non-blocking because the warning is cosmetic/non-blocking by design.)
5. **Product name non-empty is enforced only client-side** (zod `min(1)`). `assertUniqueProductName` skips empty/whitespace names and no server validator rejects them, so a direct mutation can store an empty-named product. Pre-existing; minor.
6. **`assertUniqueProductName` does a full per-org product scan** rather than an indexed `.first()` lookup (diverges from the website/slug uniqueness patterns, which use a composite index). Correct, but O(n) per create/update — consider a normalized-name index at scale.
7. **Secrets edit: description-only edit is impossible.** In edit mode the value field is always `required` and `handleSave` always sends `value`, and `isDirty` is true as soon as a prefilled description exists — so Save can appear enabled with an empty value, which native `required`/the server then block. No data loss (the server rejects empty values before upsert — confirmed no clobber), but the UX is a mild dead-end. Low.
8. **Upload hint still understates accepted types.** The issue item (TXT) is fixed — TXT is genuinely accepted and now listed. But the dropzone also accepts DOC, PPT/PPTX, XLS, and images (JPEG/PNG/GIF/WEBP) that the hint still omits. Pre-existing; low.
9. **Naming nit:** `products.edit.toast.duplicateName` is rendered as an inline field error, not a toast, so the `toast.` key segment misleads. Consider `products.edit.errors.duplicateName`.

---

### What's solid (verified)
- Backend `assertUniqueProductName` is correct and well-tested: case-insensitive + trimmed match, `excludeProductId` lets an in-place rename keep its name, empty-name skips the scan. Wired into both `createProduct` and `updateProduct` (with `args.name === undefined` correctly skipping).
- Concurrency is safe — Convex's serializable/OCC model retries a racing create so a duplicate is still caught.
- Secrets edit is secure: no reveal/plaintext leak (the value is never returned to the client; edit only writes), the name is fixed and the mutation keys off `editingName`, and `setProjectSecret` enforces `requireAdmin` + project-admin server-side. No empty-value clobber.
- i18n parity passes for all new keys across en/de/fr; translations are real (no English leakage).

**VERDICT: NOT READY — CHANGES REQUIRED** (blockers #1 and #2 above).

larryro:
## Desk review — PR #2204 (issue #1999)

Reviewed via two rounds (7 breadth subagents across correctness/robustness/elegance/tests/boundaries, then independent confirmation of every blocking candidate). Setup: cloned, checked out `tale/xs7agjbdyrrc2ykz335wjzpaeh89fj8z`, ran the touched suites once.

### CI & tests (observed)
- **CI: all green.** `gh pr checks` — every required check passes (Type check, Lint, Format, Knip, Opengrep, Build platform, all 16 Playwright platform shards, etc.); only fork-PR jobs `skipping`. None red, none pending.
- **`bunx vitest --run --project server convex/products/assert_unique_product_name.test.ts` → 5/5 passed.**
- **`bunx vitest --run --config vitest.ui.config.ts` on the 5 touched UI specs → 20/20 passed.**

### What is solid
- **Secret edit affordance** — correct. Basic secrets are stored as two independent documents (no type metadata); editing a `*_USERNAME`/`*_PASSWORD` row patches only that document by exact name (`upsertProjectSecretInternal`), so there is no pair corruption. Name is fixed/disabled (can't orphan env-var keys), value re-encrypted server-side, reveal correctly omitted (`listProjectSecrets` returns metadata only). Empty value is blocked by `required` + server `value.length === 0` rejection.
- **Agent duplicate-display-name warning** — non-blocking, case-insensitive/trimmed, safe under loading (`undefined` agents → empty set), correctly excludes nothing at create. Matches the "warn, cosmetic" intent (slug stays unique).
- **Product dialog error handling** — `extractErrorCode` duck-typing is the right reuse; `errorToast: false` correctly prevents double toasts; the `wasOpen` reset-on-open-only pattern is safe (single render site in `product-row-actions.tsx`, conditionally mounted so it unmounts between opens) and is what preserves the field error across the optimistic update/rollback.
- **TXT accept hint** — correct: `MIME_TYPES.PLAIN` is in `DOCUMENT_UPLOAD_ACCEPT` and server-accepted (`convex/documents/mutations.ts`). i18n parity (en/de/fr) complete and enforced by the parity test.

### BLOCKING

**1. Product-name uniqueness is enforced on only one of the product write paths.** The PR frames a product's name as a hard data invariant ("a product's name IS its identity… a duplicate name is a genuine ambiguity") and enforces it via `assertUniqueProductName` — but `grep` confirms that function is referenced **only** in the two interactive mutation wrappers (`convex/products/mutations.ts:30` createProduct, `convex/products/update_product.ts:32` updateProduct). Every other write path calls the helpers directly and bypasses the check, so the invariant does not actually hold:
- `internal_mutations.ts` `ingestProduct` → `ProductsHelpers.createProduct` (no check) — called by REST `POST /api/v1/products` (`rest_api.ts:51`), the agent `product_write` tool (`agent_tools/products/product_write_tool.ts:62`), and the workflow product action (`workflow_engine/action_defs/product/product_action.ts:174`).
- `internal_mutations.ts` `updateProducts` → `ProductsHelpers.updateProducts` (no check) — called by REST `PATCH /api/v1/products/:id` (`rest_api.ts:105`), the agent tool (`product_write_tool.ts:80`), and the workflow action (`product_action.ts:244`).
- Bulk import (`bulk_create_products.ts:52` → `createProductWithTranslations` directly) — the PR explicitly acknowledges this exemption; the three paths above are **not** acknowledged and appear overlooked.

This is a half-implemented invariant: an agent or workflow (first-class flows on this platform) or any REST client can create/rename a product to a name that already exists, reintroducing exactly the duplicate-name ambiguity the issue is about, while the UI now forbids it — an inconsistency between branches.

**Required fix:** enforce the rule on every product create/rename path. Cleanest is to move `assertUniqueProductName` down into the shared write helpers (`createProduct`/`createProductWithTranslations` and the `updateProducts` batch helper, gated on `name` being set) so all entry points inherit it; alternatively add the call inside the `ingestProduct` and `updateProducts` internal mutations. If any path (e.g. bulk import) is intentionally exempt, state why in code/PR. Add coverage for at least one non-UI path (e.g. `ingestProduct`/`updateProducts` rejecting a duplicate).

### Non-blocking (worth addressing, not gating)
- **Agent rename path** has no duplicate warning (only the create dialog does — `agents/$agentId` general tab edits `displayName` with no warning). Defensible given the warning is cosmetic and intent says "create dialog," but inconsistent UX; consider a follow-up.
- **Secret edit `isDirty`** allows enabling Save when only the description changed with an empty value; the empty value is then server-rejected (no data loss) but the UX is slightly confusing. Minor.
- **Magic string** `'DUPLICATE_PRODUCT_NAME'` is repeated in 3 places (backend throw + both dialogs); a shared constant would be cleaner. Nit.

---
**NOT READY — CHANGES REQUIRED.** Single blocker: extend the product-name uniqueness check (`assertUniqueProductName`) to the non-UI write paths — REST `POST`/`PATCH /api/v1/products`, the agent `product_write` tool, and the workflow product action (via the shared `createProduct`/`updateProducts` helpers or the `ingestProduct`/`updateProducts` internal mutations) — or explicitly justify each exemption, and add a test on one non-UI path. Everything else (secret edit, agent warning, product dialogs, TXT hint, tests, i18n, CI) meets the bar.

larryro:
## Desk Review — #1999 (data-quality + affordance gaps)

**Verdict: READY TO MERGE.** Reviewed in two rounds (breadth review across the 5 bar dimensions, then independent re-verification of every candidate finding). No blocking issue survived confirmation.

### CI & tests (observed)
- **CI: green** — all 40+ checks pass (`gh pr checks`): Build platform/convex/db/proxy/sandbox*, Lint, Format, Knip, Opengrep, Migrations, Analyze (JS/TS + Python), and all 16 Playwright platform shards + docs/web.
- **Full suite: green** — `bunx vitest --run --project server --project client` → **897 files / 9116 tests passed (exit 0)**.
- **Touched files: green** — the 7 new/changed test files → **30/30 passed**.

### What was verified
**1. Product name uniqueness (backend).** `assertUniqueProductName` (trimmed, case-folded, `by_organizationId` index confirmed in schema) is wired into every write path — `createProduct`, `updateProduct` (excludes self), `updateProducts` (batch: rejects renaming >1 row to one name, else self-excluded assert), and `ingestProduct` (REST POST / agent `product_write` / workflow action). The check runs *before* the insert, so no partial write; Convex mutation transactionality closes the TOCTOU window. Bulk import intentionally retains its own path.

**2. `ingestProduct` enforcement — verified NOT a regression.** Tracing all non-UI callers (REST `rest_api.ts`, `product_write_tool.ts`, `product_action.ts`): each uses an explicit create-vs-update operation; none is a by-name idempotent upsert, and there is no documented "POST twice with same name succeeds" contract. This change brings the non-UI paths to parity with the UI's pre-existing uniqueness rule — a data-quality fix, not a broken contract. The full suite passing confirms no existing caller/test relied on duplicates being allowed.

**3. Product UI (create + edit).** `extractErrorCode` duck-types `data.code` (robust against Vite chunk-split `ConvexError` copies) → create shows a `duplicateName` toast, edit sets an inline `name` field error. `errorToast:false` on `useCreateProduct` prevents a double-toast (`onError` still fires). The edit dialog's reset-only-on-open (`wasOpen` ref) correctly preserves the typed name + field error across the optimistic-update→rollback prop churn, and still re-seeds on the next open transition.

**4. Agent duplicate display-name warning.** Non-blocking client warning built from `useListAgents` + `resolveAgentLocale` (loading/undefined guarded; trimmed/case-folded). Slug uniqueness remains enforced server-side (`file_actions.ts` `DUPLICATE_NAME`), so warn-only is the right call for a cosmetic clash. Continue stays enabled (tested).

**5. Project secret Edit.** No reveal added — stored plaintext is never returned to the client; Edit re-collects a value and re-encrypts via the same `setProjectSecret` upsert, name fixed/disabled so refs can't be orphaned. Confirmed **no data-loss path**: an empty value cannot blank a secret — the value input is `required` *and* `setProjectSecret` throws `SECRET_VALUE_INVALID` on `value.length === 0` (actions.ts:58). Edit state cleanup (`resetForm` clears `editingName` on save and on close) verified.

**6. i18n + TXT hint.** All added keys present in en/de/fr with matching paths (parity test would pass). TXT genuinely accepted by the uploader (`DOCUMENT_UPLOAD_ALLOWED_TYPES`/`_EXTENSIONS` include `text/plain`/`txt`), so the dropzone hint now matches reality.

### Non-blocking observations (optional, for a follow-up — do not gate merge)
- Backend uniqueness is unit-tested on the `assertUniqueProductName` helper (all branches) and on `updateProducts`, plus UI-level handling; the thin mutation-wiring for `createProduct` / single `updateProduct` has no dedicated integration test. The delegated logic is fully covered, so this is a coverage nicety, not a gap in behavior.
- Editing a `basic`-type secret fragment (`*_USERNAME` / `*_PASSWORD`) re-saves that single value as `custom` under its fixed name — correct and lossless, but the row gives no hint it was half of a pair. Minor UX only.

Code, conventions, naming, and comments are clean and fit the repo (snake_case Convex helpers, `helpers.ts` re-export, idiomatic `wasOpen` ref). Knip clean.

PR #2216, opened 2026-06-27
fix(platform): cover pass-2 pii scrub rows in audit chain verifier (#1843)

### LINKED ISSUE #1843: Bug: GDPR erasure pass-2 PII scrub guarantees a false "hash chain broken" alarm on signed deployments
## Summary

On any deployment with `TALE_AUDIT_SIGNING_KEY` configured, **every GDPR erasure deterministically breaks audit-chain verification** and raises the critical "hash chain broken at log &lt;id&gt;" tamper alert. The erasure pipeline's pass-2 PII scrub blanks admin-authored rows _about_ the erased subject, but the verifier's scrub-trust window is keyed on the row's `actorId` — which for pass-2 rows is the **admin**, not the subject the signed checkpoint attests. The verifier therefore recomputes the hash over the blanked body and reports a mismatch.

No tampering is involved at any step; this is the product's own Art 17 feature tripping its own tamper detection.

## Mechanism

`scrubSubjectAuditLogs` (`services/platform/convex/audit_logs/internal_mutations.ts:301-449`) runs two passes:

- **Pass 1** (`:327-367`) — rows where the subject is the **actor** (`by_organizationId_and_actorId`). Blanks `actorEmail`/`actorRole`/`ipAddress`/`userAgent`/`previousState`/`newState`/`metadata`/`actorEmailHash`/`actorIpHash`, sets `piiScrubbed: true`.
- **Pass 2** (`:369-403`) — rows where the subject is the **resource** (`by_org_resourceType_resourceId`, `resourceType: 'user'`). Blanks `resourceName`/`previousState`/`newState`/`errorMessage`/`metadata`, sets `piiScrubbed: true`. The actor on these rows is the admin who performed the action — deliberately left intact (`:369-374`).

Both passes leave the original `integrityHash` in place and rely on a signed `pii_scrub` checkpoint (`:420-441`, `scrubbedSubjectId = args.userId`) to tell the verifier the divergence is intentional.

The verifier builds its trust windows keyed by `scrubbedSubjectId` and looks them up **by the row's `actorId` only** (`services/platform/convex/audit_logs/verify_integrity.ts:332-343`, `:403-408`):

```ts
const windows = subjectScrubWindows.get(actorId);
if (windows && windows.some((w) => entry.timestamp <= w.maxTimestamp)) {
  isScrubbed = true;
} else if (!hasSigningKey) { ... }
```

For a pass-2 row, `actorId` (admin) ≠ `scrubbedSubjectId` (subject) → no window matches. On a signed deployment the `!hasSigningKey` legacy branch is unreachable, so `isScrubbed` stays false, the hash is recomputed over the blanked body (`:478-495`), mismatches the stored `integrityHash`, and `verifyAuditChain` returns `valid: false` with `firstBrokenAt` = the scrubbed row.

All five blanked pass-2 fields are hash-covered (`buildAuditRecordHashInput`, `audit_logs/helpers.ts:149-187`); only `piiScrubbedAt` (in `EXCLUDED_FIELDS`) and `piiScrubbed` (destructured out by the verifier) are hash-neutral, so the mismatch is guaranteed whenever pass 2 blanked anything.

## Trigger — guaranteed, not probabilistic

The erasure flow itself always creates at least one pass-2 row: `requestGdprErasure` writes a `gdpr_erasure_requested` audit row with `actorId = <admin>`, `resourceType: 'user'`, `resourceId = <subject>`, with `resourceName` and `newState` populated (`governance/erasure.ts:573-592`). The erasure processor then calls `scrubSubjectAuditLogs` (`erasure.ts:2140-2147`), whose pass 2 selects that very row and blanks it. The next 02:00 UTC integrity cron reports the break. Other admin-authored rows about the subject (member invites, password resets via `users/set_member_password.ts`, role changes) widen the blast radius.

The whole scrub (both passes + checkpoint) commits in one mutation, so this is a permanent steady state, not a transient.

## Test gap

No test covers `pii_scrub` + verify interaction: `integrity_check.test.ts` has zero scrub coverage, and `append_only.test.ts` only exercises sequential appends and out-of-band tampering. That is why this shipped.

## Suggested fix

Make the verifier's coverage test mirror pass-2's **selection criteria**: a row is covered by a signed scrub window when

- `actorId === scrubbedSubjectId` (pass 1), **or**
- `resourceType === 'user' && resourceId === scrubbedSubjectId` (pass 2),

in both cases with `entry.timestamp <= window.maxTimestamp`. The signed checkpoint already binds `scrubbedSubjectId` into the HMAC (signature v2), so this widens trust only to rows the scrub actually attested — no new forgery surface beyond what pass 1 already accepts. Add a regression test: erase a user on a signed deployment, then assert `verifyAuditChain` returns `valid: true`.

Adjacent gap noticed while reading (can be split out if preferred): `scrubSubjectAuditLogs` performs no legal-hold check — every other per-table eraser re-checks holds mid-flight (`countOrSkip`, `erasure.ts:980-994`), but a custodian hold placed between erasure scheduling and the processor run still gets its audit-log PII blanked.

## How this was found

Investigating a production "Audit log integrity check failed" notification. Related umbrella: #1803. Sibling issues from the same investigation: #1842, #1844, #1845, #1846.

larryro:
## Summary

Fixes a false **"hash chain broken"** tamper alarm that fires deterministically after every GDPR erasure on any deployment with `TALE_AUDIT_SIGNING_KEY` configured.

`scrubSubjectAuditLogs` scrubs PII in two passes:
- **pass 1** — rows where the erased subject is the **actor** (`actorId === subject`)
- **pass 2** — admin-authored rows **about** the subject (`resourceType === 'user' && resourceId === subject`); here the actor is the admin, not the subject.

Both passes blank hash-covered fields and rely on a signed `pii_scrub` checkpoint (keyed by `scrubbedSubjectId`) to tell the verifier the divergence is intentional. But `verifyAuditChain` looked its scrub-trust window up by `actorId` **only**. For a pass-2 row, `actorId` (admin) ≠ `scrubbedSubjectId` (subject), so no window matched; on a signed deployment the `!hasSigningKey` legacy fallback is unreachable, so the verifier recomputed the hash over the blanked body and reported a mismatch. The erasure flow always writes at least one pass-2 row (`gdpr_erasure_requested`), so the break was guaranteed.

## Fix

`services/platform/convex/audit_logs/verify_integrity.ts` — the per-row scrub-coverage test now mirrors the scrub's own two-pass **selection criteria**. A scrubbed row is trusted when a signed `pii_scrub` window exists for either:
- `actorId === scrubbedSubjectId` (pass 1), **or**
- `resourceType === 'user' && resourceId === scrubbedSubjectId` (pass 2),

in both cases with `entry.timestamp <= window.maxTimestamp`. Coverage is probed under both candidate subject keys. Because the v2 signed checkpoint binds `scrubbedSubjectId` into the HMAC, this only widens trust to rows the scrub actually attested — no new forgery surface beyond what pass 1 already accepted. The `!hasSigningKey` legacy/unsigned fallback (and its `unsignedScrubCount`) is preserved unchanged.

## Test

New `services/platform/convex/audit_logs/verify_scrub.test.ts` (the prior test gap): on a signed deployment, scrub a subject with both actor-pass and resource-pass rows and assert `verifyAuditChain` returns `valid: true` with `unsignedScrubCount === 0`; plus the minimal real-world trigger (a single admin-authored pass-2 row). Both tests fail without the verifier change and pass with it.

- `bunx vitest --run --project server audit_logs/` → 32 passed
- `bunx tsc --noEmit` → clean

Closes #1843

### DIFF
--- a/services/platform/convex/audit_logs/verify_integrity.ts
+++ b/services/platform/convex/audit_logs/verify_integrity.ts
@@ -321,9 +321,11 @@ export async function verifyAuditChain(
   let isFirstEntry = args.fromTimestamp === undefined;
 
   // Build per-subject scrub windows from SIGNED pii_scrub checkpoints
-  // only. A row carrying `actorId === X` is allowed to skip hash
-  // recompute only when there is a signed checkpoint covering
-  // `(X, timestamp ≤ maxDeletedTimestamp)`. Without this scoping
+  // only. A row whose actor OR (user-)resource is subject X is allowed
+  // to skip hash recompute only when there is a signed checkpoint
+  // covering `(X, timestamp ≤ maxDeletedTimestamp)` — see the per-row
+  // coverage test below, which mirrors the scrub's two selection
+  // passes. Without this scoping
   // (the prior membership-only Set), a single pii_scrub checkpoint
   // for subject X authorized hash skip on every row that subject ever
   // touched, including future rows the checkpoint never attested.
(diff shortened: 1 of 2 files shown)


larryro:
## Desk review — PR #2216 (fixes #1843)

**Verdict: READY TO MERGE.**

Reviewed the whole implementation (not just the diff) over two rounds: 5 breadth subagents across the review-bar dimensions, then independent confirmation of every load-bearing claim.

### CI
All required checks green: **46 pass, 5 skipping** (fork-PR/conditional jobs only), **zero failing or pending**.

### Tests (run locally, once)
`bunx vitest --run --project server audit_logs/` → **32 passed (6 files)**.

I also **proved the new tests are real regression guards**: temporarily reverting the pass-2 `resourceId` probe in `verify_integrity.ts` (local scratch only, never pushed) makes both new tests fail with `expected false to be true` — the exact false "hash chain broken" alarm from #1843. With the fix they pass.

### What's correct
- **Root cause fixed.** `verify_integrity.ts` now builds `candidateSubjectKeys = [actorId, (resourceId if resourceType==='user')]` and trusts a scrubbed row when any candidate has a signed `pii_scrub` window with `entry.timestamp <= w.maxTimestamp`. This mirrors `scrubSubjectAuditLogs`'s two passes **exactly** (pass 1: `actorId === subject`; pass 2: `resourceType==='user' && resourceId === subject`), so every pass-2 row the erasure flow writes — `gdpr_erasure_requested` and all other admin-authored rows about the subject — is now covered.
- **No new tamper surface.** Trust still requires `piiScrubbed === true` AND a *signed* checkpoint window keyed by `scrubbedSubjectId`; the v2 HMAC binds `scrubbedSubjectId` + `subtype`, so the `resourceId` probe only trusts rows the scrub actually attested — no widening beyond what pass 1 already accepted. The `!hasSigningKey` unsigned-legacy fallback remains strictly gated, and unsigned checkpoints are still excluded from `subjectScrubWindows` (downgrade-attack defense intact).
- **Clean & behavior-preserving.** The two former `!hasSigningKey` branches consolidate into one with identical `unsignedScrubCount` accounting; typeof guards on `actorId`/`resourceId` are sound; `.some()` on an empty key list correctly falls through. Style/comment density matches the file.

### Non-blocking note (for a follow-up, not this PR)
`scrubSubjectAuditLogs` still performs no legal-hold check, unlike other per-table erasers (`countOrSkip`, `governance/erasure.ts`). The issue explicitly flagged this as an adjacent gap that "can be split out," and it is out of scope for this false-alarm fix. Recommend filing a separate issue to track it.

PR #2217, opened 2026-06-27
fix(platform): preserve audit chain contiguity around custodian-hold rows (#1844)

### LINKED ISSUE #1844: Bug: audit retention around custodian-hold rows leaves non-contiguous chain — guaranteed false tamper alarm
## Summary

When audit-log retention runs while a **custodian legal hold** is active, `deleteOldLogs` skips the held user's rows per-row while hard-deleting the unprotected rows around them. The surviving held rows become non-contiguous islands in the hash chain. The verifier re-anchors via checkpoint **only at the chain head** and then demands strict `previousHash` contiguity, so the next integrity run deterministically reports "hash chain broken at log &lt;id&gt;" — a false tamper alarm caused by two compliance features (retention + preservation holds) composing.

## Mechanism

`deleteOldLogs` (`services/platform/convex/audit_logs/internal_mutations.ts:144-280`) walks the org chain ascending and deletes rows older than the cutoff, but `continue`s past protected rows (`:185-192`):

```ts
if (protectedUserIds.has(log.actorId) || targetIsHeldUser) {
  continue;
}
```

`protectedUserIds` is populated from active user-custodian holds by the retention dispatcher (`governance/retention_cleanup.ts:519-524`); only an **org-wide** hold skips audit cleanup entirely (`:490-495`). No survivor re-chaining exists — the only post-delete chain writes are the checkpoint fields, never patches to surviving rows.

The verifier (`services/platform/convex/audit_logs/verify_integrity.ts`) consults checkpoints **only while `isFirstEntry`** (`:321`, `:429-459`). From the second row onward it requires `entryPreviousHash === previousExpectedHash` (`:461-476`) with no mid-chain re-anchor and no `chainSuccessor` use.

## Worked example

Org chain `R1→R2→R3→R4→R5→R6...` (each `R_{i+1}.previousHash = H(R_i)`); user U (actor of R2 and R4) is placed on a custodian hold; R1–R5 age past the retention cutoff.

1. The 04:00 UTC retention cron deletes R1, R3, R5 and skips R2, R4. Checkpoint: `lastDeletedHash = H(R5)`, `firstRetainedPreviousHash = R2.previousHash = H(R1)`.
2. The 02:00 UTC integrity cron walks: head R2 anchors successfully via `firstRetainedPreviousHash` and verifies.
3. Next survivor R4 carries `previousHash = H(R3)` — a deleted row — failing the contiguity check at `:461`. → `valid: false`, `firstBrokenAt = R4`, critical tamper alert.

With exactly one protected survivor the break lands instead on the first post-cut live row (its `previousHash = lastDeletedHash`, which the verifier never consults mid-chain). Either way the failure is deterministic and permanent — each `deleteOldLogs` batch is one serializable mutation, so the non-contiguous survivor set is the committed steady state.

## Impact

- Any org combining `auditLogRetentionDays` with at least one custodian hold whose user has audit activity older than the cutoff gets a permanent daily false tamper alarm.
- As with the other false-positive paths: the verifier stops at the first failure, so everything after the break is never verified — real tampering downstream is masked.
- `protectedUserIds` appears in zero test files; the interaction is untested.

## Suggested fix

Preferred: **stop the deletion batch at the first protected row** — delete only the contiguous unprotected prefix and resume after the hold is released. Chain contiguity is preserved by construction; the cost is that unprotected rows _behind_ a held row outlive their retention window for the duration of the hold, which is defensible (FRCP 37(e) preservation generally trumps routine retention, and the org-wide-hold branch already makes that trade).

Alternative: keep per-row skips but write a checkpoint per contiguous deleted gap and teach the verifier to re-anchor mid-chain across signed gap checkpoints. More surface, more signature semantics to get right; only worth it if long-lived holds + aggressive retention make prefix-stopping unacceptable.

Either way, add a regression test: hold one user, run retention across their interleaved rows, assert `verifyAuditChain` returns `valid: true`.

## How this was found

Investigating a production "Audit log integrity check failed" notification. Related umbrella: #1803. Sibling issues from the same investigation: #1842, #1843, #1845, #1846.

larryro:
## Summary

When audit-log retention runs while a **custodian legal hold** is active, `deleteOldLogs` skipped each protected row *per-row* while hard-deleting the unprotected rows around it. The surviving held rows became non-contiguous islands in the SHA-256 hash chain. The verifier re-anchors via checkpoint **only at the chain head** and then demands strict `previousHash` contiguity, so the next integrity run deterministically reported "hash chain broken at log <id>" — a permanent daily **false tamper alarm** for any org combining `auditLogRetentionDays` with at least one custodian hold.

## Fix

Stop the deletion sweep at the **first protected row** (`continue` → `break` in `audit_logs/internal_mutations.ts`). Only the contiguous unprotected **prefix** is deleted, so every surviving row's `previousHash` still references either another retained row or the checkpoint's `lastDeletedHash` — chain contiguity is preserved by construction.

The cost — unprotected rows *behind* a held row outlive their retention window until the hold is released — is the same trade the org-wide-hold branch in `retention_cleanup.ts` already makes, and is defensible under FRCP 37(e) (preservation duty trumps routine retention). The sweep resumes deleting that backlog automatically once the hold is lifted.

## Tests

New `audit_logs/retention_holds.test.ts` (3 cases):
- held **actor** rows interleaved with an unprotected actor → chain stays contiguous, `verifyAuditChain` returns `valid: true`;
- rows that merely **target** a held user (`resourceType: 'user'`) are preserved too;
- once the hold is released the backlog is deleted and the chain remains valid.

All three fail against the old per-row `continue` and pass with the fix. `tsc --noEmit` and `oxlint --type-aware` are clean.

Closes #1844

### DIFF
--- a/services/platform/convex/audit_logs/internal_mutations.ts
+++ b/services/platform/convex/audit_logs/internal_mutations.ts
@@ -176,19 +176,35 @@ export const deleteOldLogs = internalMutation({
         break;
       }
 
-      // Skip rows whose actor OR user-typed resource is on a custodian
-      // hold. Cross-cutting iteration: we keep paging until we find a
-      // deletable row OR exhaust this batch. (`hasMore` below uses
-      // `deletedCount` rather than scanned count, so a window full of
-      // protected rows correctly reports `hasMore: false` when nothing
-      // was deletable in this slice — the cleanup will revisit later.)
+      // Custodian-hold preservation (FRCP 37(e)): a held user's audit
+      // rows — as the `actorId` OR as a user-typed resource target —
+      // must survive past their retention age. STOP the deletion sweep
+      // at the first protected row instead of skipping past it.
+      //
+      // Deleting only the contiguous unprotected PREFIX keeps the
+      // surviving hash chain contiguous by construction: every retained
+      // row's `previousHash` still references either another retained
+      // row or the checkpoint's `lastDeletedHash`. The earlier per-row
+      // `continue` stranded protected rows as non-contiguous islands —
+      // a survivor's `previousHash` pointed at a now-deleted neighbour —
+      // so the next integrity run deterministically reported a false
+      // "hash chain broken" tamper alarm (#1844). The verifier
+      // re-anchors via checkpoint only at the chain head and demands
+      // strict contiguity thereafter, with no mid-chain re-anchor.
+      //
+      // The cost — unprotected rows BEHIND a held row outlive their
+      // retention window until the hold is released — is the same trade
+      // the org-wide-hold branch in retention_cleanup.ts already makes,
+      // and is defensible: preservation duty trumps routine retention.
+      // The sweep resumes deleting that backlog automatically once the
+      // hold is lifted and the row is no longer protected.
       const targetIsHeldUser =
         (log.resourceType === 'user' ||
           log.resourceType === 'userMembership') &&
         log.resourceId !== undefined &&
         protectedUserIds.has(log.resourceId);
       if (protectedUserIds.has(log.actorId) || targetIsHeldUser) {
-        continue;
+        break;
       }
 
       // Track the chain head we're deleting so the checkpoint can
--- a/services/platform/convex/audit_logs/retention_holds.test.ts
+++ b/services/platform/convex/audit_logs/retention_holds.test.ts
@@ -1,0 +1,181 @@
+import { convexTest, type TestConvex } from 'convex-test';
+import { describe, expect, it } from 'vitest';
+
+import { internal } from '../_generated/api';
+import type { Id } from '../_generated/dataModel';
+import schema from '../schema';
+
+// convex-test module map keyed relative to the convex/ root. This file lives
+// at convex/audit_logs/, so resolve glob keys against that base (mirrors
+// append_only.test.ts).
+const TEST_DIR_FROM_CONVEX_ROOT = 'audit_logs';
+function toConvexRootKey(globKey: string): string {
+  const stack: string[] = [];
+  for (const part of `${TEST_DIR_FROM_CONVEX_ROOT}/${globKey}`.split('/')) {
+    if (part === '' || part === '.') continue;
+    if (part === '..') stack.pop();
+    else stack.push(part);
+  }
+  return stack.join('/');
+}
+const rawModules = import.meta.glob('../**/*.*s');
+const modules: Record<string, () => Promise<unknown>> = {};
+for (const [key, loader] of Object.entries(rawModules)) {
+  modules[toConvexRootKey(key)] = loader;
+}
+
+const ORG = 'org_audit_retention_holds';
+const FAR_FUTURE = 4_102_444_800_000; // 2100-01-01: every seeded row is "old".
+
+type T = TestConvex<typeof schema>;
+
+interface SeedOpts {
+  actorId?: string;
+  resourceType?: string;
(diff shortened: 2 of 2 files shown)


larryro:
## Desk review — PR #2217 (fix for #1844)

**Verdict: READY TO MERGE.**

Reviewed the whole implementation (not just the diff) over two rounds: 5 breadth subagents across the correctness/robustness/elegance/tests/boundaries dimensions, then a confirmation round re-verifying every candidate finding. Setup was a single clone + checkout + one test run.

### What the fix does
`deleteOldLogs` changes the custodian-hold branch from `continue;` to `break;`, so the retention sweep stops at the first protected row and deletes only the **contiguous unprotected prefix**. The surviving suffix is contiguous by construction: every retained row's `previousHash` references either another retained row or the checkpoint's `lastDeletedHash`. This is exactly the issue's *preferred* fix.

### Evidence
- **CI: green.** `gh pr checks` — all 49 checks pass (Type check, Lint, Unit, Build convex/platform, Playwright 1–16, etc.). The only non-pass jobs are `skipping` fork-PR/Trivy variants that don't run on same-repo PRs.
- **Tests (run locally on the branch, `bunx vitest --run --project server`):**
  - `audit_logs/retention_holds.test.ts` — **3/3 pass**
  - full `audit_logs/` suite — **33/33 pass**
  - `governance/retention*` (the caller) — **61/61 pass**
- **Regression test is genuine:** reverting `break`→`continue` makes all 3 new tests fail (e.g. `deletedCount` 2 vs expected 1), so they lock in the exact #1844 behavior.

### Correctness — confirmed sound across all branches
- Verifier head-anchor (`verify_integrity.ts:434-456`) matches the surviving head via **either** `firstRetainedPreviousHash` **or** `lastDeletedHash`, so both the terminal-batch (`hasMore=false`) and mid-batch (`hasMore=true`, `firstRetainedPreviousHash` undefined) cases re-anchor correctly.
- Protected-row-at-head ⇒ `deletedCount=0` ⇒ no checkpoint written (correct). Two consecutive runs under the same persistent hold ⇒ second run breaks immediately, writes no checkpoint, chain still verifies. Hold release ⇒ backlog + previously-held rows delete, a newer checkpoint is written, and `anchorCandidates` (sorted desc by `createdAt`) correctly picks the most-recent checkpoint — no stale-checkpoint mismatch.
- `audit_log.retention_deleted` bookkeeping row stays correctly chained at the tail.

### Robustness / resolution
- Dispatcher (`governance/retention_cleanup.ts:507-530`) loops while `result.hasMore`; with `break`, `hasMore=false` once a held row is hit, so the loop exits cleanly — no spin, no starvation. Unprotected rows *behind* a held row wait for hold release (the documented FRCP 37(e) trade, matching the org-wide-hold branch). Only one functional caller of `deleteOldLogs`.

### Non-blocking nit (does not gate merge)
- `internal_mutations.ts:151` — the `protectedUserIds` JSDoc still reads *"Skipped at the per-row level"*, which now contradicts the break-at-first-protected behavior. The accurate large comment block just below it (lines ~179-200) already describes the new behavior; recommend tightening line 151 to match. Optional test polish: `retention_holds.test.ts:176` uses `toBeGreaterThan(0)` where `toBe(2)` would be exact, and `userMembership`-target / protected-row-at-head branches are untested (both behave identically to covered paths).

None of these affect correctness, and the core bar is met.

**READY TO MERGE.**

PR #2219, opened 2026-06-27
fix(convex): validate product translation language against locale whitelist (#2008)

### LINKED ISSUE #2008: Bug: Product translation language parameter accepts arbitrary strings — no locale whitelist validation
## Area

Products — Convex backend (`services/platform/convex/products/`)

## Steps to reproduce

Call `upsertProductTranslation` (or pass a translation object via `createProduct`/`updateProduct`) with a `language` value of any arbitrary string, e.g. `\"xyz\"`, `\"not-a-locale\"`, or a 10 000-character string:

\`\`\`ts
await convex.mutation(api.products.mutations.upsertProductTranslation, {
  productId: someId,
  language: "not-a-real-locale",
  name: "Test",
});
\`\`\`

The mutation succeeds and persists the garbage locale code to the `translations` array.

## Expected

`language` should be validated against the platform's supported locale set.  The project already defines `SUPPORTED_LOCALES = ['en', 'de', 'fr']` in `lib/shared/constants/locales.ts` and a broader 42-locale `SUPPORTED_DATASET_LOCALES` in `lib/shared/constants/dataset-locales.ts`.  The server should reject values outside the accepted set with a `ConvexError`.

## Actual

`language: v.string()` — no enum constraint, no length cap, no whitelist check anywhere in the handler body.

## Evidence

- `services/platform/convex/products/mutations.ts:94` — `language: v.string()`
- `services/platform/convex/products/validators.ts:19` — `productTranslationValidator` also uses `language: v.string()`
- `services/platform/convex/products/upsert_product_translation.ts:6` — interface typed `language: string`
- `services/platform/convex/products/schema.ts:27` — schema stores `language: v.string()` with no constraint
- `services/platform/lib/shared/constants/locales.ts:11` — `SUPPORTED_LOCALES` defined but never referenced in product validators

## Notes

- This is not covered by #1991 (which covers task title / API-key / project-secret / team name) or #1999 (duplicate display names).
- The `createProduct` and `updateProduct` mutations also accept `translations` arrays with the same unvalidated `language` field via `productTranslationValidator`.
- Fix: replace `v.string()` with a `v.union(v.literal('en'), v.literal('de'), v.literal('fr'))` (or broader IETF tag set) in `productTranslationValidator` and the `upsertProductTranslation` args, and throw a `ConvexError` in the handler for rejected values.

---
*Filed by an automated validation-gap audit of `services/platform/convex/products/` (commit `7b961c445`).*

larryro:
## Summary

`upsertProductTranslation` (and the `translations` arrays accepted by `createProduct`/`updateProduct`) typed `language` as `v.string()` with no whitelist, so arbitrary garbage locale codes — `"xyz"`, `"not-a-locale"`, even a 10 000-character string — were persisted to the `translations` array unchecked.

This validates `language` against the platform's supported locale set (`SUPPORTED_LOCALES = ['en', 'de', 'fr']`).

## Changes

- **Argument boundary (strict):** added `productLocaleValidator` (a `v.union` of `SUPPORTED_LOCALES` literals) and a strict `productTranslationInputValidator`; the `upsertProductTranslation`, `createProduct`, and `updateProduct` mutation args now use them, so unsupported locales are rejected at the Convex validation boundary.
- **Handler (defence-in-depth):** added `isSupportedLocale` to `lib/shared/constants/locales.ts` and `assertSupportedProductLocale` (throws a structured `ConvexError` with code `invalid_locale`) in the three write helpers, guarding direct/internal callers that bypass the arg validators.
- **Read compatibility:** the read/doc validator (`productTranslationValidator`) and the schema keep `language: v.string()` so reads of any pre-existing, pre-whitelist documents never fail return validation.

## Tests

Added `convex/products/locale_validation.test.ts` — covers the whitelist guard (accepts `en`/`de`/`fr`, rejects `xyz`, `not-a-locale`, `EN`, `en-US`, empty, 10k-char) and that each write helper throws a `ConvexError` before touching the database while supported locales still persist.

- `bunx vitest --run --project server convex/products/` → 19 passed
- `bunx tsc --noEmit` → 0 errors
- `bunx oxlint --type-aware` on changed files → 0 warnings / 0 errors

Closes #2008


larryro:
## Desk Review — PR #2219 (fixes #2008)

**Verdict: READY TO MERGE.**

Reviewed the whole implementation on branch `tale/xs74xf8chd627jrrk2e0sq3zq989ev34` (two-round, fan-out review), not just the diff.

### CI & tests
- **CI: GREEN.** All checks `pass` (Type check, Lint, Knip, Unit, Build convex, all 16 Playwright shards, etc.). Only `skipping` entries are fork-only/Trivy jobs — no failures, no pending.
- **Tests run locally:** `cd services/platform && bunx vitest --run --project server convex/products/` → **19 passed (2 files)**, including all 13 cases in the new `locale_validation.test.ts`. `bunx tsc --noEmit` → **0 errors**.

### What the fix does (correct & complete)
`language` is now validated against `SUPPORTED_LOCALES = ['en','de','fr']` at two layers:
1. **Convex arg boundary** — `productLocaleValidator` (`v.union` of literals) + `productTranslationInputValidator` on `createProduct` / `updateProduct` / `upsertProductTranslation` args.
2. **Handler defence-in-depth** — `assertSupportedProductLocale` (throws `ConvexError{code:'invalid_locale'}`) in all three write helpers, before any DB read/write (fail-fast, verified).

I exhaustively enumerated every path that can persist `translations[].language`: the three mutations are covered; `bulkCreateProducts`, REST `ingestProduct`/`updateProducts`, the `product_write` agent tool, and the Shopify/workflow product actions **structurally do not accept translations**, so there is no bypass. The read/storage validator and schema deliberately keep `language: v.string()` so reads of legacy, pre-whitelist documents don't fail return-validation — a correct, intentional asymmetry. The 10k-char / `"xyz"` / `""` / `"EN"` / `"en-US"` vectors from the issue are all rejected before any write.

### Non-blocking suggestions (do not block merge)
1. **Test the arg-boundary directly.** Current tests exercise the helper-level asserts (which *are* the production code path, so the rejection guarantee is genuinely covered). A `convex-test` case driving `api.products.mutations.upsertProductTranslation` with `language:'xyz'` would additionally lock down the `productLocaleValidator` outer layer. `convex-test` is already set up and used widely in this repo.
2. **`v.union` fragility.** `productLocaleValidator = v.union(...SUPPORTED_LOCALES.map(v.literal))` is fine with 3 locales, but `v.union` requires ≥2 members — if the list is ever reduced to one, this throws at module load. Worth a guard/comment.
3. **Structured error reachability.** On the public mutation path the arg validator rejects first with a generic `ArgumentValidationError`, so the nicer `ConvexError{code:'invalid_locale'}` is only observed by internal/helper callers. Fine as defence-in-depth; just be aware clients won't see the structured code on the normal path.
4. Minor: `updateProduct` has no happy-path "persist" test (upsert/create do); "Product not found" still throws plain `Error` while the new guard throws `ConvexError` (mixed contract within the same handlers).

None of the above block merge. The change is correct, minimal, well-documented, fully tested at the behavior level, and CI is green.

**READY TO MERGE.**

PR #2223, opened 2026-06-27
fix: read governance policy server state directly instead of mirroring in useState (#2023)

### LINKED ISSUE #2023: Bug: RunCodePolicyRoute and PersonalizationPolicyToggle mirror server state in useState via useEffect — stale data flash on load
## Area

Governance settings — Run Code Policy page and Personalization Policy toggles.

### RunCodePolicyRoute (primary violation)
1. Navigate to **Settings → Governance → Run Code Policy**.
2. Open DevTools → React DevTools or add a `console.log` to the `RunCodePolicyRoute` component.
3. Observe: on initial load, the text-area inputs and mode radio briefly display the `useState` initial defaults (`'denylist'`, empty strings) before the `useEffect` fires to populate from server data.

### PersonalizationPolicyToggle (secondary violation)
1. Navigate to **Settings → Governance → Personalization**.
2. Open DevTools and add a log before the return. On first render after `isLoading → false`, the `enabled` state is `false` (the `useState` default) even if the server-stored value is `true`, until the `useEffect` runs.

## Expected

Server-derived values should be available on the same render in which they become known. Components should read server state directly (or derive it via `useMemo`) rather than copying it into `useState` via `useEffect`.

### RunCodePolicyRoute
`services/platform/app/routes/dashboard/$id/settings/governance/run-code-policy.tsx:135-147`

```tsx
const savedDraft = useMemo<PolicyDraft>(() => { ... }, [policy]);  // server-derived

const [defaultMode, setDefaultMode] = useState<PolicyMode>('denylist');  // stale default
const [pythonAllowText, setPythonAllowText] = useState('');
const [pythonDenyText, setPythonDenyText] = useState('');
const [nodeAllowText, setNodeAllowText] = useState('');
const [nodeDenyText, setNodeDenyText] = useState('');

useEffect(() => {
  setDefaultMode(savedDraft.defaultMode);
  setPythonAllowText(listToString(savedDraft.pythonAllow));
  // ... etc
}, [savedDraft]);
```

Because `useEffect` fires after the browser paints, once `isLoading` becomes `false` the form briefly renders with `'denylist'` / empty strings before the effect fires to set the real values. The `<Skeletonize>` wrapping masks this during the loading phase but not during the transition flash.

### PersonalizationPolicyToggle
`services/platform/app/features/settings/governance/components/personalization-policy-editor.tsx:51-56`

```tsx
const savedEnabled = useMemo(() => readEnabled(policy?.config), [policy]);
const [enabled, setEnabled] = useState(false);  // starts as false
useEffect(() => {
  setEnabled(savedEnabled);
}, [savedEnabled]);
```

## Evidence

- `services/platform/app/routes/dashboard/$id/settings/governance/run-code-policy.tsx:135-147`
- `services/platform/app/features/settings/governance/components/personalization-policy-editor.tsx:51-56`

## Notes

The project's data-loading doc explicitly states: "never useState-mirror server state". The fix for RunCodePolicyRoute is to use `savedDraft` directly in the form (controlled inputs read from `savedDraft`, local edits tracked as a separate delta or overriding memo). For the PersonalizationPolicyToggle, since the local state IS needed for optimistic updates, the fix is to initialise `useState` from `savedEnabled` (e.g. `useState(() => readEnabled(policy?.config))`) and drop the `useEffect` — accepting that external-admin-changes only reflect after a component remount (which is the correct behaviour for an optimistic toggle).

larryro:
## Summary

Fixes the stale-data flash on the **Run Code Policy** page and the **Personalization** toggles, where server-derived values were copied into `useState` via `useEffect`. Because effects run after paint, once `isLoading` flipped to `false` the form briefly rendered the `useState` defaults (`'denylist'` / empty strings / `false`) before the effect populated the real values.

The fix follows the project's React doctrine ("never mirror server state in `useState`", "resist the `useEffect` reflex") — server state is now read directly each render, so the real values are present on the same render in which they become known.

### `run-code-policy.tsx` (`RunCodePolicyRoute`)
- Removed the five `useState` mirrors + the `useEffect` that synced them from `savedDraft`.
- Local edits are now tracked as a `Partial<FormState>` overriding the server-derived values (`edits.field ?? savedValue`), so an unedited field always reflects server state — no flash. Edits are cleared on a successful save so the form re-reads the freshly-saved server value.
- Dropped the now-unused `useEffect` import.

### `personalization-policy-editor.tsx` (`PersonalizationPolicyToggle`)
- Replaced the `useState` + `useEffect` mirror with a derived value: `enabled = pending ?? savedEnabled`, where `savedEnabled` is read directly from the server config and `pending` holds only the optimistic value while a save is in flight (cleared in `finally`).
- This avoids the stale `false` flash while preserving optimistic toggling, and lets later server changes flow through once the save settles. (The lazy-`useState` approach is unsuitable here because the component mounts while `isLoading` is `true`, so a one-time initializer would capture `false` and never update.)
- Dropped the now-unused `useEffect` / `useMemo` imports.

## Tests
- Added a regression test asserting the toggle reflects the server-stored `enabled` value on the first loaded render.
- `oxlint --type-aware`, `tsc --noEmit`, and the UI vitest suite for the editor all pass locally.

Closes #2023

### DIFF
--- a/services/platform/app/features/settings/governance/components/personalization-policy-editor.test.tsx
+++ b/services/platform/app/features/settings/governance/components/personalization-policy-editor.test.tsx
@@ -64,6 +64,18 @@ describe('PersonalizationPolicyEditor', () => {
       render(<PersonalizationPolicyEditor organizationId="org-1" />);
       expect(screen.queryByRole('status')).not.toBeInTheDocument();
     });
+
+    it('reflects the server-stored enabled value on the first loaded render', () => {
+      // Regression for #2023: the toggle used to copy server state into
+      // `useState` via `useEffect`, so it rendered the `false` default first.
+      // It now derives `enabled` from server state directly, so a stored
+      // `enabled: true` is checked immediately — no stale `false` flash.
+      setLoaded();
+      render(<PersonalizationPolicyEditor organizationId="org-1" />);
+      for (const toggle of screen.getAllByRole('switch')) {
+        expect(toggle).toHaveAttribute('aria-checked', 'true');
+      }
+    });
   });
 
   describe('loading state (skeletonized)', () => {
--- a/services/platform/app/features/settings/governance/components/personalization-policy-editor.test.tsx
+++ b/services/platform/app/features/settings/governance/components/personalization-policy-editor.test.tsx
@@ -1,6 +1,6 @@
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { render, screen } from '@/tests/utils/render';
+import { act, render, screen } from '@/tests/utils/render';
 
 import { PersonalizationPolicyEditor } from './personalization-policy-editor';
 
@@ -15,8 +15,35 @@ vi.mock('@/app/hooks/use-ability', () => ({
   }),
 }));
 
+// Controllable save mutation. `mode: 'defer'` hands the test the in-flight
+// promise's resolve/reject so it can assert the optimistic override is visible
+// while the save is pending, then settle it; `'resolve'`/`'reject'` complete
+// synchronously. Hoisted so the (hoisted) `vi.mock` factory can read it.
+const { mutation } = vi.hoisted(() => {
+  const m = {
+    mode: 'resolve' as 'resolve' | 'reject' | 'defer',
+    resolvePending: null as null | (() => void),
+    rejectPending: null as null | ((reason?: unknown) => void),
+    mutateAsync: vi.fn(),
+  };
+  m.mutateAsync = vi.fn(() => {
+    if (m.mode === 'defer') {
+      return new Promise<void>((resolve, reject) => {
+        m.resolvePending = resolve;
+        m.rejectPending = reject;
+      });
+    }
+    if (m.mode === 'reject') return Promise.reject(new Error('save failed'));
+    return Promise.resolve();
+  });
+  return { mutation: m };
+});
+
 vi.mock('../hooks/mutations', () => ({
-  useUpsertGovernancePolicy: () => ({ mutateAsync: vi.fn(), isPending: false }),
+  useUpsertGovernancePolicy: () => ({
+    mutateAsync: mutation.mutateAsync,
+    isPending: false,
+  }),
 }));
 
 // Mutable, hoisted so the mock factory can read it (vi.mock is hoisted above
@@ -44,6 +71,13 @@ function setLoading() {
   state.config = undefined;
 }
 
+beforeEach(() => {
+  mutation.mode = 'resolve';
+  mutation.resolvePending = null;
+  mutation.rejectPending = null;
+  mutation.mutateAsync.mockClear();
+});
+
 describe('PersonalizationPolicyEditor', () => {
   describe('loaded state', () => {
     it('renders the real switches (in the a11y tree)', () => {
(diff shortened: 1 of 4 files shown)


larryro:
## Desk Review — PR #2223 (fixes #2023: stale-data-flash from `useState`/`useEffect` mirroring)

**Verdict: NOT READY — CHANGES REQUIRED.** The production fix is correct and resolves the reported issue for both components, and CI is fully green — but the test coverage does not actually guard the regression, and the issue's *primary* component has none.

### What's good (confirmed)
- **Both components now read server state directly**, eliminating the load flash:
  - `run-code-policy.tsx` — `edits: Partial<FormState>` overlay merged with server-derived `savedDraft` via `useMemo` (`edits.field ?? serverValue`). `??` correctly preserves empty-string and `false` edits (only `undefined` falls back).
  - `personalization-policy-editor.tsx` — `enabled = pending ?? savedEnabled`, `savedEnabled` read each render; `false`/optimistic values handled correctly.
- Unused `useEffect`/`useMemo` imports cleaned up. Error path keeps user edits (run-code) and reverts the optimistic toggle (personalization) — both reasonable. Permission gating (`cannotManage`) and validation (`parseList`, `runCodePolicyConfigSchema.safeParse`, `readEnabled`) preserved.
- The implementer's `pending ?? savedEnabled` choice is arguably better than the issue's suggested `useState(initializer)` — external/admin changes flow through without a remount.

### Blocking findings

**1. The added regression test does not guard the regression (it passes against the buggy code).**
`personalization-policy-editor.test.tsx:68–78` ("reflects the server-stored enabled value on the first loaded render") asserts `aria-checked='true'` after `render()`. React Testing Library's `render()` wraps in `act()`, which flushes `useEffect` synchronously — so the **old** `useState(false)` + `useEffect(setEnabled(savedEnabled))` implementation also shows `aria-checked='true'` by the time the assertion runs. **Verified empirically:** I temporarily restored the old implementation and ran this exact test — it still passed. The test therefore provides false confidence; reintroducing the bug would not turn it red.
- Fix: replace it with a test that actually distinguishes the two implementations. jsdom can't capture the real pre-paint flash, so test the observable contract the refactor guarantees instead — e.g. an optimistic `pending` value overriding the server value during an in-flight save and then dropping back to the server value, and/or a changed server `config` flowing through on re-render. The test must fail against `useState(false)+useEffect`.

**2. RunCodePolicyRoute — the issue's *primary* violation — has zero test coverage of its fix.**
No test file exists for `run-code-policy.tsx` (`find . -name '*run-code*'` and a `RunCodePolicyRoute` grep over `*.test.*` both return nothing). The new `edits`-overlay logic is non-trivial and untested. Add tests covering:
   - form fields render the server values on the first loaded render (no `denylist`/empty defaults);
   - a user edit overrides the server value for that field while other fields still mirror server state;
   - a **successful** save clears edits (`setEdits({})`) so the form re-reads freshly-saved server state;
   - a **failed** save preserves the user's edits.

### Non-blocking notes (not required for merge)
- `personalization-policy-editor.tsx:76` `catch { toast(...) }` does not `console.error` the error, unlike `run-code-policy.tsx:207`. Not an empty catch and oxlint passes, but logging would match the sibling and the repo's "log or re-throw" guidance.
- Lost-edit-during-save (textareas stay editable during `await`; success `setEdits({})` discards edits typed mid-save) exists, but the old `useEffect`-on-`savedDraft` had the identical behavior — **pre-existing, not introduced here**.
- A brief revert-to-old-value flicker is theoretically possible if the reactive `getPolicy` query hasn't propagated by the time `setEdits({})`/`setPending(null)` runs after a successful action; this matches the established codebase pattern documented in `mutations.ts`, so not a defect of this PR.

### Evidence
- CI: `gh pr checks` → 46 pass, 5 conditional-skip, 0 fail/0 pending (green).
- Tests run locally: `bunx vitest --run --config vitest.ui.config.ts app/features/settings/governance/components/personalization-policy-editor.test.tsx` → **7 passed**. Tautology confirmation: same command with the old implementation restored, `-t "reflects the server-stored enabled value"` → **1 passed** (should have failed).

**NOT READY — CHANGES REQUIRED** (see blocking findings 1 and 2).

larryro:
## Desk Review — PR #2223 (fixes #2023)

**Verdict: ✅ READY TO MERGE.**

Reviewed by a two-round method (5 parallel review subagents across correctness / robustness / elegance / tests / boundaries, then independent confirmation of every load-bearing claim).

### CI
All required checks **green** at the reviewed head: Type check, Lint, Format, Knip, **Unit**, **UI**, Browser, Storybook, CodeQL, Opengrep, all 16 E2E Playwright shards, Build/Smoke/Validate. The only non-SUCCESS entries are `SKIPPED` fork-PR jobs and a `NEUTRAL` Trivy run — none blocking.

### Tests
- `personalization-policy-editor.test.tsx` — ran locally (`bunx vitest --run --config vitest.ui.config.ts --pool=forks`): **9/9 pass**.
- `run-code-policy.test.tsx` — green in CI's **UI** check (4 tests). It could not be run in my sandbox due to an unrelated `import { z } from 'zod'` returning `undefined` under bun's vitest worker pool (reproduced with a one-line zod import test → environment artifact, not a code defect).
- Coverage is a genuine **behavioral regression guard** for #2023, not a value-only check that the old code would also pass:
  - personalization: "shows the optimistic value while saving, then re-derives from server state" — the final assertion (`aria-checked='false'` after a successful save with unchanged server config) would FAIL against the old `useState(false)+useEffect` mirror.
  - run-code-policy: "clears edits on a successful save so the form re-reads server state" asserts the freshly-saved canonical server value (not the typed edit, not the stale pre-edit value).
  - Both failure paths covered (revert / preserve edits) plus loading/skeleton states.

### Findings (none blocking)
- **Correctness** — verified across all branches. `enabled = pending ?? savedEnabled` correctly distinguishes a legitimate `false` from `null` (nullish coalescing). `edits.X ?? listToString(savedDraft.X)` correctly preserves an empty-string edit (cleared textarea) since `??` is nullish, not falsy. Unedited fields stay reactive to server changes. Save success drops the overlay and re-reads server; save error preserves the user's edits.
- **Boundaries / resolution** — the `useEffect` server→`useState` mirror is fully removed from **both** components named in the issue (only comment references remain). Server config is validated via `runCodePolicyConfigSchema.safeParse(... ?? {})` with an `EMPTY_DRAFT` fallback; `readEnabled` type-guards `unknown`. The chosen approach (read-direct + transient overlay) is strictly better than the issue's suggested `useState(()=>...)` — it also reflects external server changes without a remount.
- **Robustness** — error paths log via `console.error` + toast (satisfies the repo's no-empty-catch rule). The personalization Switch and the run-code Save button are disabled while `upsertMutation.isPending`, preventing overlapping saves. Note (non-blocking, pre-existing): `setEdits({})`/`toast` after `await` could fire post-unmount, but `origin/main` already had `useEffect`+`await` with the same property — not a regression.
- **Elegance** — minimal, focused change; removed imports cleaned up (Knip green); `FormState`/`edits` overlay and `pending` naming are clear; comments explain *why*. No dead code or duplication.

### Non-blocking nits (optional)
1. `handleSave` is re-created on every keystroke (depends on `liveDraft` → `form` → `edits`). Harmless; could memoize against a ref if desired.
2. Tester-widget logic, the mode radio standalone, and the `cannotManage` disabled path aren't directly unit-tested — all outside #2023's scope.

**READY TO MERGE.**

PR #2226, opened 2026-06-27
fix(platform): show conversation status-tab count badges, drop dead ?priority= param

### LINKED ISSUE #1982: Improvement: Conversations — status-tab count badges are prefetched but never displayed; dead ?priority= route param
### What should be improved?
Two small conversations-inbox refinements.

### Current behavior
1. Per-status approximate counts are prefetched on the conversations route but never rendered — the status tabs show no count badge (the prefetch is currently load-bearing only). 2. The route accepts a `?priority=` search param but there is no UI to set or clear it.

### Proposed change
1. Pass the prefetched counts into the tab's trailing badge slot (or drop the now-unused prefetch). 2. Either expose a priority filter control or remove the dead `?priority=` param.

### Affected service
Platform (web app)

UX refinement; relates to #1966.

---

larryro:
## Summary

Two small conversations-inbox refinements from #1982:

1. **Status-tab count badges now render.** The per-status approximate counts were already prefetched on the conversations route loader but never displayed (the prefetch was load-bearing only). `ConversationsNavigation` now subscribes to those same counts and renders each as a trailing badge on its status tab (open / closed / spam / archived), shown only when the count is > 0 and capped as `20+` (matching the backend `DEFAULT_COUNT_CAP`). Because the badge query shares the subscription primed by the route loader's prefetch, the badges paint without a refetch on first nav.
2. **Dead `?priority=` route param removed.** The conversations status route accepted a `?priority=` search param, but no UI ever set or cleared it. Removed it from the route's search schema and its plumbing into `useListConversationsPaginated`. (The backend query still supports priority filtering via the hook should a UI control be added later.)

## Testing
- `tsc --noEmit` — clean
- `oxlint --type-aware` on changed files — 0 warnings / 0 errors
- UI tests: `app/features/conversations` — 7 files / 32 tests pass (includes the navigation a11y/axe test, updated to stub the new count query)

Closes #1982

### DIFF
--- a/services/platform/app/features/conversations/components/conversations-navigation.test.tsx
+++ b/services/platform/app/features/conversations/components/conversations-navigation.test.tsx
@@ -28,6 +28,10 @@ vi.mock('@/app/hooks/use-ability', () => ({
   }),
 }));
 
+vi.mock('../hooks/queries', () => ({
+  useApproxConversationCountByStatus: () => ({ data: 3 }),
+}));
+
 describe('ConversationsNavigation', () => {
   describe('accessibility', () => {
     it('passes axe audit', async () => {
--- a/services/platform/app/features/conversations/components/conversations-navigation.tsx
+++ b/services/platform/app/features/conversations/components/conversations-navigation.tsx
@@ -4,7 +4,10 @@ import {
   TabNavigation,
   type TabNavigationItem,
 } from '@/app/components/ui/navigation/tab-navigation';
+import { DEFAULT_COUNT_CAP } from '@/convex/lib/helpers/count_items_in_org';
 import { useT } from '@/lib/i18n/client';
+
+import { useApproxConversationCountByStatus } from '../hooks/queries';
 
 interface ConversationsNavigationProps {
   organizationId: string;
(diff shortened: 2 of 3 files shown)


larryro:
## Desk review — #1982 (status-tab count badges + dead `?priority=` param)

**Verdict: READY TO MERGE.**

Reviewed the full implementation on branch `tale/xs7afk3471v4jbjdp93njz16eh89fn7r` (not just the diff), across correctness, robustness, elegance, tests, and boundaries, with an independent confirmation pass on every point I'd block on.

### CI & tests
- **CI: green.** All required checks pass (Type check, Lint, Format, Knip, Unit, UI, Storybook, Build platform/convex, Migrations, all 16 Playwright platform shards, Playwright web/docs, Opengrep, Trivy, Scan platform, Smoke). The only non-pass rows are fork-PR / docs-container / web-container jobs marked `skipping` by design — no failures, nothing pending.
- **Tests run locally:** `bunx vitest --run --config vitest.ui.config.ts app/features/conversations` → **7 files / 32 tests passed**.

### What I verified
- **Badges wired to the prefetched data (part 1).** `ConversationsNavigation` subscribes via `useApproxConversationCountByStatus` (`useConvexQuery` → `convexQuery(api.conversations.queries.approxCountConversationsByStatus, {organizationId, status})`), the *same* query+args the parent route loader (`conversations.tsx`) prefetches — identical react-query key, so badges paint from the primed cache without a refetch. The prefetch is no longer "load-bearing only".
- **Cap boundary correct.** Backend `approxCountConversationsByStatus` breaks at `count >= DEFAULT_COUNT_CAP` (20), so it returns at most 20; `CountBadge` renders `20+` at `>= 20`. Semantics match an approximate "at least N" count.
- **Zero/undefined handling correct.** Badge renders only when `count !== undefined && count > 0` — no `0` badge, no flash of `undefined`/`NaN` while loading or on query error (graceful degradation).
- **`?priority=` fully removed from the route (part 2).** Dropped from `searchSchema`, from `Route.useSearch()` destructuring, and from the `useListConversationsPaginated` call in `$status.tsx`. Old bookmarked `?priority=…` URLs degrade safely — zod strips unknown keys rather than throwing.
- **No new subscription leak.** `$status.tsx` already subscribed to the same 4 counts (pre-existing); the nav's 4 calls dedupe to the same cache keys.
- **Boundaries.** Count query is `queryWithRLS` (org-scoped, no cross-org leakage) and `status` is constrained to the 4 literals; both parts of the issue are genuinely resolved.

### Non-blocking notes (optional cleanup — do not block merge)
1. `ListConversationsPaginatedArgs.priority?: string` (`app/features/conversations/hooks/queries.ts:26`) now has no frontend caller. Leaving it is defensible (backend still supports priority filtering and the PR documents this as intentional for a future UI control); removing it would be marginally cleaner.
2. The added test (`conversations-navigation.test.tsx`) is a11y-only and does not assert badge value / `20+` cap / hide-on-zero. This matches the repo's own convention for sibling display components (`conversation-list-toolbar`, `activate-conversations-empty-state`, `dot-icon` are all a11y-only), so it is not a blocker — but a couple of behavioral assertions on the badge branches would be a nice-to-have.

**READY TO MERGE.**

Issue #1425, opened 2026-04-11
“Continue” button should remain disabled until required fields are completed in Create Automation modal


mdabdulkarimqa456-code:
Description:

In the Create automation modal, the Name field is mandatory, but the Continue button is not properly tied to the required field validation state.

From a usability and form-validation perspective, the Continue button should remain disabled until all required fields are completed with valid input. Otherwise, users may try to proceed before the form is in a valid state.

Steps to Reproduce:
Navigate to the Automation creation flow
Open the Create automation modal
Leave the required Name field empty
Observe the validation message
Check the state of the Continue button
Expected Result:
The Continue button should remain disabled until the required Name field is filled
Once the mandatory field is completed, the button should become enabled
Actual Result:
The form shows “Name is required”, but the button behavior is not properly aligned with the validation flow
Users can reach a confusing state where required field validation is shown, but the action button logic is not clearly enforced

PR #2229, opened 2026-06-28
fix: Notifications — "Mark all as read" stays enabled while personal-stream mutation is in-flight (#2019)

### LINKED ISSUE #2019: Bug: Notifications — "Mark all as read" button stays enabled while personal-stream mutation is in-flight
## Area

Notifications panel — \"Mark all as read\" button disabled state logic

## Steps to reproduce

1. Open the notifications bell / panel with at least one unread personal notification (review request, task assignment, etc.).
2. Click **Mark all as read** rapidly twice in quick succession.

## Expected

The button is disabled for the entire duration of **both** mutations it fires — org stream (`markAllRead`) and personal stream (`markAllNotificationsRead`) — preventing double submission.

## Actual

The button's `disabled` prop is only bound to `markAllRead.isPending` (the org-notification mutation). The personal-stream mutation (`markAllMyRead`) runs concurrently and has no guard. If the org mutation resolves first, the button becomes re-enabled while `markAllMyRead` is still in-flight, allowing a second `markAllNotificationsRead` call to be submitted.

## Evidence

`services/platform/app/features/notifications/components/notification-list-panel.tsx:201`
```tsx
<Button
  variant="ghost"
  disabled={markAllRead.isPending}   // ← only guards the org stream
  onClick={handleMarkAllRead}
>
```

`handleMarkAllRead` (line 130–131) fires both mutations:
```ts
void markAllRead.mutateAsync({ organizationId });      // org stream
void markAllMyRead.mutateAsync({ organizationId });    // personal stream — NOT guarded
```

The fix is:
```tsx
disabled={markAllRead.isPending || markAllMyRead.isPending}
```

## Notes

- Both `useMarkAllNotificationsRead` and `useMarkAllMyNotificationsRead` come from separate `useConvexMutation` instances, so their `isPending` states are independent.
- The personal `markAllNotificationsRead` mutation (`convex/collab/notifications.ts:120`) is idempotent (it only patches `read: false` rows), so the duplicate call produces no data corruption — the concern is unnecessary network traffic and the Convex write quota cost of a redundant mutation execution.

larryro:
Closes #2019

## Problem

The "Mark all as read" button in the notifications panel fires two independent mutations — the org stream (`markAllRead`) and the personal stream (`markAllMyRead`) — but its `disabled` prop was only bound to `markAllRead.isPending`. Since the two `useConvexMutation` instances have independent `isPending` states, if the org mutation resolves first the button re-enables while the personal-stream mutation is still in-flight, allowing a redundant `markAllNotificationsRead` submission.

## Fix

Bind the button `disabled` state to both pending flags so it stays disabled for the full duration of both mutations:

```tsx
disabled={markAllRead.isPending || markAllMyRead.isPending}
```

## Verification

- `tsc --noEmit` (platform) — passes
- `oxlint --type-aware` on the changed file — 0 warnings, 0 errors


larryro:
## Desk review — #2019

**Verdict: NOT READY — CHANGES REQUIRED** (one blocking item — a missing regression test. The code change itself is correct and should be kept exactly as written.)

### The fix is correct
`disabled={markAllRead.isPending || markAllMyRead.isPending}` is the right fix. I verified across all dimensions:
- **Correctness:** `isPending` (TanStack `useMutation`) is true while the promise is in-flight, false once settled (success *or* error), so the button stays disabled for the full duration of both mutations and re-enables only when both settle. Neither mutation nor the unread-count queries use `optimisticUpdate`, so `unreadCount` doesn't drop mid-flight and the button can't unmount and defeat the guard. Fixes the exact reported race.
- **Robustness:** On a rejection, that mutation's `isPending` returns to false, so the button is never left permanently disabled; errors are logged + toasted by `use-convex-mutation`, not swallowed.
- **Boundaries:** Both underlying `markAll…` mutations are idempotent; pure UI change, no auth surface; this is the only "Mark all as read" button — no other site has the bug.
- **Elegance:** Minimal, idiomatic, consistent with sibling guards.

### Blocking: no regression test
This PR adds no test. By the repo's own bar that's a gap:
- AGENTS.md DoD for a bug fix: *"Reproduce → write the failing regression test → minimal targeted fix → confirm green"* / *"Tests carry the change."*
- Direct precedent in-repo: `app/features/automations/components/automation-create-dialog.test.tsx` — *"Regression test for #1425: Continue must stay disabled until the required name field is filled"* — the same disabled-while-pending pattern (`expect(submit).toBeDisabled()`).
- The notification panel has no component test, and `tests/manual/notifications.md` (untouched here) only covers the basic F4 happy path, not the in-flight double-submit race. So reverting the `|| markAllMyRead.isPending` clause would be caught by nothing.

**Required change**
Add `services/platform/app/features/notifications/components/notification-list-panel.test.tsx` (the `client`/jsdom vitest project auto-includes `app/features/**/*.test.tsx`), following the `automation-create-dialog.test.tsx` (#1425) pattern. Mock the two mark-all hooks — `useMarkAllNotificationsRead` from `../hooks/mutations` (→ `markAllRead`) and from `@/app/features/inbox/hooks/mutations` (→ `markAllMyRead`) — plus the unread-count/list query hooks so the panel renders with `unreadCount > 0`. Assert the regression: with `markAllRead.isPending === false` and `markAllMyRead.isPending === true`, the button is `toBeDisabled()` (fails on old code, passes on the fix); and with both `false`, the button is enabled. Confirm green in the `client` project.

CI is fully green; `tsc --noEmit` and `oxlint --type-aware` pass locally. The only thing standing between this and merge is the regression test above.

larryro:

Reviewed the fix for #2019 ("Mark all as read" stays enabled while the personal-stream mutation is in-flight) via a two-round review (5 parallel breadth subagents across correctness / robustness / elegance / tests / boundaries, then independent confirmation).

**The fix** (`notification-list-panel.tsx:201`): `disabled={markAllRead.isPending}` → `disabled={markAllRead.isPending || markAllMyRead.isPending}`. `handleMarkAllRead` fires *both* the org-stream (`markAllRead`) and personal-stream (`markAllMyRead`) mutations (lines 130–131), but the button previously only gated on the org stream — leaving the exact double-submit window the issue reports. The OR-gate now keeps the button disabled until **both** independent `useMutation` instances settle. `isPending` resets on settle (success *or* error), so no stuck-disabled state. Errors are not swallowed — `use-convex-mutation` logs + toasts on failure.

**Completeness:** grep confirms this is the *only* UI surface with a "Mark all as read" button firing both streams (NotificationBell just wraps this panel). No other surface carries the same bug. No new boundary/security surface.

**Tests:** New `notification-list-panel.test.tsx` (3 cases) covers the exact bug window (org settled, personal in-flight), the inverse, and both-settled. I verified it is a *genuine* regression test by temporarily reverting the one-line fix locally — the personal-stream case fails (`Received element is not disabled`) against the old code and passes against the fix. Mocks map correctly to the org vs. personal hook module paths.

**Test run:** `bunx vitest --run --config vitest.ui.config.ts app/features/notifications/components/notification-list-panel.test.tsx` → **3 passed**.

**CI:** All required checks green (Unit, UI, Type check, Lint, Format, Knip, Build platform, all 16 Playwright platform shards, etc.); only fork-PR/container jobs skipped. `mergeable: MERGEABLE`.

Minor, non-blocking: the suite focuses on the disabled-gating (no test asserts the click fires both mutations, or that the button is hidden when `unreadCount === 0`) — these are pre-existing feature-coverage gaps, not regressions, and don't block this fix.

**Verdict: READY TO MERGE.**

PR #2232, opened 2026-06-28
fix: scan all pages for list search/filter, not just loaded ones (#2054)

### LINKED ISSUE #2054: Bug: List search/filter only scans already-loaded pages (client-side filter + infinite-scroll) — knowledge-entries, websites, conversations
**Area / route**
Three list views run search/filter purely client-side over the already-loaded paginated buffer in `displayMode:'infiniteScroll'` (the `useListPage` default). The backend queries have no search arg, the eager full-load drain lives only in the `'pagination'` branch (`use-list-page.ts:348-356`), and when the filter narrows loaded rows to zero, DataTable suppresses the infinite-scroll sentinel (`data-table.tsx:1053`, gated on `data.length>0`) so no further pages load — the user sees empty 'no results' though a match exists deeper. Products/Customers/Vendors avoid this via `displayMode:'pagination'`. Each surface preserved below.

**Sites**

- **[68] Knowledge-entries list search only matches already-loaded pages (client-side filter + infinite-scroll, no server-side search), silently missing entries on unloaded pages**
  - Evidence: knowledge-entries-table.tsx:71-86 (useListPage with search fields ['topic','content'] and no displayMode); app/hooks/use-list-page.ts:171 (displayMode default 'infiniteScroll'), :208-236 (search filters rawData only), :348-355 (eager loadMore guarded by displayMode==='pagination'); products-table.tsx:114-134 (displayMode:'pagination')
- **[75] Websites list search only filters already-loaded pages (infinite-scroll), silently missing matching websites on un-loaded pages**
  - Evidence: websites-table.tsx:168-187 (useListPage with no displayMode + search.fields client-side) vs customers-table.tsx:190 / vendors-table.tsx:160 (displayMode:'pagination'); use-list-page.ts:208-236 filters rawData only; eager-load-all path is gated on displayMode==='pagination' at use-list-page.ts:348-356.
- **[80] Conversations search and read-status filter only scan loaded pages — a match beyond the first 30 falsely shows "No conversations in this tab"**
  - Evidence: conversations.tsx:110-130 (filterByTextSearch + readFilter over paginatedResult.results); list_conversations_paginated.ts:59-97 (no search arg); conversations-list.tsx:406-417 empty branch has no sentinel vs 449-452 sentinel only when CanLoadMore/LoadingMore

**Notes**
This is about server-side vs client-side SCOPE, not field coverage — distinct from #1996 (conversations omits the customer-name field) and #1992 (clearing search stuck via ?search=); all can be true simultaneously. Lowest-effort parity fix per table: add `displayMode:'pagination'` (eager full-load). More scalable: add a server-side search arg to each paginated query (wire the existing unused `searchWebsites` helper for [75]; address #1996's field set in the conversations [80] server search). A client-only fix must keep auto-loading pages while a search is active even when the filtered view is empty.

---

larryro:
## Summary

Three list views (knowledge-entries, websites, conversations) ran search/filter purely **client-side over the already-loaded paginated buffer** while using infinite-scroll. The backend queries had no search arg and the eager full-load drain only existed in the `'pagination'` display mode, so:

- matches on un-loaded pages were silently missed, and
- when a filter narrowed the loaded rows to zero, the infinite-scroll sentinel was suppressed (gated on `data.length > 0` / the empty branch had no sentinel), so no further pages loaded — the user saw a false "no results" / "No conversations" even though a match existed deeper.

This implements the client-only fix the issue calls for: **keep auto-loading backend pages while a search/filter is active, even when the filtered view is empty**, so the client-side filter scans the full dataset. Infinite-scroll UX is preserved when no filter is active.

## Changes

- **`app/hooks/use-list-page.ts`** (knowledge-entries [68] + websites [75]): detect an active client-side search / managed field filter (`hasActiveClientFilter`) and, in infinite-scroll mode, eagerly drain remaining backend pages (`loadMore`) until exhausted — mirroring the existing `'pagination'`-mode drain.
- **`app/components/ui/data-table/data-table.tsx`**: when a filter empties the loaded buffer but pages are still draining (`hasActiveFilters && infiniteScroll.hasMore`), show the loading/skeleton state instead of "no results", so a deep match isn't prematurely reported as empty.
- **`app/features/conversations/components/conversations.tsx` + `conversations-list.tsx`** (conversations [80]): thread an `isFiltering` flag; drive pagination directly while a search/read-filter is active (the empty-buffer branch has no sentinel), and show a spinner instead of "No conversations" while still draining.

## Tests

- Extended `app/hooks/use-list-page.test.ts` with cases covering: no drain while idle, drain on active search, drain on active managed filter, and no drain when the backend is exhausted.
- `bunx vitest --run` (use-list-page + conversations + data-table UI suites): **all pass**.
- `bunx tsc --noEmit`: clean. `bunx oxlint --type-aware`: clean.

Closes #2054

### DIFF
--- a/services/platform/app/components/ui/data-table/data-table.tsx
+++ b/services/platform/app/components/ui/data-table/data-table.tsx
@@ -444,6 +444,10 @@ export function DataTable<TData, TValue = unknown>({
     if (!isDataLoading) {
       // Has data
       if (data.length > 0) return 'data';
+      // A filter narrowed the loaded rows to zero, but infinite-scroll is still
+      // draining backend pages — show loading, not "no results", so a match on
+      // an un-loaded page isn't prematurely reported as empty (#2054).
+      if (hasActiveFilters && infiniteScroll?.hasMore) return 'skeleton';
       // Has filters
       if (hasActiveFilters) return 'filtered-empty';
       // Has empty state
@@ -469,6 +473,7 @@ export function DataTable<TData, TValue = unknown>({
     data.length,
     emptyState,
     hasActiveFilters,
+    infiniteScroll?.hasMore,
   ]);
 
   const isSkeleton =
--- a/services/platform/app/features/conversations/components/conversations-list.tsx
+++ b/services/platform/app/features/conversations/components/conversations-list.tsx
@@ -103,6 +103,12 @@ interface ConversationsListProps {
   loadMore?: (numItems: number) => void;
   /** Number of placeholder rows to render while `conversations` is undefined. */
   skeletonRows?: number;
+  /**
+   * Whether a client-side search/read-filter is active. When true, remaining
+   * backend pages are eagerly drained so the filter scans the full dataset
+   * rather than only already-loaded pages (#2054).
+   */
+  isFiltering?: boolean;
 }
 
 const priorityConfig = {
@@ -342,6 +348,7 @@ export function ConversationsList({
   paginationStatus,
   loadMore,
   skeletonRows = 12,
+  isFiltering = false,
 }: ConversationsListProps) {
   const { formatDateSmart } = useFormatDate();
   const { t } = useT('conversations');
@@ -383,6 +390,16 @@ export function ConversationsList({
     return () => observer.disconnect();
   }, [handleIntersect]);
 
+  // While a search/read-filter is active the client-side filter only matches
+  // already-loaded pages, and a filter that empties the loaded buffer hides the
+  // scroll sentinel above — so drive pagination directly here, draining pages
+  // until a match loads or the list is exhausted (#2054).
+  useEffect(() => {
+    if (isFiltering && paginationStatus === 'CanLoadMore' && loadMore) {
+      loadMore(PAGE_SIZE);
+    }
+  }, [isFiltering, paginationStatus, loadMore]);
+
   // One real tree, always. Loading, empty and populated states all render
   // inside the SAME <Skeletonize> wrapper so the base layout never shifts
   // between states (mirrors the DataTable body-state pattern):
@@ -394,6 +411,14 @@ export function ConversationsList({
   const isEmpty = conversations !== undefined && conversations.length === 0;
   const placeholderCount = Math.min(skeletonRows, 12);
 
+  // A filter narrowed the loaded buffer to zero, but more pages are still
+  // draining — show loading, not "no conversations", so a match on an un-loaded
+  // page isn't prematurely reported as empty (#2054).
+  const isSearchingMore =
+    isEmpty &&
+    isFiltering &&
+    (paginationStatus === 'CanLoadMore' || paginationStatus === 'LoadingMore');
+
   return (
     <Skeletonize
       loading={isLoading}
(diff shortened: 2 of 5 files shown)


larryro:
## Desk Review — #2054 (PR #2232): scan all pages for list search/filter

**Verdict: READY TO MERGE.**

### CI & tests (observed)
- `gh pr checks` → **46 pass, 5 skipping** (skips are fork-PR-only jobs: Validate images, Trivy, Smoke test (fork PR)). No failures, nothing pending. CI is **green** — incl. Type check, Lint, Build platform, Storybook, UI, and all 16 Playwright (platform) shards.
- Ran locally (bun + vitest, `client` jsdom project):
  - `bunx vitest --run --project client app/hooks/use-list-page.test.ts` → **11 passed**
  - `app/components/ui/data-table/data-table.test.tsx` → **17 passed**
  - `app/features/conversations/` → **32 passed (7 files)**
  - full `--project client` run → exit 0
- `bunx tsc --noEmit` → clean (no errors on the changed files).

### What the fix does (verified correct across all three surfaces)
- **`use-list-page.ts`** — new `hasActiveClientFilter` + a render-body drain (`loadMore(pageSize*3)` while `CanLoadMore`) in the infinite-scroll branch. This fixes **knowledge-entries [68]** and **websites [75]** automatically (both call `useListPage` with managed `search.fields` and no `displayMode`). The drain mirrors the long-standing pagination-mode pattern (`loadMore` during render, gated on status), and converges: each load flips status to `LoadingMore` → page arrives → `CanLoadMore` → next load → `Exhausted`. Gated so it does nothing when idle (covered by the "does not drain while idle" test).
- **`data-table.tsx`** — `if (hasActiveFilters && infiniteScroll?.hasMore) return 'skeleton'` prevents the false "no results" flash while pages still drain; once `Exhausted`, `hasMore` is false and it correctly falls through to `filtered-empty`. Dep array updated with `infiniteScroll?.hasMore`. Correct.
- **conversations [80]** (`conversations.tsx` + `conversations-list.tsx`) — `isFiltering = Boolean(searchQuery||initialSearch) || readFilter!=='all'` threaded to the list; a `useEffect` drives `loadMore` directly while filtering (the empty-buffer branch has no scroll sentinel), and `isSearchingMore` shows a spinner instead of "No conversations" until `Exhausted`. Full-dataset scan is guaranteed (drain runs regardless of filtered-buffer size). Converges; deps correct.

### Round-2 verification of candidate concerns (all cleared, non-blocking)
- **Websites controlled `status`/`interval` filters not covered by `hasActiveClientFilter`** → NOT a bug: those are **server-side** query args (`useListWebsitesPaginated({ status, scanInterval })`), so they re-query the backend filtered+paginated and don't need client-side draining. Only client-side (managed) search/filters need it, and those are covered.
- **`t('history.loadingMore')` aria-label** → **pre-existing** in `origin/main` (line 445, same `useT('conversations')`); the new spinner reuses the identical established key. Not introduced here.
- **Render-time `loadMore` side-effect** → mirrors the existing, in-production pagination-mode drain; converges via Convex status transitions. Acceptable.

### Non-blocking notes (recommendations, not merge blockers)
1. **Test coverage gap:** the hook drain is well unit-tested (4 new cases), but the new `data-table.tsx` skeleton branch and the `conversations-list` drain/`isSearchingMore` behavior have no dedicated unit tests (E2E/Playwright covers the surfaces). Worth adding a data-table "active filter + hasMore → skeleton, not filtered-empty" case and a conversations-list drain test.
2. **Assertion strength:** `expect(loadMore).toHaveBeenCalled()` could assert `toHaveBeenCalledWith(pageSize*3)`; "does not drain while idle" would pass even if the fix were reverted.
3. **Skeleton with no `approxRowCount`:** when the new skeleton state triggers for a table that omits `approxRowCount`, `skeletonRowCount` is 0 (blank body). The two target tables (knowledge-entries, websites) both pass `approxRowCount: count`, so they're fine; only unrelated tables (skills/api-keys/teams) would briefly show a blank body instead of a spinner when a filter empties the buffer mid-drain.
4. **No search debounce:** every keystroke can kick a drain. The issue explicitly accepts the client-only full-load tradeoff and each drain is capped at `pageSize*3` with status-gated re-entry, so this is acceptable; a debounce would be a nice follow-up.

None of the above blocks merge. The change is minimal, follows repo conventions, fixes the reported issue across all three surfaces, tests pass, and CI is green.

**READY TO MERGE**

PR #2233, opened 2026-06-28
fix: dialogs and command palette restore focus to trigger on close (#1988)

### LINKED ISSUE #1988: Bug: Dialogs and the command palette do not return focus to their trigger on close (focus falls to <body>) — WCAG 2.4.3
### What happened?
Programmatically-opened overlays do not restore keyboard focus to the control that opened them; on close (Escape or the X button) `document.activeElement` becomes `<body>`. A keyboard/screen-reader user is dumped at the top of the page and loses their place. Reproduced across the change-password dialog, the governance DSAR & legal-hold dialogs, the chat Export & Share dialogs, and the Cmd/Ctrl+K command palette (each 2/2 runs).

### Steps to reproduce
1. Open any of: Account → Change password; Governance → File erasure request / Place legal hold; chat header → Export / Share; Cmd/Ctrl+K palette.
2. Close via Escape or the X.
3. `document.activeElement` is `<body>`, not the trigger.

### Expected behavior
On close, focus returns to the control that opened the overlay (Radix Dialog does this when given a `DialogTrigger`).

### Affected service
Platform (web app)

### Version or commit
`cd13268b8`

### Browser and OS

Live `page.evaluate` readbacks: change-password, DSAR, legal-hold, export, share, palette all return `{tag:'BODY'}` after close (focus WAS correctly trapped while open).

**Root cause.** These overlays are opened programmatically via `useState` without the shared Dialog's `trigger` prop (`services/platform/app/components/ui/dialog/dialog.tsx` — `onCloseAutoFocus` is only handled when `preventCloseAutoFocus` is set), so Radix has no trigger to restore to. The command palette (`packages/ui/src/components/search/search-command.tsx`) has the same gap.

**Fix.** Capture the previously-focused element on open and refocus it on close (or pass a real trigger / set `onCloseAutoFocus` to refocus the opener) in the Dialog primitive and SearchCommand.

---

larryro:
## Summary

Programmatically-opened overlays (dialogs opened via `useState`, the Cmd/Ctrl+K command palette) did not restore keyboard focus to the control that opened them. On close (Escape or the X button), `document.activeElement` became `<body>`, dumping keyboard and screen-reader users at the top of the page — a **WCAG 2.4.3 (Focus Order)** failure.

### Root cause

Radix Dialog's built-in `onCloseAutoFocus` restores focus to its `DialogTrigger` specifically (`triggerRef.current?.focus()`). Overlays opened without a `trigger` prop render no `DialogTrigger`, so Radix has nothing to focus and focus falls to `<body>`.

### Fix

New shared hook `@tale/ui/use-restore-focus` captures `document.activeElement` the moment the overlay opens (in a layout effect, which runs before Radix's passive focus-trap effect, so the *opener* is captured rather than an element inside the dialog) and returns an `onCloseAutoFocus` handler that re-focuses it on close. Wired into:

- `services/platform/app/components/ui/dialog/dialog.tsx` — the shared Dialog primitive (used by change-password, DSAR, legal-hold, Export, Share, …). Respects the existing `preventCloseAutoFocus` opt-out.
- `packages/ui/src/components/search/search-command.tsx` — the Cmd/Ctrl+K command palette.

The handler only calls `preventDefault()` when the opener is still in the DOM, so when the opener has unmounted Radix keeps its default behaviour.

### Tests

- New unit tests `packages/ui/src/hooks/use-restore-focus.test.ts` (4 cases: restores opener, skips removed opener, captures only while open, stable handler).
- Full `@tale/ui` suite: 1001 passed. Typecheck, lint (oxlint), and knip all green.

Closes #1988


larryro (commented):
## Desk review — focus restoration (#1988)

**Verdict: READY TO MERGE.** The fix is correct, minimal, well-tested, and fully resolves the issue. Two non-blocking polish notes below.

### Verification performed
- **CI:** all checks green (Type check, Lint, Knip, Format, Unit, UI, Storybook, Browser, all 16 Playwright platform shards, Smoke, etc.). Only fork-PR/Trivy jobs skipped, which is expected for a non-fork PR.
- **Tests run locally:** `bun run test` in `packages/ui` → **110 files, 1001 passed**. New `use-restore-focus.test.ts` → **4/4 passed**. Local `tsc --noEmit` (ui + platform), `oxlint --type-aware` → clean.

### What I confirmed
- **Effect-ordering claim is correct.** Radix `react-focus-scope@1.1.7` moves focus into the dialog in a **passive `React.useEffect`** (`dist/index.mjs:71`, focus at L81/L83). React runs all layout effects before any passive effect, so the hook's `useLayoutEffect` captures the *true opener* before Radix moves focus inward. The mechanism is sound.
- **`preventDefault()` wiring is correct.** The handler's `preventDefault()` suppresses Radix's own `focus(previouslyFocusedElement ?? document.body)` fallback (`index.mjs:90-95`), and the hook's `target.focus()` is what restores focus. The `setTimeout(0)` means the `isConnected` check runs at the right (async) time.
- **Edge cases sound.** Disconnected opener → no `preventDefault`, Radix default runs (no regression). Connected-but-non-focusable opener → falls back to `<body>`, i.e. equal to the pre-fix baseline, never worse. `preventCloseAutoFocus` opt-out preserved. Nested dialogs use independent refs — no collision.
- **Scope complete.** All overlays named in #1988 (change-password, DSAR, legal-hold, Export, Share, Cmd/Ctrl+K palette) route through the shared `Dialog` primitive or `SearchCommand`, both now wired.

### Non-blocking notes (optional polish, not required to merge)
1. **`focus()` without `{ preventScroll: true }`** (`use-restore-focus.ts:39`): Radix's own focus helper uses `{ preventScroll: true }` (`index.mjs:175`). Plain `focus()` can scroll the page to an off-screen opener on close. Arguably desirable in some cases, but diverges from Radix's convention — consider matching it.
2. **No test asserts the integration timing.** The unit tests verify the hook's logic in isolation but don't exercise the layout-vs-passive ordering against a real Radix mount, and no e2e asserts focus returns to the trigger after closing a real palette/dialog. The ordering is verified by source-reading + green Playwright, so this is non-blocking — but an e2e assertion (e.g. in `search.spec.ts`: `await expect(searchButton).toBeFocused()` after Escape) would lock the behavior in.

PR #2236, opened 2026-06-28
fix: confirm before overwriting an existing project secret (#1983)

### LINKED ISSUE #1983: Improvement: Project "Add secret" silently overwrites an existing secret with no confirmation
### What should be improved?
Adding a project secret whose name matches an existing secret silently overwrites the previous value with no confirmation or warning — a quiet way to lose a stored credential.

### Current behavior
The Add-secret dialog accepts a duplicate name and overwrites the existing secret without surfacing that a secret of that name already exists.

### Proposed change
Detect the name collision and either show an inline “name already exists” / require a different name, or require explicit confirmation before overwriting.

### Affected service
Platform (web app)

Data-safety refinement.

---

larryro:
## Summary

Resolves the data-safety issue where the Project **Add secret** dialog silently overwrote an existing secret of the same name with no warning — a quiet way to lose a stored credential.

## What changed

- `project-secrets-tab.tsx`: the add-secret form now detects a name collision against the project's existing secrets (mirroring the save logic — the `basic` type checks both the `_USERNAME`/`_PASSWORD` suffixed names). On a collision it:
  - shows an inline **"Name already in use"** warning naming the secret(s) that would be replaced,
  - gates submission behind an explicit **"Overwrite the existing secret"** confirmation checkbox (the form is invalid until checked), and
  - relabels the submit button to **Overwrite**.
  - The confirmation resets whenever the target name or type changes, so each collision is acknowledged afresh.
- Added `overwriteWarningTitle`, `overwriteWarningBody`, `overwriteConfirmLabel`, and `overwriteButton` strings to `en`/`de`/`fr` message catalogs.
- Extended `project-secrets-tab.test.tsx` with coverage that the warning appears, Save is blocked until confirmed, and a non-colliding name is unaffected.

## Verification

- `vitest` (UI project): `project-secrets-tab` — 7 passed
- `vitest` (server project): i18n `messages` parity/usage — 23 passed
- `tsc --noEmit`: clean

Closes #1983

### DIFF
--- a/services/platform/app/features/projects/components/project-secrets-tab.test.tsx
+++ b/services/platform/app/features/projects/components/project-secrets-tab.test.tsx
@@ -379,6 +379,67 @@ describe('ProjectSecretsTab', () => {
     });
   });
 
+  describe('overwrite protection', () => {
+    it('warns and blocks Save until the user confirms when the name collides with an existing secret', async () => {
+      secretsFixture = [{ name: 'EXISTING_SECRET' }];
+      const { user } = renderTab();
+
+      await user.click(screen.getByRole('button', { name: 'Add secret' }));
+      await screen.findByRole('dialog', { name: 'Add secret' });
+
+      // No warning until a colliding name is typed.
+      expect(screen.queryByText('Name already in use')).not.toBeInTheDocument();
+
+      await user.type(
+        screen.getByLabelText('Name', { exact: false }),
+        'existing_secret',
+      );
+      await user.type(
+        screen.getByLabelText('API key', { exact: false }),
+        'replacement-value',
+      );
+
+      // The collision is surfaced and the submit button becomes "Overwrite".
+      expect(screen.getByText('Name already in use')).toBeInTheDocument();
+      const overwriteButton = screen.getByRole('button', { name: 'Overwrite' });
+      expect(overwriteButton).toBeDisabled();
+
+      // Clicking while unconfirmed must not persist anything.
+      await user.click(overwriteButton);
+      expect(mockSetMutateAsync).not.toHaveBeenCalled();
+
+      // Confirming the overwrite enables the submit and lets the save through.
+      await user.click(
+        screen.getByRole('checkbox', { name: 'Overwrite the existing secret' }),
+      );
+      expect(overwriteButton).toBeEnabled();
+      await user.click(overwriteButton);
+
+      await waitFor(() => {
+        expect(mockSetMutateAsync).toHaveBeenCalledTimes(1);
+      });
+      expect(mockSetMutateAsync).toHaveBeenCalledWith(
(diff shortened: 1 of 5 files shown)


larryro:
## Desk review — #1983 (confirm before overwriting a project secret)

**Verdict: READY TO MERGE.**

I checked out the PR head (`tale/xs7ac5k74qsx8q21k61pa8m4j589f1yq`), ran the project's own tests, and reviewed the whole implementation (two rounds, fan-out across correctness / robustness / elegance / tests / boundaries, then independent re-verification of every candidate finding).

### CI & tests
- **CI: green.** `gh pr checks` — all required checks pass (Lint, Type check, Unit, UI, Browser, Storybook, Knip, Format, Opengrep, Playwright platform 1–16, Smoke, Validate images, builds). Only conditional fork-PR/Trivy jobs are skipped; nothing failing or pending.
- `bunx vitest --run --config vitest.ui.config.ts project-secrets-tab` → **7 passed**
- `bunx vitest --run --project server lib/i18n/messages.test.ts` → **23 passed**

### What's good
- Collision detection mirrors `handleSave` exactly: single name for `api_key`/`bearer`/`custom`, and the `_USERNAME`/`_PASSWORD` pair for `basic`. Names are trimmed + upper-cased on both sides, so the check matches what gets written. Partial `basic` collisions (only one of the pair exists) are correctly surfaced.
- Submission is genuinely gated — `isValid={!hasCollision || confirmOverwrite}` feeds `FormDialog`'s `disabled={... || !isValid}`, and `confirmOverwrite` resets on name/type change so each collision is re-acknowledged. A click while unconfirmed persists nothing (asserted by the test).
- Minimal, focused change; clear naming; the four new i18n keys are called literally and present with the `{names}` placeholder in en/de/fr.
- Genuinely resolves the reported issue: the dialog is the only write entry point (no edit flow), and the warning exposes only secret *names* (already visible in the list), never values.

### Non-blocking notes (optional hardening — not required for merge)
1. **Test gap, `basic` type.** The new tests cover the `api_key` path only. The `basic` two-name (`_USERNAME`/`_PASSWORD`) collision branch is untested. I verified it correct by inspection, but a test setting `type='basic'` against an existing `FOO_USERNAME` would lock the suffix-expansion behavior.
2. **Loading window.** `useProjectSecrets` returns `data ?? []`; if the dialog were opened and saved before the list resolves, the collision wouldn't be detected. Not a regression (main had no protection at all) and not realistic in the shared-query view, but gating the form on `isLoading` would close the gap fully.
3. **Pre-existing, out of scope:** a whitespace-only name passes the HTML `required`/`isDirty` check but is rejected server-side (`SECRET_NAME_RE` → `SECRET_NAME_INVALID`), so no silent loss; and the backend upsert has no server-side overwrite guard (multi-tab/multi-user races). Both predate this PR and belong to separate issues.

PR #2239, opened 2026-06-28
fix: include the entire end day in executions date-range filter (#2075)

### LINKED ISSUE #2075: Bug: Executions date-range filter excludes the entire end day (UTC-midnight collapse) — "Today" returns no rows
**Area / route**

Automations -> a workflow -> Executions sub-tab. Route: `/dashboard/$id/automations/$amId/executions`.

**Steps to reproduce**

1. Open the Executions sub-tab of any workflow that has at least one run today (after 00:00 UTC).
2. Open the date-range filter and choose a preset such as "Today" or "Last 7 days".
3. Observe the resulting table.

**Expected**

- "Today" returns runs that happened today.
- "Last 7 / 14 / 30 days" (and any range whose end date is today) include today's runs.
- More generally, executions that ran on the chosen end date appear, regardless of the time of day they ran.

**Actual**

The entire end day is excluded:

- "Today" returns essentially zero rows (only a run at exactly 00:00:00.000 UTC could match).
- Every range preset silently drops today's runs.

Root cause is a date-to-timestamp collapse with no end-of-day handling and a local-vs-UTC mismatch:

- The presets set `to = startOfDay(today)` in **local** time.
- The table serializes `to` as a bare date string: `formatISO(range.to, { representation: 'date' })` -> e.g. `"2026-06-24"`. No end-of-day adjustment.
- The backend does `toDate = new Date("2026-06-24").getTime()`, which JS parses as **UTC** midnight (`2026-06-24T00:00:00.000Z`), then filters with `.lte('startedAt', toDate)`.
- `wfExecutions.startedAt` is a numeric epoch timestamp, so any run at a time > 00:00 UTC on the end day has `startedAt > toDate` and fails the predicate.

Confirmed by computation: `new Date('2026-06-24').getTime()` = `1782259200000` (UTC midnight); a run at `2026-06-24T07:30:00Z` = `1782286200000`; `1782286200000 <= 1782259200000` is `false`, so the run is excluded.

A secondary skew exists because the picker computes `startOfDay` in local time while the backend interprets the serialized date as UTC; in non-UTC timezones this shifts the boundary further.

- `services/platform/app/components/ui/forms/date-range-picker.tsx:60-82` — `getPresetDateRange` returns `to: startOfDay(today)` (local midnight) for every range/`today` preset.
- `services/platform/app/features/automations/executions/executions-table.tsx:372-377` — `handleDateRangeChange` serializes `to` via `formatISO(range.to, { representation: 'date' })`, producing a bare `YYYY-MM-DD` with no end-of-day adjustment.
- `services/platform/app/routes/dashboard/$id/automations/$amId/executions.tsx:13,39` and `services/platform/app/features/automations/hooks/queries.ts:52-67` — `dateTo` is passed through unchanged.
- `services/platform/convex/workflow_executions/queries.ts:102-114` — `listExecutions` forwards `args` directly to the helper.
- `services/platform/convex/workflows/executions/list_executions_paginated_native.ts:30,47,61,71` — `toDate = new Date(args.dateTo).getTime()` (UTC midnight) and `.lte('startedAt', toDate)` on every index branch.
- `services/platform/convex/workflows/schema.ts:28` — `startedAt: v.number()` (epoch ms), indexed `by_definition_startedAt` (line 69).

**Notes**

- Fix the end boundary to be inclusive of the whole end day. Two clean options: (a) on serialization, send `dateTo` as the end-of-day instant (e.g. `endOfDay(range.to)` -> a full ISO datetime, with the local/UTC convention made explicit and consistent with how `dateFrom` is interpreted); or (b) in the backend, treat a bare date `dateTo` as end-of-day by adding 24h - 1ms (or using a strict `<` against the start of the next day). Whichever is chosen, make the picker's local `startOfDay` and the backend's parsing agree on a single timezone convention so the `dateFrom` lower bound and `dateTo` upper bound are symmetric.
- The same picker/serialization pattern may be reused by other list filters; if the fix lands in the shared `date-range-picker` serialization path or a shared backend date-parse helper, audit other consumers for the same off-by-one-day exclusion.

---

larryro:
Closes #2075

## Problem

The Executions date-range filter excluded the entire end day. Picking "Today" (or any range whose end is today) returned essentially no rows.

Root cause was a date-to-timestamp collapse:
- The table serialized the picker `to` value as a bare `YYYY-MM-DD` string with no end-of-day handling.
- The backend ran `new Date("2026-06-24").getTime()`, which JS parses as **UTC midnight**, then filtered with `.lte("startedAt", toDate)`.
- Since `startedAt` is an epoch timestamp, any run later than 00:00 UTC on the end day failed the predicate and was dropped.

A secondary skew existed because the picker computed `startOfDay` in **local** time while the backend interpreted the serialized date as **UTC**.

## Fix

Make the picker boundaries and the backend parsing agree on a single, fully-inclusive convention:

- **Frontend** (`executions-table.tsx`): serialize the actual full-day boundary instants — `startOfDay(from).toISOString()` and `endOfDay(to).toISOString()` — instead of a bare date. This captures the user’s local-day boundaries as absolute instants, fixing both the end-day collapse and the local/UTC skew.
- **Backend** (`date_range_filter.ts`, new shared helper): `parseExecutionDateBounds` parses those instants directly. For backward compatibility with older links that carry a bare `YYYY-MM-DD` `dateTo`, it expands the value to the last millisecond of that UTC day so the end day stays inclusive. Applied to all three consumers (`list_executions`, `list_executions_cursor`, `list_executions_paginated_native`) so they no longer diverge.

## Tests

- New `date_range_filter.test.ts` covering full-ISO parsing, bare-date end-of-day expansion, the midday-run regression, and invalid input.
- Existing backend and executions-table tests pass; `tsc --noEmit` and `oxlint --type-aware` are clean.

### DIFF
--- a/services/platform/app/features/automations/executions/executions-table.tsx
+++ b/services/platform/app/features/automations/executions/executions-table.tsx
@@ -7,7 +7,7 @@ import { HStack } from '@tale/ui/layout';
 import { Text } from '@tale/ui/text';
 import { useNavigate } from '@tanstack/react-router';
 import { type ColumnDef, type Row } from '@tanstack/react-table';
-import { parseISO, formatISO } from 'date-fns';
+import { parseISO, startOfDay, endOfDay } from 'date-fns';
 import { Copy, Check } from 'lucide-react';
 import { useState, useMemo, useCallback, memo } from 'react';
 
@@ -369,12 +369,14 @@ export function ExecutionsTable({
         query: searchTerm,
         status: status?.[0],
         triggeredBy: triggeredBy,
+        // Serialize the full-day boundary instants (start of the `from` day,
+        // end of the `to` day) so the backend's inclusive `.lte('startedAt')`
+        // covers the entire end day instead of collapsing to UTC midnight and
+        // dropping today's runs — issue #2075.
         dateFrom: range?.from
-          ? formatISO(range.from, { representation: 'date' })
+          ? startOfDay(range.from).toISOString()
           : undefined,
-        dateTo: range?.to
-          ? formatISO(range.to, { representation: 'date' })
-          : undefined,
+        dateTo: range?.to ? endOfDay(range.to).toISOString() : undefined,
       },
     });
   };
--- a/services/platform/convex/workflows/executions/date_range_filter.test.ts
+++ b/services/platform/convex/workflows/executions/date_range_filter.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, it } from 'vitest';
+
+import { parseExecutionDateBounds } from './date_range_filter';
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+describe('parseExecutionDateBounds', () => {
+  it('returns undefined bounds when no dates are provided', () => {
+    expect(parseExecutionDateBounds()).toEqual({
+      fromDate: undefined,
+      toDate: undefined,
+    });
+  });
+
+  it('parses full ISO instants directly without adjustment', () => {
+    const from = '2026-06-24T00:00:00.000Z';
+    const to = '2026-06-24T23:59:59.999Z';
+
+    expect(parseExecutionDateBounds(from, to)).toEqual({
+      fromDate: new Date(from).getTime(),
+      toDate: new Date(to).getTime(),
+    });
+  });
+
+  it('includes runs throughout the end day for a full end-of-day instant', () => {
+    const { toDate } = parseExecutionDateBounds(
+      undefined,
+      '2026-06-24T23:59:59.999Z',
+    );
+    const middayRun = new Date('2026-06-24T07:30:00.000Z').getTime();
+
+    // The regression: a midday run must satisfy `startedAt <= toDate`.
+    expect(middayRun <= (toDate ?? 0)).toBe(true);
+  });
+
+  it('expands a bare YYYY-MM-DD end date to the last ms of that UTC day', () => {
+    const { fromDate, toDate } = parseExecutionDateBounds(
+      '2026-06-24',
+      '2026-06-24',
+    );
+    const utcMidnight = new Date('2026-06-24').getTime();
+
+    expect(fromDate).toBe(utcMidnight);
+    expect(toDate).toBe(utcMidnight + DAY_MS - 1);
+  });
+
+  it('includes a midday run on a bare end date (issue #2075 regression)', () => {
+    const { toDate } = parseExecutionDateBounds(undefined, '2026-06-24');
+    const middayRun = new Date('2026-06-24T07:30:00.000Z').getTime();
+
+    // Before the fix this collapsed to UTC midnight and excluded the run.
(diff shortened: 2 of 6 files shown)


larryro:
## Desk Review — #2075 (Executions date-range end-day exclusion)

**Verdict: READY TO MERGE** ✅

Reviewed the whole implementation (not just the diff) over two rounds: 5 breadth subagents (correctness / robustness / elegance / tests / resolution+consumers) and a confirmation pass on the one potentially-blocking finding.

### Evidence
- **CI: all green.** All required checks pass (Unit, Type check, Lint, Format, Knip, Build platform/convex, Playwright ×16, Smoke test, Opengrep/Trivy scans, etc.). The only non-pass entries are appropriately `skipping` fork-PR/container jobs.
- **Tests: pass locally.** `cd services/platform && bunx vitest --run --project server convex/workflows/executions` → **9 files, 70 tests passed**. The new `date_range_filter.test.ts` directly asserts the #2075 regression (midday run on a bare end-date is now included).

### Why it's correct
- Frontend (`executions-table.tsx`) now serializes the picker selection as absolute instants — `startOfDay(from).toISOString()` / `endOfDay(to).toISOString()` — fixing both the end-day collapse and the local↔UTC skew. The `parseISO` redisplay round-trip stays correct.
- New `parseExecutionDateBounds` parses ISO instants directly; the `^\d{4}-\d{2}-\d{2}$` guard means full ISO instants are **not** given the `+DAY_MS-1` expansion, while legacy bare-date `dateTo` is — and the asymmetry (bare `dateFrom` left at UTC-midnight) is the right behavior. Epoch arithmetic is DST/overflow-safe.
- All three execution-listing paths now route through the helper — `list_executions`, `list_executions_cursor`, `list_executions_paginated_native`. Every entry point (`workflow_executions/queries.ts`, `internal_queries.ts`, and the REST API in `rest_api.ts`) lands in one of these; no remaining raw `new Date(args.dateTo).getTime()` or bare-date `formatISO`. Removed `formatISO` import is clean (Knip/Lint green).

### Non-blocking notes (optional, do not gate merge)
1. **Malformed-date degradation**: a garbage `dateFrom`/`dateTo` (only reachable via hand-crafted REST/URL input — the UI never emits one) yields `undefined` → the date filter is silently dropped. Confirmed this is **not** a data-disclosure issue: results stay scoped by `wfDefinitionId` + org (`callerOrgId` post-filter in `internal_queries.ts`, RLS in `rls_rules.ts`). The sibling `agent_tools/documents/helpers/list_documents.ts` instead throws on invalid dates; harmonizing is a possible future hardening.
2. **Test gaps (minor)**: no explicit assertion that a bare `dateFrom` is left un-expanded, and no `dateFrom`-only / mixed-format cases; the helper is otherwise well covered and the predicate paths are exercised by existing suites.
3. **Style**: `DAY_MS` constant vs the raw `86_400_000` literal used in `list_documents.ts` — cosmetic.

None of the above blocks merge. The fix is focused, well-tested, and fully resolves the reported issue.

PR #2242, opened 2026-06-28
fix: add accessible name to list-filter search inputs (#1989)

### LINKED ISSUE #1989: Bug: List-filter search inputs (agents, documents, …) have no accessible name (placeholder only) — WCAG 4.1.2
### What happened?
The list-filter search box on the agents and documents pages (and the shared list primitive behind them) has only a `placeholder` ("Search agents" / "Search documents") — no `aria-label`, no `<label for>`, no wrapping `<label>`, no `title`. A placeholder is not an accessible name; it disappears on input and is not reliably announced by assistive tech.

### Steps to reproduce
1. Open `/dashboard/{org}/agents/all` or `/dashboard/{org}/documents`.
2. Inspect the search input's accessible name.

### Expected behavior
The search input exposes a programmatic accessible name (aria-label or an associated visually-hidden label).

### Affected service
Platform (web app)

### Version or commit
`cd13268b8`

### Browser and OS

Live DOM: `agents_search_input = {ariaLabel:null, ariaLabelledby:null, title:null, placeholder:'Search agents', hasLabelFor:false, wrappedInLabel:false}` — identical for documents. (Distinct from the earlier hidden file-input false positive.)

**Fix.** Add `aria-label` (or a visually-hidden `<label for>`) to the shared list-search input primitive — one fix likely covers all list pages.

---

larryro:
## Summary

The list-filter search inputs on the agents and documents pages (and the shared list primitive behind them) exposed only a `placeholder` ("Search agents" / "Search documents") and no programmatic accessible name. A placeholder is not an accessible name — it disappears on input and is not reliably announced by assistive technology, violating WCAG 4.1.2 (Name, Role, Value).

## Fix

One fix in the shared `SearchInput` primitive (`services/platform/app/components/ui/forms/search-input.tsx`) covers every list page:

- When there is **no visible `<label>`** and **no explicit `aria-label`**, the input now derives its `aria-label` from the `placeholder`, so it always exposes an accessible name.
- An explicit `aria-label` still takes precedence.
- When a visible `<label for>` exists it already names the input, so no redundant `aria-label` is added (avoids double-announcement).

## Tests

Added unit/a11y tests to `search-input.test.tsx` covering the placeholder fallback, explicit-aria-label precedence, no-duplication with a visible label, and an axe audit of the placeholder-only case. Full `search-input` UI suite passes (23 tests). Lint (oxlint) and typecheck (tsc) are clean.

Closes #1989

### DIFF
--- a/services/platform/app/components/ui/forms/search-input.test.tsx
+++ b/services/platform/app/components/ui/forms/search-input.test.tsx
@@ -162,6 +162,57 @@ describe('SearchInput', () => {
       expect(screen.getByRole('alert')).toBeInTheDocument();
     });
 
+    it('falls back to placeholder as accessible name when no label', () => {
+      render(
+        <SearchInput value="" onChange={vi.fn()} placeholder="Search agents" />,
+      );
+      // Placeholder is not an accessible name on its own, so the input is
+      // exposed via aria-label derived from the placeholder (WCAG 4.1.2).
+      expect(screen.getByRole('textbox')).toHaveAttribute(
+        'aria-label',
+        'Search agents',
+      );
+      expect(screen.getByLabelText('Search agents')).toBeInTheDocument();
+    });
+
+    it('prefers an explicit aria-label over the placeholder', () => {
+      render(
+        <SearchInput
+          value=""
+          onChange={vi.fn()}
+          placeholder="Search agents"
+          aria-label="Filter agents"
+        />,
+      );
+      expect(screen.getByRole('textbox')).toHaveAttribute(
+        'aria-label',
+        'Filter agents',
+      );
+    });
+
+    it('does not duplicate the accessible name when a visible label exists', () => {
+      render(
+        <SearchInput
+          value=""
+          onChange={vi.fn()}
+          label="Search"
+          placeholder="Search agents"
+        />,
+      );
+      // A visible <label for> already names the input; avoid a redundant
+      // aria-label that would override it.
+      expect(
+        screen.getByLabelText('Search', { exact: false }),
+      ).not.toHaveAttribute('aria-label');
+    });
+
+    it('passes axe audit with placeholder-only accessible name', async () => {
+      const { container } = render(
+        <SearchInput value="" onChange={vi.fn()} placeholder="Search agents" />,
+      );
+      await checkAccessibility(container);
+    });
+
     it('label is associated with input', () => {
       render(
         <SearchInput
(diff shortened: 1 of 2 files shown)


larryro:
## Desk Review — #2242 (fixes #1989, WCAG 4.1.2 accessible name for list-filter search inputs)

**Verdict: READY TO MERGE.**

### CI
All required checks are **green** (verified via `gh pr checks`). The only non-passing jobs are `skipping` conditional jobs (fork-PR variants, Trivy, container tests) — expected, not failures.

### Tests (run locally)
`bunx vitest --run --config vitest.ui.config.ts app/components/ui/forms/search-input.test.tsx` → **23 passed (23)**.

I also verified the new tests genuinely guard the regression: transiently removing `aria-label={accessibleName}` made the two direct-assertion tests ("falls back to placeholder…", "prefers an explicit aria-label…") **fail**, then I restored the file. (Note: the placeholder-only axe test alone does *not* catch the revert — jsdom/axe accepts a placeholder as a name — but the two direct attribute assertions do, so the guard holds.)

### Findings
- **Correctness** — `const accessibleName = ariaLabel ?? (label ? undefined : placeholder)` is correct across all real branches: explicit `aria-label` wins; a visible `<Label>` suppresses `aria-label` (no double-naming); placeholder-only inputs now expose an `aria-label`. `placeholder` and `aria-label` are destructured out of `...props`, so the spread cannot clobber the computed values. `aria-label={undefined}` is omitted by React. ✓
- **Resolution / boundaries** — Fix is at the shared `SearchInput` level; all 13 list-filter consumers (agents, conversations, documents/data-table, automations, websites, integrations, members, providers, OneDrive picker) pass a `placeholder` and no `label`, so all are now named. `searchable-select.tsx` already sets its own `aria-label` and is unaffected. Placeholders are developer-controlled i18n strings — no injection surface. ✓
- **Robustness / elegance** — Minimal, focused change consistent with the sibling `Input` component's patterns; no interaction with the readonly-until-focus, shake, or skeleton logic. ✓

### Non-blocking notes (do not require changes)
1. The degenerate case of *no* `label`, `aria-label`, **and** `placeholder` yields no accessible name. No consumer hits this and it is out of scope for #1989; out of scope, not a defect in this fix.
2. Minor coverage gap: no explicit test for `aria-label` + `label` both supplied (logically handled by `??`). Optional to add.

**READY TO MERGE.**

PR #2244, opened 2026-06-28
fix(platform): disclose auto-disabled personalization on thread share (#2071)

### LINKED ISSUE #2071: Improvement: Sharing a project thread silently force-disables personalization; the autoDisabledPersonalization flag is ignored so the owner is never told
**Area / route**

Projects → Threads tab. `/dashboard/{org}/projects/{projectId}` (Threads tab).
Component: `services/platform/app/features/projects/components/project-threads-tab.tsx`.

**Steps to reproduce**

1. Open a project's Threads tab as the owner of one of "Your chats".
2. Pick a chat that currently has personalization enabled (i.e. `disablePersonalization` is false/unset).
3. Toggle "Share with project" ON for that chat.

**Expected**

Because sharing force-sets `disablePersonalization: true` on the thread (to prevent the owner's memories and custom instructions from leaking into replies that other project members will read), the success toast should disclose, for this specific action, that personalization was just turned off. The backend already returns `autoDisabledPersonalization` precisely so the UI can do this — it is also recorded in the audit log metadata.

**Actual**

`handleToggleShare` awaits the mutation but discards its return value and always shows the same fixed toast — `threads.shareSuccess` ("Chat shared with project") on enable, `threads.unshareSuccess` on disable. The owner gets no per-action indication that their personalization was just disabled for that thread. The only personalization-related hint is the static section-header disclosure (`threads.shareToggleDisclosure`: "Sharing this chat hides your personal memories and instructions from the responses other members will see."), which is a generic, always-present warning rather than confirmation that a state change just happened.

**Evidence**

- `services/platform/convex/projects/mutations.ts:1076-1134` — `setThreadSharedWithProject` declares `returns: { autoDisabledPersonalization: v.boolean() }`, computes `willAutoDisable = args.shared && !previousDisable` (line 1095), patches `disablePersonalization: args.shared ? true : previousDisable` (line 1100), records `autoDisabledPersonalization: willAutoDisable` in audit metadata (line 1128), and returns `{ autoDisabledPersonalization: willAutoDisable }` (line 1134).
- `services/platform/app/features/projects/components/project-threads-tab.tsx:51-76` — `await setShared({ threadId, shared: nextShared })` resolves to `{ autoDisabledPersonalization }` but the result is never captured; the toast title is `nextShared ? t('threads.shareSuccess') : t('threads.unshareSuccess')`.
- `services/platform/messages/en.json:4953-4956` — `shareToggle`, `shareToggleDisclosure`, `shareSuccess`, `unshareSuccess` strings; no per-action "personalization was disabled" string exists.
- grep confirms `autoDisabledPersonalization` has no consumer anywhere under `services/platform/app/`, and `useSetThreadSharedWithProject` has exactly one caller (this component).

**Notes**

- Functionally correct: sharing works and the server-side personalization disable is applied as intended. This is a UX/disclosure gap, not a data-integrity bug — hence improvement/medium.
- Suggested fix: capture the result and, when `autoDisabledPersonalization` is true, surface it (e.g. an additional toast line / different toast variant) using a new i18n key in all locales, e.g. `threads.shareDisabledPersonalization`. Add the matching key to every locale file under `services/platform/messages/`.
- The flag is intentionally false when sharing a thread that already had personalization disabled, and the backend deliberately does NOT re-enable personalization on unshare (mutations.ts:1099-1100), so the new toast should only fire on the auto-disable transition.

---

larryro:
## Summary

Sharing a project thread force-sets `disablePersonalization: true` so the owner's memories and custom instructions don't leak into replies other project members will read. The backend already returns `autoDisabledPersonalization` for exactly this purpose, but `handleToggleShare` awaited the mutation and **discarded** its return value, always showing the same fixed toast. The owner was never told their personalization had just been turned off for that thread.

## Changes

- `project-threads-tab.tsx`: capture the mutation result and, **only on the auto-disable transition**, add a description line to the success toast.
- New i18n key `threads.shareDisabledPersonalization` in `en`/`de`/`fr` (`de-CH` has no override for these keys → falls back to `de`).
- New component test (`project-threads-tab.test.tsx`) covering: the auto-disable transition shows the disclosure; sharing an already-non-personalized chat does not; unshare shows the plain toast.

The disclosure only fires on the `shared && !previouslyDisabled` transition — sharing a thread that already had personalization disabled, and unsharing (which intentionally does not re-enable personalization), show no extra line.

## Verification

- `bun run --filter @tale/platform test lib/i18n/messages.test.ts` — i18n parity green (23 passed)
- `bunx vitest --run --config vitest.ui.config.ts project-threads-tab` — 4 passed
- `bun run --filter @tale/platform typecheck` — clean
- `bun run --filter @tale/ui test` — 997 passed (glossary/i18n checks)
- `bun run format:check` — clean
- `oxlint` on the changed files — 0 warnings, 0 errors

Docs: N/A — reinforces the existing `threads.shareToggleDisclosure`; the share/unshare toasts have no dedicated docs section.

Closes #2071

### DIFF
--- a/services/platform/app/features/projects/components/project-threads-tab.test.tsx
+++ b/services/platform/app/features/projects/components/project-threads-tab.test.tsx
@@ -1,0 +1,210 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+import type { Id } from '@/convex/_generated/dataModel';
+import { checkAccessibility } from '@/tests/utils/a11y';
+import { render, screen, waitFor } from '@/tests/utils/render';
+
+import { ProjectThreadsTab } from './project-threads-tab';
+
+// The owner toggling "Share with project" on a personalized chat must be told,
+// for that specific action, that personalization was just turned off — the
+// backend returns `autoDisabledPersonalization` precisely so the UI can say so
+// (issue #2071). These tests pin that the success toast carries the disclosure
+// only on the auto-disable transition, and not on a plain share / unshare.
+
+const mockSetMutateAsync = vi.fn();
+const mockToast = vi.fn();
+
+type ThreadFixture = {
+  _id: string;
+  threadId: string;
+  title?: string;
+  sharedWithProject?: boolean;
+  userId: string;
+};
+
+let yoursFixture: ThreadFixture[] = [];
+let sharedFixture: ThreadFixture[] = [];
+
+vi.mock('../hooks/queries', () => ({
+  useProjectThreadSegments: () => ({
+    yours: yoursFixture,
+    shared: sharedFixture,
+  }),
+}));
+
+vi.mock('../hooks/mutations', () => ({
+  useSetThreadSharedWithProject: () => ({
+    mutateAsync: mockSetMutateAsync,
+    isPending: false,
+  }),
+}));
+
+vi.mock('@/app/hooks/use-current-user', () => ({
+  useCurrentUser: () => ({ data: { userId: 'user-1' } }),
+}));
+
+// The component imports the standalone `toast` fn directly; stub it so the
+// success/error toasts don't reach the real toast store.
+vi.mock('@/app/hooks/use-toast', () => ({
+  toast: (...args: unknown[]) => mockToast(...args),
+  useToast: () => ({ toast: mockToast }),
+}));
+
+// `<Link>` / `useNavigate` need a RouterProvider that this isolated render has
+// no access to; stub the router like the other projects component tests do.
+vi.mock('@tanstack/react-router', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('@tanstack/react-router')>()),
+  useNavigate: () => vi.fn(),
+  Link: ({ children }: { children: React.ReactNode }) => (
+    <span>{children}</span>
+  ),
+}));
+
+const PROJECT_ID = 'proj-1' as Id<'projects'>;
+
+function renderTab() {
+  return render(
+    <ProjectThreadsTab organizationId="org-1" projectId={PROJECT_ID} />,
+  );
+}
+
+// The static page chrome stacks the StickySectionHeader (h2) above the
+// PageSection (h2/h3) without an h1 in isolation; that heading-order jump is a
+// standalone-render artifact — the tab sits under the page's h1 in the app.
+const NO_HEADING_ORDER = { rules: { 'heading-order': { enabled: false } } };
+
+describe('ProjectThreadsTab', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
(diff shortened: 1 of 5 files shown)


larryro:

Reviewed PR #2244 (branch `tale/xs7c46v0ggqrysn9h2ytkdnpz989a0y3`) for issue #2071 via a two-round fan-out review (5 breadth subagents across correctness / robustness / elegance / tests / boundaries, then independent confirmation of the load-bearing claims).

**Scope of change:** `project-threads-tab.tsx` now destructures `autoDisabledPersonalization` from the `setThreadSharedWithProject` mutation result and adds a success-toast `description` only on the auto-disable transition; new i18n key `threads.shareDisabledPersonalization` added to en/de/fr. The backend already returned the flag — this is purely the missing UI surfacing, which is exactly the reported bug.

### Verification
- **CI:** all required checks **green** (`gh pr checks`). Only fork-PR-only jobs and Trivy are `skipping`, which is expected for an in-repo branch.
- **Tests:** ran `bunx vitest --run --config vitest.ui.config.ts project-threads-tab` → **4 passed / 4**. Tests use real i18n (no `useT` stub), so the asserted English strings prove the new key resolves correctly.

### Findings (all confirmed)
- **Correctness:** every branch is sound. `willAutoDisable = args.shared && !previousDisable` can only be `true` when sharing, so the disclosure never appears on unshare and the title (`shareSuccess` vs `unshareSuccess`) always matches. The destructure is safe — the mutation's `returns` validator guarantees `{ autoDisabledPersonalization: boolean }`, and on any throw the catch path runs before the destructure is reached.
- **Tests:** all three behavioral branches covered (auto-disable → description; already-non-personalized share → no description; unshare → no description, correct title) and would fail on revert. Error path is unchanged by this PR, so out of scope.
- **i18n:** key added to all three full locales; correctly nested under `threads`. `de-CH.json` (243-line partial override → falls back to `de`) and `global.json` carry none of the sibling `share*` keys, so omitting it there is consistent, not a gap.
- **Elegance:** minimal, idiomatic; `description: undefined` matches existing toast usage elsewhere in the codebase.
- **Boundaries / resolution:** owner-only mutation (auth + ownership enforced backend-side); the disclosure is shown to the owner who triggered it — no leakage. Confirmed `project-threads-tab.tsx` is the **only** UI surface that toggles sharing (`project-overview.tsx` only renders a read-only "shared" badge), so the fix is complete.

### Non-blocking note (pre-existing, out of scope)
The share `Switch` has no `disabled={isPending}` / debounce, so a rapid double-toggle can fire overlapping mutations. This predates the PR and the mutation is idempotent — not a blocker for this change.

**Verdict: READY TO MERGE.**

PR #2250, opened 2026-06-28
fix(platform): add apps catalog discovery, backToApp i18n key, and e2e coverage (#1979)

### LINKED ISSUE #1979: Improvement: Apps marketplace has no in-UI catalog/discovery, a missing back-link i18n key, and zero e2e coverage
### What should be improved?
The Apps marketplace (shipped in #1911) cannot be exercised end-to-end from the UI of a fresh org: there is no in-app catalog to discover or install an app, so every Apps hub shows the empty state and every app-detail page renders “App not found” unless apps are seeded out-of-band.

### Current behavior
No app-catalog/discovery route in the UI; installation appears to be config-seed/CLI only. Separately, the run-detail “Back to app” link has no i18n key (`apps.runs.backToApp` is MISSING — it falls back to a hardcoded English `defaultValue`), and the Apps feature has no Playwright e2e spec.

### Proposed change
Add (or document) an in-UI app catalog/install path; add `apps.runs.backToApp` to messages and translate it; add a small apps e2e spec (hub empty-state, app-detail not-found, run-detail back link) as a regression net.

### Affected service
Platform (web app)

Part of the Apps marketplace (#1911).

---

larryro:
## Summary

Resolves the three gaps called out in #1979 for the Apps marketplace (shipped in #1911):

1. **In-UI catalog / discovery** — the Apps hub previously listed only apps already copied into the org dir, so a fresh org always saw the empty state and could install only via config-seed/CLI. Added a `listCatalogApps` action that reads the built-in catalog (`<builtin>/apps/<slug>`) and a `useAppCatalog` hook; the hub now renders the **union** of installed apps and the catalog (keyed by slug, installed entry wins). Not-yet-installed catalog apps show the existing **Install** button, which already drives `installApp` (it copies from the same catalog). Refactored `install_fs` to share the new `resolveCatalogAppDir` resolver so the catalog path has a single source of truth.

2. **Missing back-link i18n key** — added `apps.runs.backToApp` to `en`/`de`/`fr` and dropped the hardcoded English `defaultValue` fallback from both run-detail routes (org- and project-scoped).

3. **Zero e2e coverage** — added `tests/e2e/specs/apps.spec.ts` covering the hub empty-state, the app-detail not-found state, and the run-detail back link (regression guard for the i18n key + navigation). The hermetic stack pins the catalog to the empty fixture dir, so the empty-state assertion stays valid.

## Verification

- `tsc --noEmit` — clean
- `oxlint --type-aware` (changed files) — 0 warnings/errors
- `oxfmt --check` (changed files) — clean
- `knip` — clean
- `vitest --project server` (apps + i18n messages) — 41 tests passing
- i18n parity across `en`/`de`/`fr` confirmed

Closes #1979


larryro:
## Desk review — NOT READY (one blocking item)

Reviewed via two rounds (5 breadth subagents + independent confirmation). **CI is green** (all 16 Playwright shards + every build/lint/type/unit check pass; the `skipping` entries are conditional fork-PR/path-filtered jobs, not failures). Locally `bunx vitest --run --project server convex/apps lib/shared/schemas/apps messages` → **55 passed**.

All three gaps from #1979 are genuinely addressed and the code is correct — but one gap blocks merge.

### 🔴 Blocking — the headline catalog-discovery path has no test coverage
The new feature (gap #1) is `listCatalogApps` → `useAppCatalog` → the union in `apps-grid.tsx` → the Install button. **Every existing test runs against an empty catalog** (the e2e stack pins `TALE_CONFIG_BUILTIN_DIR` to `tests/e2e/fixtures/config/default/apps`, which holds only `.gitkeep`), so:
- `listCatalogApps` — only its empty/ENOENT branch is reached; the **manifest-projection** branch and the **malformed-manifest skip** branch are untested (no `convex/apps/file_actions.test.ts`).
- `apps-grid.tsx` — the union `useMemo` (dedupe-by-slug, installed-wins, sort-by-name) and the catalog-only Install wiring have no component test, and the e2e union is always trivially empty.

A regression in the catalog projection, the union, or the install-from-catalog wiring would pass every current test. For a PR whose stated purpose includes adding coverage, the primary new behavior must be exercised.

**Please add one of:**
- (preferred) a server/unit test for `listCatalogApps` asserting the projected summary fields for a valid manifest **and** that a malformed manifest is skipped (non-fatal); or
- an e2e fixture catalog with one app + an assertion in `tests/e2e/specs/apps.spec.ts` that a fresh org's hub renders it as an installable discovery card.

### ✅ Verified clean / refuted (no action needed)
- **Install-path refactor parity** — `appBundleTemplateDir` → `resolveCatalogAppDir` resolves byte-identically to the old inline form; the added `isValidAppSlug` check is additive hardening (all callers already pre-validate).
- **"Path traversal in the catalog manifest read" — not a vuln.** `isValidAppSlug` (`/^[a-z0-9]+(-[a-z0-9]+)*$/`, len ≤ 64) admits no `.`/`/`, and slugs come from `readdir` already filtered by it. Mirroring `listApps`' `safeJoinWithinDir` helper would be a nice consistency tweak but is not required.
- **"Unset `TALE_CONFIG_BUILTIN_DIR` throws instead of returning []" — intended.** Matches the install path and the resolver comment; the docstring's "missing dir → empty list" is the handled ENOENT case.
- Nits (optional): `?? []` fresh-array fallbacks make the union `useMemo` recompute while loading (negligible); the `listApps`/`listCatalogApps` projection duplication is acceptable per existing convention.

### Resolution confirmed
- Gap 1: discovery and install share the same `<builtin>/apps` source, so every discovered app is installable; installed entry wins the union. Logic correct — only the testing is missing.
- Gap 2: `apps.runs.backToApp` added to en/de/fr (correctly nested); `defaultValue` removed from both run-detail routes; parity test passes.
- Gap 3: `apps.spec.ts` covers empty-state, not-found, and back-link navigation.

**Verdict: NOT READY — CHANGES REQUIRED.** Add the missing test for the populated catalog-discovery path; once it lands and CI stays green, this is mergeable.

larryro:
## Desk Review — PR #2250 (resolves #1979): Apps catalog discovery, `backToApp` i18n key, e2e coverage

**VERDICT: NOT READY — CHANGES REQUIRED.** One confirmed blocking defect: the new in-UI catalog discovery is half-wired — clicking a discovered (not-yet-installed) app card lands on an "App not found" dead-end instead of its pre-install details/install page.

### CI state
All required checks **PASS** (green) at HEAD `ceb7af6` — verified via `gh pr checks` and `statusCheckRollup` (0 non-success/non-skipped checks). Includes Type check, Lint, Format, Knip, Unit, all 16 Playwright platform shards, builds, and security scans. The few `skipping` entries are fork-PR/optional jobs, not failures. CI is **not** a blocker.

### Tests I ran (locally, on the checked-out branch)
- `bunx vitest --run --project server` (full server project) → **543 files, 6505 tests passed**. No regressions.
- `bunx vitest --run --project server convex/apps` (incl. the new `file_actions.test.ts`) → **17 passed**.
- `bunx vitest --run --project server lib/i18n/messages.test.ts` (i18n parity) → **23 passed**.

---

### BLOCKING

**1. Catalog-only discovery cards dead-end on "App not found" (the core of the new feature is half-wired).**
- `services/platform/app/features/apps/components/apps-grid.tsx:92` — the **entire** app card is wrapped, unconditionally, in `<Link to="/dashboard/$id/apps/$appSlug">` (`cardLink`, lines 86–88). This now applies to catalog-only apps too (the new union surfaces them), so clicking anywhere on a discovery card navigates to the org-level app detail page.
- `services/platform/app/features/apps/components/app-page.tsx:700` resolves the app via `const app = apps.find(a => a.slug === appSlug)`, where `apps` comes **only** from `useApps` (line 696). `useApps` → `listApps` reads **only the org's on-disk apps dir** (`resolveAppsDir(orgSlug)`); it never includes the built-in catalog. `app-page.tsx` does **not** import or union `useAppCatalog`.
- Therefore, for a not-yet-installed catalog app, `app` is `undefined`, and `app-page.tsx:712` short-circuits to the `notFound.title` ("App not found") EmptyState — **before** reaching the pre-install `AppDetails` branch at lines 755–764, which is itself gated on `app` being truthy and so is unreachable for a catalog-only app.
- Net effect: the hub invites the user to click a discovered app, and the click contradicts itself with "App not found." This breaks the feature's stated goal ("a fresh org can browse and install from the UI") and is an inconsistency between the installed-card branch (card → working detail) and the new catalog-card branch (card → dead end). Independently confirmed by a refutation pass: the card Link is unconditional, `listApps` reads the org dir only, no route `loader`/`beforeLoad` intercepts the slug, no org-creation step seeds catalog apps into the org dir, and `AppDetails` requires `app` to be truthy.
- **Required fix (name the file + branch):** In `services/platform/app/features/apps/components/app-page.tsx`, make the not-yet-installed (`!state`) org-route branch resolve catalog apps too — union `useAppCatalog(organizationId)` into the `app` lookup the same way `apps-grid.tsx` does (installed entry wins), so a catalog-only slug resolves and renders the existing `AppDetails` (full description + Install CTA) instead of `notFound`. Alternatively, route catalog-only cards in `apps-grid.tsx` to the install/details flow rather than the dead detail Link. Whichever path is chosen, a not-yet-installed app discovered in the hub must NOT render "App not found" when clicked. Add e2e coverage for it (see below) so it can't regress.

---

### Non-blocking (should address, but not gating on their own)

2. **The headline feature has no end-to-end coverage of a *populated* catalog.** `services/platform/playwright.config.ts:129` pins `TALE_CONFIG_BUILTIN_DIR` to `tests/e2e/fixtures/config/default`, whose `apps/` subdir is empty (only `.gitkeep`). So `listCatalogApps` returns `[]` in e2e and the new spec (`tests/e2e/specs/apps.spec.ts`) only ever exercises the **empty** hub, the unknown-slug not-found, and the back link. The discovery cards, the Install button, and (critically) the catalog-card navigation are never rendered in any test — which is exactly why the blocking dead-end above is invisible to CI. The three specs do match the issue's literal ask, so this isn't independently blocking, but a populated-catalog fixture + a spec asserting a discovered card opens its pre-install page (not "App not found") would have caught Finding 1 and should accompany the fix.

3. **Minor — variable shadowing.** `apps-grid.tsx`: the `useMemo` builds a local `const bySlug = new Map(...)` (line ~53) that shadows the outer `const { bySlug } = useAppInstallStates(...)` (line ~60). No functional bug (the inner one is block-scoped), but it's confusing; rename the local (e.g. `unionBySlug`).

4. **Minor — untested branch.** `listCatalogApps` rethrows non-`ENOENT` `readdir` errors (`file_actions.ts`, the `catch`/`throw err` after the directory read); that branch is untested. This parallels `listApps` (also untested there), so it's pre-existing parity, not a new gap — optional.

---

### Verified solid (no action needed)
- **`resolveCatalogAppsDir` / `resolveCatalogAppDir` (`file_utils.ts`) + `install_fs.ts` refactor:** path-equivalent to the removed inline resolver for valid slugs; the new `isValidAppSlug` guard is a safe, redundant defense (callers already validate). `BUILTIN_ENV` cleanly relocated; no stale refs; `path` import still used.
- **Path-traversal safety:** `isValidAppSlug` regex `^[a-z0-9]+(-[a-z0-9]+)*$` (≤64) rejects `..`, `/`, `\`, dots, underscores, uppercase.
- **`listCatalogApps` projection:** field-by-field parity with `listApps` (minus per-install `views`/`messages`, intentionally `views: []` / `messages` omitted, documented); ENOENT→`[]`, malformed manifest skipped with a warn, slug-filtered, sorted.
- **`useAppCatalog` hook:** distinct query key `['apps','catalog',orgId]` (no collision with `['apps','list',orgId]`); correct return shape. Post-install the card flips correctly because installed wins in the union and `useAppInstallStates` drives the badge.
- **i18n:** `apps.runs.backToApp` added to en/de/fr with correct translations; both run-detail routes (`$id/.../runs/$executionId.tsx` and the `$projectId` one) now call `t('runs.backToApp')` under `useT('apps')`; parity test passes.
- **Unit tests (`file_actions.test.ts`):** sound mocks (action passthrough + membership stub), correct env save/restore, real branch coverage (valid/minimal/malformed×2/invalid-slug/ENOENT/sort) with non-tautological assertions (incl. `not.toHaveProperty` for optional-field leakage and warn-count==2).
- **e2e spec (for what it covers):** empty-state, not-found, and back-link assertions are correct and non-flaky; back link renders unconditionally; selectors are role/level-stable.

---

**FINAL VERDICT: NOT READY — CHANGES REQUIRED.**
1. `app-page.tsx` (org route, not-installed `!state` branch): resolve catalog apps via `useAppCatalog` (union, installed wins) so a not-yet-installed app discovered in the hub renders its `AppDetails` pre-install/Install page instead of the `notFound` EmptyState; equivalently, route catalog-only cards in `apps-grid.tsx` to the install/details flow. A discovered app must never show "App not found" when its card is clicked.
2. (Strongly recommended, with the fix) Add e2e coverage with a non-empty catalog fixture asserting a discovered card opens its pre-install page, so this path stops being untested.

larryro:
## Rework — addressed the 15:42 desk review (HEAD `ceb7af6` → `30cdbece`)

**Blocking #1 — catalog-only discovery cards dead-ended on "App not found":** fixed in `app-page.tsx`. `AppPage` now falls back to `useAppCatalog(organizationId)` when the org's installed list doesn't carry the slug (installed entry wins, mirroring the union in `apps-grid.tsx`). A not-yet-installed app discovered in the hub now resolves to its pre-install `AppDetails` page (full description + Install CTA) instead of the `notFound` EmptyState. The loading guard also waits on `catalogLoading` so we never flash "App not found" while the catalog is still loading.

**Coverage for the fix:** added `app-page.test.tsx` (client/jsdom) with two cases — (a) an org with an empty installed list + a one-app catalog renders the discovered app's name + **Install** button and **not** "App not found"; (b) a slug present in neither still renders the "App not found" heading. This directly guards the dead-end regression.

A note on the e2e route (Finding #2): `TALE_CONFIG_BUILTIN_DIR` is set once on the Playwright `webServer`, so it's process-wide for the whole run. The existing **empty-state** spec depends on an empty catalog under that same dir, so populating the fixture catalog would break it. The component test exercises the exact union/fallback logic deterministically without that conflict, so I covered the fix there instead.

**Non-blocking #3 — variable shadowing:** renamed the inner `bySlug` map in `apps-grid.tsx`'s union `useMemo` to `unionBySlug`.

### Verification (local)
- `tsc --noEmit` — clean
- `oxlint --type-aware` (changed files) — 0 warnings/errors
- `vitest --project client app/features/apps/components` — 6 passed (incl. the 2 new app-page cases)
- `vitest --project server convex/apps lib/i18n/messages.test.ts lib/shared/schemas/apps` — 47 passed

larryro:
## Desk review — PR #2250 (issue #1979): Apps catalog discovery, `backToApp` i18n key, e2e coverage

**Verdict: NOT READY — one required change (a test gap). Implementation itself is correct, secure, and CI-green.**

Reviewed at head of `tale/xs7bt6hmv537295kexynpt1cx989erq5` via a two-round fan-out (breadth across correctness / robustness / tests / elegance / boundaries, then independent re-verification of every blocking candidate).

### CI & tests (observed)
- **CI: green.** All required checks pass (Type check, Lint, Format, Knip, Unit, UI, Smoke, Validate images, all 16 Playwright platform shards, etc.); the only non-passing entries are `skipping` (fork-PR/container variants), none failing/pending.
- **Tests I ran locally** (deps installed via `bun install`; this sandbox has a pre-existing zod/edge-runtime bundler resolution quirk that fails ~15 unrelated convex suites at `z.object` import, so I re-ran the affected suites with zod inlined to exercise the real logic):
  - New backend unit: `vitest convex/apps/file_actions.test.ts` → **6/6 pass**.
  - New component: `vitest app/features/apps/components/app-page.test.tsx` → **2/2 pass**.
  - Unaffected suite sanity: **800 tests pass**.

### What's correct (verified, not blocking)
- `listCatalogApps` projection matches `listApps` field-for-field (minus the intentional per-install `views`/`messages`); all branches sound: ENOENT→`[]`, non-ENOENT rethrows, malformed/non-JSON manifest → warn+continue (skips one app, never the list), invalid-slug dirs filtered, deterministic sort.
- `install_fs` refactor cleanly delegates to the shared `resolveCatalogAppDir`; `BUILTIN_ENV` re-homed to `file_utils.ts` with no dead code/imports left behind.
- Frontend union (`apps-grid.tsx`) and page fallback (`app-page.tsx`) are correct: Map keyed by slug, installed inserted last so it wins; catalog-only apps render the pre-install AppDetails (empty `views`, undefined install-state/messages all handled without crashing) and the Install CTA drives `installApp`. Loading gate `((isLoading || catalogLoading) && !app) || stateLoading` is correct.
- Security/boundaries: membership-gated; slugs only ever reach `path.join` after `isValidAppSlug` (kebab-case, ≤64, no `.`/`_`) — no traversal surface. All three #1979 gaps are genuinely addressed; `apps.runs.backToApp` added to en/de/fr with correct nesting (de-CH is a partial override that inherits from de; global.json is a separate namespace).

### BLOCKING — required change
1. **No test exercises the catalog↔installed union — the heart of gap #1 (in-UI discovery).** `apps-grid.tsx:52-59` builds the union, lets an installed entry **win over a same-slug catalog entry** (so per-install `views`/`messages` aren't lost), and **sorts by `name`**. There is **no `apps-grid.test.tsx`**, no test imports `AppsGrid`, and the e2e suite pins the catalog to an **empty** fixture (`tests/e2e/fixtures/config/default/apps/` = only `.gitkeep`), so the overlap/precedence/sort path is never run by any unit, component, or e2e test. The page-level fallback `apps.find() ?? catalog.find()` in `app-page.tsx:707-709` is likewise only tested for catalog-only (`installed=[]`), never for the both-present case where installed must win. A future loop-order/precedence regression would silently drop installed apps' data with nothing to catch it.
   - **Fix:** add an `apps-grid` test (mirror `app-page.test.tsx`'s mocking of `useApps`/`useAppCatalog`/`useAppInstallStates`) asserting (a) the rendered set is the union by slug, (b) an installed entry overrides a same-slug catalog entry, (c) ordering is by `name`. Add (or extend `app-page.test.tsx` with) a case where the same slug is in both installed and catalog to lock in installed-wins at the page level too.

### Non-blocking recommendations (not required to merge)
- `file_actions.ts`: `listCatalogApps` re-duplicates the full summary projection from `listApps`. Shapes match **today**, but a future manifest field added to one and not the other will silently desync the union. Extract a shared `projectAppSummary(manifest)` helper as the single source of truth.
- Robustness (defensive only): if `TALE_CONFIG_BUILTIN_DIR` were unset, `listCatalogApps` throws (vs `listApps` returning `[]`) and `apps-grid` ignores the query `error`, so the hub would render the empty state rather than an error. In practice the env is always set (Dockerfile, `dev-engine.ts` auto-default, `playwright.config.ts`) and org scaffolding/install already hard-require it, so this is not a practical regression — noted for completeness.

**NOT READY — CHANGES REQUIRED** (single blocking item: the missing union/precedence test above).

PR #2251, opened 2026-06-28
fix(platform): cover customer name in conversations inbox search (#1996)

### LINKED ISSUE #1996: Improvement: Conversations inbox search does not cover the customer name (the most prominent row label)
### What should be improved?
The conversations inbox search filters by subject/body but not by the customer name — the most prominent label on each row — so searching for the visible customer returns zero results.

### Current behavior
Typing a customer name that is clearly shown on the rows yields an empty list; only subject/body matches are returned.

### Proposed change
Include the resolved customer name in the inbox search index/filter (and consider matching the conversation subject as primary).

### Affected service
Platform (web app)

UX/search-coverage refinement; relates to the conversations i18n work in #1978.

---

larryro:
## Summary

The conversations inbox search filtered by subject/body but **not** the customer name — the most prominent label on each row — so searching for a visible customer returned zero results.

## Changes
- Extend `filterByTextSearch` (`lib/utils/filtering.ts`) to accept **accessor functions** in addition to top-level keys, so callers can match nested values such as a related customer's name. Accessor return type is constrained to stringifiable primitives.
- Include the resolved customer name (`(c) => c.customer?.name`) in the conversations inbox search fields (`app/features/conversations/components/conversations.tsx`).
- Add unit tests for `filterByTextSearch` covering accessor-based (nested) matching, the empty-search case, null-safe accessors, and `startsWith` mode.

## Verification
- `vitest --run --project server lib/utils/filtering.test.ts` → 6/6 pass
- `tsc --noEmit` → clean
- `oxlint --type-aware` → 0 errors/warnings

Closes #1996

### DIFF
--- a/services/platform/app/features/conversations/components/conversations.tsx
+++ b/services/platform/app/features/conversations/components/conversations.tsx
@@ -117,6 +117,9 @@ export function Conversations({
         'description',
         'subject',
         'externalMessageId',
+        // The customer name is the most prominent label on each row, so the
+        // search must cover it too — it lives on the nested `customer` object.
+        (c) => c.customer?.name,
       ]);
     }
 
--- a/services/platform/lib/utils/filtering.test.ts
+++ b/services/platform/lib/utils/filtering.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from 'vitest';
+
+import { filterByTextSearch } from './filtering';
+
+interface Conversation {
+  title: string;
+  subject: string;
+  customer: { name?: string } | null;
+}
+
+const conversations: Conversation[] = [
+  {
+    title: 'Refund request',
+    subject: 'Order #12',
+    customer: { name: 'Alice Johnson' },
+  },
+  {
+    title: 'Shipping delay',
+    subject: 'Where is it',
+    customer: { name: 'Bob Smith' },
+  },
+  { title: 'General question', subject: 'Hi', customer: null },
+];
+
+describe('filterByTextSearch', () => {
+  it('returns all items when the search term is empty', () => {
+    expect(filterByTextSearch(conversations, '', ['title'])).toHaveLength(3);
+    expect(filterByTextSearch(conversations, '   ', ['title'])).toHaveLength(3);
+  });
+
+  it('matches a top-level field case-insensitively', () => {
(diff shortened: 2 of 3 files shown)


larryro:
## Desk Review — Issue #1996 (inbox search to cover customer name)

**Verdict: READY TO MERGE.**

### CI
All required checks green: `gh pr checks` reports **46 pass, 5 skipping, 0 failing, 0 pending** — including Build platform, Lint, Format, Knip, Opengrep, and the full Playwright matrix.

### Tests
Ran the new suite under the project's own config:
`bunx vitest --run --project server lib/utils/filtering.test.ts` → **6 passed (6)**.
The file lives in `lib/utils/` and is not excluded, so it is picked up by the `server` project and runs in `bun test` / CI.

### Review (5-dimension fan-out, two rounds — all findings re-verified)
- **Correctness** — `filterByTextSearch` resolves `typeof field === 'function' ? field(item) : item[field]` (filtering.ts:90). The `value == null` guard (filtering.ts:91) safely short-circuits when `customer.name` is `undefined` (validator marks `name` optional). Empty/whitespace term, includes/startsWith modes, and String() coercion all behave consistently with the existing fields. No branch fails silently.
- **Resolution** — `conversations.tsx:122` adds `(c) => c.customer?.name` to the inbox search fields. The row's prominent label is `conversation.customer?.name || ...` inside the `<Heading>` (conversations-list.tsx:232), so the fix targets exactly the user-visible label the issue describes. Search is client-side for all fields (no server-side search path to keep in sync — confirmed at the route).
- **Robustness** — Pure function; `customer` is a required object per `conversationItemValidator`, and optional chaining covers the runtime edge. No leaks/lifecycle concerns.
- **Elegance** — Minimal, focused change. `SearchField<T>` widening (`keyof T | accessor`) is backward compatible; existing callers (use-list-page.ts:213, filter-documents.ts:61/66) pass string keys and remain valid. Clear explanatory comment.
- **Tests** — 6 tests cover empty/whitespace, top-level match, nested accessor match, mixed fields, null-safe accessor, and startsWith mode.

### Non-blocking nit (optional)
- `filtering.ts:66` — JSDoc example reads `(c) => c.customer.name` without optional chaining, whereas the real usage and tests use `c.customer?.name`. Consider matching the example to the safe form. Does not affect correctness or the verdict.

PR #2253, opened 2026-06-28
fix: enforce org-membership check on file_metadata mutations (#2039)

### LINKED ISSUE #2039: Bug: file_metadata mutations skip the org-membership check — cross-org write + RAG knowledge-poisoning
**Area / route**
Knowledge — documents upload / transcription / `fileMetadata` state machine. Backend: `services/platform/convex/file_metadata/mutations.ts`.

**Steps to reproduce**
As an authenticated user who is a member of org A only:
1. Call `files.mutations.generateUploadUrl` (open to any authenticated user — `services/platform/convex/files/mutations.ts:7-18`) and POST bytes to obtain a `storageId`.
2. Call `file_metadata.mutations.saveFileMetadata` with `organizationId` set to a **different** org B that you are **not** a member of (and no `threadId`, or a `threadId` that doesn't resolve). The mutation runs only `getAuthUserIdentity` (JWT presence) + `checkUploadPolicy` (no membership check) and inserts a `fileMetadata` row scoped to org B, then schedules `uploadFileToRag(organizationId: B)` and `runRetentionCleanup` for org B.
3. To clobber/re-trigger transcription on a foreign file, call `retryTranscription` / `skipTranscription` with a foreign `storageId` plus the matching `organizationId` of org B. The only org check is `metadata.organizationId === args.organizationId`, which the caller controls on both sides and which never asserts membership.

**Expected**
Every write mutation asserts the caller is a member of `args.organizationId` (or, for retry/skip, of `metadata.organizationId`) via `getOrganizationMember(ctx, organizationId, authUser)` — exactly as the sibling `documents/mutations.ts` does (`updateDocument` :49, `deleteDocument` :75, `createDocumentFromUpload` :138). `getOrganizationMember` throws `UnauthorizedError` for non-members and for `disabled` members.

**Actual**
`saveFileMetadata`, `retryTranscription`, and `skipTranscription` gate only on JWT presence (`getAuthUserIdentity`) and — for retry/skip — a self-reported org-equality check. No `getOrganizationMember` call. Consequences: a non-member (or removed/disabled member who still holds a valid JWT) can inject `fileMetadata` rows into a foreign org and RAG-index attacker-controlled content into that org's RAG namespace (knowledge poisoning), drive retention cleanup against an org they don't belong to, and clobber a transcript / re-trigger Whisper billing on a foreign `storageId`.

- `services/platform/convex/file_metadata/mutations.ts:36-39` — `saveFileMetadata` only `getAuthUserIdentity`.
- `services/platform/convex/file_metadata/mutations.ts:43-53` — `checkUploadPolicy` is the only org-aware call; it has no membership check.
- `services/platform/convex/governance/upload_enforcement.ts:29-30` — returns `{ allowed: true }` when no upload policy is configured (the default), so for most orgs there is zero gating.
- `services/platform/convex/file_metadata/mutations.ts:267-279` (`skipTranscription`) and `:321-333` (`retryTranscription`) — only `metadata.organizationId === args.organizationId`, both values caller-supplied.
- `services/platform/convex/file_metadata/internal_actions.ts:31-56` — `uploadFileToRag` consumes `args.organizationId` for the RAG namespace; `services/platform/convex/workflow_engine/action_defs/rag/rag_action.ts:71` documents the exact threat ("namespace — cost shift + content injection").
- `services/platform/convex/files/mutations.ts:11-17` — `generateUploadUrl` open to any authenticated user.
- Contrast: `services/platform/convex/documents/mutations.ts:49, 75, 138` — sibling mutations call `getOrganizationMember`.
- `services/platform/convex/lib/rls/organization/get_organization_member.ts:133-143` — the authoritative gate (throws for non-member and `disabled`).
- These are public mutations (`server.d.ts:52` — `MutationBuilder<DataModel,"public">`) and are wired to the client at `services/platform/app/features/chat/hooks/use-convex-file-upload.ts:66-73`.

**Notes**
Fix: add `await getOrganizationMember(ctx, args.organizationId, authUser)` to `saveFileMetadata` (after the auth check, before `checkUploadPolicy`), and to `retryTranscription` / `skipTranscription` using `metadata.organizationId` (after the row is loaded). The existing `threadId` cross-org gate in `saveFileMetadata` (lines 59-72) is not a substitute: it only fires when a `threadId` is supplied and resolves to a thread, so Document-Hub / no-thread uploads bypass it entirely. The self-reported `organizationId === metadata.organizationId` checks in retry/skip should be kept (they catch caller mistakes) but are not an authorization boundary on their own.

---

larryro:
## Summary

Closes #2039.

`file_metadata` write mutations gated only on JWT presence (`getAuthUserIdentity`) plus a self-reported `organizationId === metadata.organizationId` check — neither of which asserts the caller is actually a member of the target org. Any authenticated user could therefore:

- inject `fileMetadata` rows into a **foreign** org and RAG-index attacker-controlled content into that org's namespace (knowledge poisoning),
- drive retention cleanup against an org they don't belong to,
- clobber a transcript / re-trigger Whisper billing on a foreign `storageId`.

This mirrors the sibling `documents/mutations.ts`, which already calls `getOrganizationMember`.

## Changes

`services/platform/convex/file_metadata/mutations.ts`:
- `saveFileMetadata` — `await getOrganizationMember(ctx, args.organizationId, authUser)` after the auth check, before `checkUploadPolicy`.
- `skipTranscription` / `retryTranscription` — `await getOrganizationMember(ctx, metadata.organizationId, authUser)` after the row loads. The existing self-reported `organizationId === metadata.organizationId` checks are kept (they catch caller mistakes) but are no longer relied on as the authorization boundary.

`getOrganizationMember` throws `UnauthorizedError` for non-members and for `disabled` members (so a removed/disabled member holding a still-valid JWT is rejected).

## Tests

`services/platform/convex/file_metadata/mutations.test.ts` — added authorization tests for all three mutations (non-member is rejected before any write/schedule; valid member still succeeds).

- `bunx vitest --run --project server file_metadata/mutations.test.ts` → 19 passed
- `bunx tsc --noEmit` → clean
- `bunx oxlint --type-aware` on changed files → 0 warnings / 0 errors

### DIFF
--- a/services/platform/convex/file_metadata/mutations.test.ts
+++ b/services/platform/convex/file_metadata/mutations.test.ts
@@ -69,6 +69,21 @@ vi.mock('../governance/upload_enforcement', () => ({
   checkUploadPolicy: vi.fn().mockResolvedValue({ allowed: true }),
 }));
 
+// Org-membership gate (#2039). By default every test runs as a valid member;
+// the authorization tests override this to reject (non-member).
+const mockGetOrganizationMember = vi.fn();
+vi.mock('../lib/rls/organization/get_organization_member', () => ({
+  getOrganizationMember: (...args: unknown[]) =>
+    mockGetOrganizationMember(...args),
+}));
+
+const VALID_MEMBER = {
+  _id: 'member_1',
+  organizationId: 'org_1',
+  userId: 'user_1',
+  role: 'member',
+};
+
 function createMockCtx(existingDoc: Record<string, unknown> | null = null) {
   const builder = {
     withIndex: vi.fn().mockReturnThis(),
@@ -131,6 +146,7 @@ const AUTH_USER = {
 describe('saveFileMetadata (public)', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    mockGetOrganizationMember.mockResolvedValue(VALID_MEMBER);
   });
 
   it('rejects unauthenticated requests', async () => {
@@ -141,6 +157,27 @@ describe('saveFileMetadata (public)', () => {
     await expect(handler(ctx, baseArgs)).rejects.toMatchObject({
       data: { code: 'UNAUTHENTICATED' },
     });
+  });
+
+  it('rejects callers who are not a member of the target org', async () => {
+    mockGetAuthUser.mockResolvedValue(AUTH_USER);
+    // Caller holds a valid JWT but is not a member of org_1 — the gate
+    // must throw before any fileMetadata row is written or RAG is queued.
+    mockGetOrganizationMember.mockRejectedValue(
+      new Error('Not a member of organization org_1'),
+    );
+    const { ctx } = createMockCtx(null);
+    const handler = await getHandler();
+
+    await expect(handler(ctx, baseArgs)).rejects.toThrow(
+      'Not a member of organization org_1',
+    );
+    expect(mockGetOrganizationMember).toHaveBeenCalledWith(
+      ctx,
+      'org_1',
+      expect.objectContaining({ userId: 'user_1' }),
+    );
+    expect(ctx.db.insert).not.toHaveBeenCalled();
   });
 
   it('rejects uploads blocked by organization policy, preserving the reason', async () => {
@@ -517,6 +554,7 @@ const skipArgs = { storageId: 'storage_1', organizationId: 'org_1' };
 describe('skipTranscription (public)', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    mockGetOrganizationMember.mockResolvedValue(VALID_MEMBER);
   });
 
   it('rejects unauthenticated requests', async () => {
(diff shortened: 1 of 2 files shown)


larryro:
## Desk Review — #2039 org-membership gate on `file_metadata` mutations

**Verdict: READY TO MERGE.**

Reviewed via a two-round method (5 breadth subagents across correctness / robustness / elegance / tests / boundaries, then independent confirmation of every claim I would block on).

### CI & tests
- **CI: green.** `gh pr checks` → 46 pass, 5 skipping (conditional: fork-PR / container / Trivy), 0 failing, 0 pending.
- **Tests: pass.** Ran `bunx vitest --run --project server convex/file_metadata/mutations.test.ts` → **19 passed**.

### What's correct
- All three public mutations now gate on membership before any write/schedule:
  - `saveFileMetadata` — `getOrganizationMember(ctx, args.organizationId, authUser)` at `mutations.ts:60`, after the auth check and **before** `checkUploadPolicy`, the `threadId` gate, and every `db.insert`/`scheduler.runAfter` (both the new-row and existing-row patch paths).
  - `skipTranscription` — `mutations.ts:304`, against `metadata.organizationId` (the row's own org) after the row loads, before `runMutation`.
  - `retryTranscription` — `mutations.ts:363`, same pattern, before `db.patch`/`scheduler.runAfter`.
- Checking `skip`/`retry` against `metadata.organizationId` (not the caller-supplied `args.organizationId`) is the right boundary; the retained `metadata.organizationId === args.organizationId` equality check is correctly kept as a caller-mistake guard, as the issue prescribed.
- Pattern mirrors the sibling `documents/mutations.ts` (`:49,:75,:138`) exactly — import placement, call shape, defensive position.
- **Completeness:** `file_metadata/` has exactly these 3 public `mutation()` exports; all other writers (`internal_mutations.ts`, `video_links`, `webdav` via the *internal* `saveFileMetadata`) are `internalMutation` and not client-callable. `uploadFileToRag` is only scheduled from the now-gated public mutations and internal mutations. All four reported consequences (foreign-org row injection, RAG namespace poisoning, retention-cleanup abuse, transcript clobber / Whisper re-bill) are closed.
- Tests assert the gate is actually invoked (`toHaveBeenCalledWith`) and that no `insert`/`patch`/`runMutation`/`scheduler.runAfter` fires on rejection, for all three mutations, plus valid-member happy paths.

### Non-blocking observations (optional)
1. The disabled-member rejection path (`getOrganizationMember` also throws for `role === 'disabled'`) is exercised in the helper's own tests but not separately in `mutations.test.ts`; the non-member rejection test already proves the gate is wired and rejects before any write, so this is a nice-to-have, not a gap in the fix.
2. The gate throws `UnauthorizedError` while adjacent failures throw generic `Error` — this is consistent with the established `documents/mutations.ts` convention, so no change required.

Neither affects the merge decision.

PR #2255, opened 2026-06-28
fix: report the actual per-type ceiling in the chat upload size-limit toast (#2048)

### LINKED ISSUE #2048: Bug: Audio/video chat uploads in the 100MB–200MB band are wrongly rejected — Math.min clamps the per-type media ceiling back to the generic cap

larryro:
## Summary

Follow-up to #2156, which fixed the primary defect in #2048 (the `Math.min(maxFileSize, mediaCeiling)` clamp that wrongly rejected 100–200 MB audio/video at the per-file gate). This PR closes the **second defect** the issue called out: the file-too-large rejection toast always reported `mergedConfig.maxFileSize` (100 MB) regardless of file type, so when a media file *was* rejected (e.g. above the 2 GB media cap, or against a governance policy cap) the toast stated the wrong limit.

## Changes

- `services/platform/app/features/chat/hooks/use-convex-file-upload.ts`
  - `rejectedTooLarge` now records the actual ceiling each file exceeded (`{ file, limit }`) instead of bare `File`s.
  - The rejection toast groups rejected files by the limit they exceeded and reports that limit, so media states its elevated per-type cap (2 GB) and documents state the generic cap (100 MB) — never a blanket 100 MB.
- `services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts`
  - The i18n `t` mock now echoes interpolation params so the reported `maxSize` is assertable.
  - Added two regression tests: a 3 GB media file's toast reports `2048` (MB) not `100`; and a policy-capped rejection reports the policy value.

## Verification

- `vitest --run --project client use-convex-file-upload.cap.test.ts` — 4 passed
- `tsc --noEmit` — clean (exit 0)
- `oxlint --type-aware` on changed files — 0 warnings, 0 errors

Closes #2048

### DIFF
--- a/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
+++ b/services/platform/app/features/chat/hooks/use-convex-file-upload.cap.test.ts
@@ -36,7 +36,12 @@ vi.mock('@/app/features/settings/governance/hooks/queries', () => ({
 vi.mock('@/app/hooks/use-toast', () => ({ toast: vi.fn() }));
 
 vi.mock('@/lib/i18n/client', () => ({
-  useT: () => ({ t: (key: string) => key }),
+  useT: () => ({
+    // Echo the key, appending interpolation params so tests can assert the
+    // reported limit (e.g. the `maxSize` in the file-too-large toast).
+    t: (key: string, params?: Record<string, unknown>) =>
+      params ? `${key}:${JSON.stringify(params)}` : key,
+  }),
 }));
 
 vi.mock('../utils/get-audio-duration', () => ({
@@ -136,5 +141,49 @@ describe('useConvexFileUpload — per-type size ceiling (#2048)', () => {
     expect(
       toastMock.mock.calls.some(([arg]) => arg?.title === 'invalidFiles'),
     ).toBe(true);
+  });
+
+  it('reports the elevated media ceiling (2048MB), not the 100MB generic cap, in the rejection toast', async () => {
+    const { result } = renderHook(() => useConvexFileUpload(config));
+
+    await act(async () => {
+      // 3 GB audio file — above the 2 GB media per-type cap, so it is rejected.
+      await result.current.uploadFiles([
+        makeAudioFile('huge.mp3', 3 * 1024 * MB),
+      ]);
+    });
+
+    expect(result.current.attachments).toHaveLength(0);
+    const tooLargeToast = toastMock.mock.calls.find(
+      ([arg]) => arg?.title === 'invalidFiles',
+    );
+    expect(tooLargeToast).toBeTruthy();
+    // The toast must state the media ceiling (2 GB = 2048 MB), not 100 MB.
+    expect(tooLargeToast?.[0]?.description).toContain('2048');
+    expect(tooLargeToast?.[0]?.description).not.toContain('100');
+  });
+
+  it('reports the governance policy cap in the rejection toast', async () => {
+    useUploadPolicyMock.mockReturnValue({
+      policyEnabled: true,
+      maxFileSize: 50 * MB,
+      documentMaxFileSize: 50 * MB,
+      allowedTypes: ['audio/mpeg'],
+      blockedExtensions: [],
+      allowedExtensions: [],
+    });
+
+    const { result } = renderHook(() => useConvexFileUpload(config));
+
+    await act(async () => {
+      await result.current.uploadFiles([
+        makeAudioFile('podcast.mp3', 150 * MB),
+      ]);
+    });
+
+    const tooLargeToast = toastMock.mock.calls.find(
+      ([arg]) => arg?.title === 'invalidFiles',
+    );
+    expect(tooLargeToast?.[0]?.description).toContain('50');
   });
 });
--- a/services/platform/app/features/chat/hooks/use-convex-file-upload.ts
+++ b/services/platform/app/features/chat/hooks/use-convex-file-upload.ts
@@ -101,7 +101,7 @@ export function useConvexFileUpload(config: ConvexFileUploadConfig) {
   const uploadFiles = useCallback(
     async (files: File[]) => {
       const validFiles: { file: File; resolvedType: string }[] = [];
-      const rejectedTooLarge: File[] = [];
+      const rejectedTooLarge: { file: File; limit: number }[] = [];
       const rejectedType: File[] = [];
       const rejectedAudioDuration: File[] = [];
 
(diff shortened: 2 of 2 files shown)


larryro:
## Desk review — PR #2255 (issue #2048): per-type ceiling in the chat upload size toast

**Verdict: READY TO MERGE.**

Reviewed the checked-out head ref `tale/xs7f35tr7xt6h3cag20qwrzqf589b180` (commit `0bccdb1bd`) via a two-round fan-out review (5 dimension reviewers + independent confirmation).

### Scope confirmed
The underlying rejection bug — `Math.min` clamping the 2 GB media ceiling back to the 100 MB generic cap — was already fixed and merged in **#2156** (`833565c0`), and that fix is present in this branch (`use-convex-file-upload.ts:139-142`: `policyLimits.policyEnabled ? Math.min(maxFileSize, typeCeiling) : typeCeiling`). This follow-up correctly completes #2048 by making the rejection **toast** report the per-type ceiling each file actually exceeded (grouping rejected files by limit) instead of always reporting 100 MB.

### CI
All required checks **green** — 46 pass, 5 skipping (Trivy / fork-PR / container jobs, conditionally skipped), 0 failing, 0 pending.

### Tests
Ran the project's own suite for the changed hook:
`bunx vitest --run --project client app/features/chat/hooks/use-convex-file-upload.cap.test.ts` → **4/4 passed**. New tests prove the toast reports 2048 MB for media (not 100) and the governance cap (50 MB) under policy.

### Findings
- **Correctness / robustness / elegance / boundaries:** no issues. The reported limit always equals the gate that rejected the file (`limit: perTypeLimit` is the same value compared); grouping emits at most 2 toasts (the only two distinct ceilings); `useCallback` deps unchanged-and-correct; no caller passes a custom `maxFileSize`, so no regression from media using `typeCeiling` when policy is disabled.

### Non-blocking suggestions (optional polish, not required to merge)
1. The mixed-rejection case (a document **and** a media file rejected together → two toasts with different numbers) — the whole point of the new grouping — is not directly tested, partly because the test stubs `detectMediaMime` to always return `audio/mpeg`. Consider a test that produces two distinct limits and asserts two toasts.
2. `expect(...).not.toContain('100')` (cap.test.ts:163) is a whole-string substring check; a media filename containing `100` (e.g. `clip-1080p.mp4`) would fail it spuriously. Asserting `'"maxSize":2048'` would be stronger and non-brittle.

These are minor test-quality nits; the two headline behaviors are proven and CI is green.

PR #2266, opened 2026-06-28
fix(platform): clear orphaned sharedWithTeamIds when a project goes org-wide (#2046)

### LINKED ISSUE #2046: Bug: Switching a project's owning team to "Org-wide" orphans sharedWithTeamIds — project stays team-restricted while the UI shows "Org-wide"
**Area / route**

Projects → Overview → Sharing. `/dashboard/{org}/projects/{projectId}` — `services/platform/app/features/projects/components/project-sharing-section.tsx` (+ server `services/platform/convex/projects/mutations.ts::updateProjectSharing`).

**Steps to reproduce**

1. As an org admin/owner, open a project owned by Team A that is also shared with Team B (`teamId = A`, `sharedWithTeamIds = [B]`).
2. In the Sharing section, change the "Owning team" Select from Team A to "Org-wide".
3. Confirm the narrowing-warning dialog (it appears because Team A loses access).

**Expected**

Either the project genuinely becomes org-wide (`sharedWithTeamIds` cleared so `getProjectTeamIds` returns `[]`), OR the still-shared teams stay visible/removable so the admin can see and edit the real audience. The displayed Visibility must match effective access.

**Actual**

`handleOwningTeamChange` computes `nextShared = sharedWithTeamIds.filter(id => id !== nextOwningTeam)`. Going org-wide makes `nextOwningTeam = null`, and `null` is never an element of the shared array, so no shared team is dropped. It saves `{ teamId: null, sharedWithTeamIds: [B] }`. The server stores this verbatim (`teamId` cleared, `sharedWithTeamIds = [B]`) — `validateSharing` only rejects when the *owning* team is in the shared list, which no longer applies. So `getProjectTeamIds()` returns `[B]` and `isOrgWideProject` is `false`: the project is actually restricted to Team B (plus admins), not org-wide.

On the re-rendered Sharing section the Select shows `value = teamId ?? NONE_OWNING_TEAM` = "Org-wide", and the "Also shared with" multiselect is hidden because its render guard is `teamId && shareableTeams.length > 0` (now `undefined && …` → falsy). The admin sees "Org-wide" selected with no control to view or remove Team B. Meanwhile the project overview header and the projects table both compute `teamCount = (teamId?1:0) + sharedWithTeamIds.length = 1` and show "1 team". The Sharing Select silently contradicts the overview/table, effective access disagrees with the displayed Visibility, and the orphaned share can never be cleared through the UI.

**Evidence** (verified)

- `services/platform/app/features/projects/components/project-sharing-section.tsx:142-145` — `sharedWithTeamIds.filter(id => id !== nextOwningTeam)` never drops a shared team when `nextOwningTeam` is `null` (org-wide).
- `:197` — `value={teamId ?? NONE_OWNING_TEAM}` renders "Org-wide" once `teamId` is cleared.
- `:203` — multiselect guard `teamId && shareableTeams.length > 0` hides the "Also shared with" control (and thus Team B) when `teamId` is null/undefined.
- `services/platform/convex/projects/mutations.ts:539-571` — `updateProjectSharing` patches `teamId=undefined` and `sharedWithTeamIds=[B]` verbatim; no normalization clearing shares when going org-wide.
- `services/platform/convex/projects/mutations.ts:245-260` — `validateSharing` only rejects when the owning team appears in the shared list; passes here.
- `services/platform/convex/projects/access.ts:36-49,54-55` — `getProjectTeamIds` returns `[B]` and `isOrgWideProject` is `false` with no owning team, so access is restricted to Team B.
- `services/platform/app/features/projects/components/project-overview.tsx:157-158,175-178` and `projects-table.tsx:135` — both show "1 team", contradicting the "Org-wide" Select.

**Notes**

Two-part fix candidate: (1) When `nextOwningTeam === null`, clear `sharedWithTeamIds` to genuinely make the project org-wide (and/or have the server normalize: dropping the owning team without re-targeting shares is the ambiguous case). (2) Show the "Also shared with" audience even when org-wide so any remaining shares are visible and removable. The narrowing-confirmation dialog already fires correctly here (Team A losing access is detected), but the post-confirm result is the buggy orphaned state.

---

larryro:
## Summary

Fixes #2046. Switching a project's owning team to **Org-wide** used to leave `sharedWithTeamIds` intact, so `getProjectTeamIds()` still returned those teams: the project stayed restricted to them while the Sharing Select, overview header, and projects table all displayed "Org-wide". The orphaned shares were also unreachable from the UI (the "Also shared with" control is hidden when there's no owning team).

A shared-with team is always *additional* to an owning team, so a project with no owning team cannot retain shares.

## Changes

- **Server** (`convex/projects/access.ts`, `mutations.ts`): new pure `normalizeSharing()` helper, applied in `updateProjectSharing` so an effective owning team of `null` clears `sharedWithTeamIds`. Authoritative backstop regardless of caller (also cleans legacy orphaned rows on next save).
- **Client** (`project-sharing-section.tsx`): `handleOwningTeamChange` now clears the shared list when the owning team becomes Org-wide, instead of `filter(id => id !== null)` which never matched.
- **Tests** (`access.test.ts`): unit coverage for `normalizeSharing`, including that the result is genuinely org-wide per `isOrgWideProject`/`getProjectTeamIds`.

## Verification

- `vitest --project server convex/projects/` — 38 passed
- `tsc --noEmit` — exit 0
- `oxlint --type-aware` on changed files — 0 warnings/errors

Closes #2046

### DIFF
--- a/services/platform/app/features/projects/components/project-sharing-section.tsx
+++ b/services/platform/app/features/projects/components/project-sharing-section.tsx
@@ -136,12 +136,15 @@ export function ProjectSharingSection({
   const handleOwningTeamChange = useCallback(
     (value: string) => {
       const nextOwningTeam = value === NONE_OWNING_TEAM ? null : value;
-      // Drop the new owning team from the shared list if it's there — mirrors
-      // the server-side rule that the owning team can't appear in
-      // `sharedWithTeamIds`.
-      const nextShared = sharedWithTeamIds.filter(
-        (id) => id !== nextOwningTeam,
-      );
+      // Going org-wide clears every share — a "shared-with" team is additional
+      // to an owning team, so without an owner those shares would silently keep
+      // the project restricted while the Select reads "Org-wide" (mirrors the
+      // server-side `normalizeSharing`). Otherwise drop just the new owning team
+      // from the shared list (the owning team can't also appear in it).
+      const nextShared =
+        nextOwningTeam === null
+          ? []
+          : sharedWithTeamIds.filter((id) => id !== nextOwningTeam);
       commit({ teamId: nextOwningTeam, sharedWithTeamIds: nextShared });
     },
     [commit, sharedWithTeamIds],
--- a/services/platform/convex/projects/access.test.ts
+++ b/services/platform/convex/projects/access.test.ts
@@ -5,6 +5,7 @@ import {
   getProjectTeamIds,
   hasProjectAccess,
   isOrgWideProject,
+  normalizeSharing,
 } from './access';
 
 describe('getProjectTeamIds', () => {
@@ -60,6 +61,45 @@ describe('isOrgWideProject', () => {
   it('returns false when project has any team', () => {
     expect(isOrgWideProject({ teamId: 'team-1' })).toBe(false);
     expect(isOrgWideProject({ sharedWithTeamIds: ['team-2'] })).toBe(false);
+  });
+});
+
+describe('normalizeSharing', () => {
+  it('clears shared teams when going org-wide (no owning team)', () => {
+    // The bug: switching the owning team to "Org-wide" used to keep
+    // `sharedWithTeamIds`, leaving the project restricted to those teams while
+    // the UI showed "Org-wide".
+    expect(normalizeSharing(null, ['team-b'])).toEqual({
+      teamId: null,
+      sharedWithTeamIds: [],
+    });
+  });
+
+  it('leaves an already org-wide project org-wide', () => {
+    expect(normalizeSharing(null, [])).toEqual({
+      teamId: null,
+      sharedWithTeamIds: [],
+    });
+  });
+
+  it('preserves shared teams when an owning team is set', () => {
+    expect(normalizeSharing('team-a', ['team-b', 'team-c'])).toEqual({
+      teamId: 'team-a',
+      sharedWithTeamIds: ['team-b', 'team-c'],
+    });
+  });
+
+  it('keeps an owning team with no shares unchanged', () => {
+    expect(normalizeSharing('team-a', [])).toEqual({
+      teamId: 'team-a',
+      sharedWithTeamIds: [],
+    });
+  });
+
+  it('result is genuinely org-wide per isOrgWideProject', () => {
+    const normalized = normalizeSharing(null, ['team-b']);
+    expect(isOrgWideProject(normalized)).toBe(true);
+    expect(getProjectTeamIds(normalized)).toEqual([]);
   });
 });
 
--- a/services/platform/convex/projects/access.ts
+++ b/services/platform/convex/projects/access.ts
@@ -56,6 +56,26 @@ export function isOrgWideProject(project: ProjectAccessInput | null): boolean {
 }
 
 /**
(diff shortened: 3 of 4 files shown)


larryro:
## Desk Review — Issue #2046 (orphaned `sharedWithTeamIds` when a project goes Org-wide)

**Verdict: READY TO MERGE.**

### CI
All required checks **pass** (Type check, Lint, Format, Unit, Build platform/convex, all 16 Playwright platform shards, UI, Browser, Migrations, CodeQL/Opengrep/Trivy scans). Only fork-PR-only and container-test jobs are `skipping`, which is expected on a same-repo branch. CI is confirmed green.

### Tests
Ran the project's own suite from `services/platform`:
`bunx vitest --run --project server convex/projects/` → **38 passed (3 files)**, including the 30 in `access.test.ts` with the 5 new `normalizeSharing` cases. Verified the round-trip test asserting the normalized result satisfies `isOrgWideProject` / `getProjectTeamIds`.

### What the fix does (and why it's correct)
- New pure helper `normalizeSharing(teamId, sharedWithTeamIds)` in `convex/projects/access.ts` returns `{teamId:null, sharedWithTeamIds:[]}` when there is no owning team, else passes through. This makes "Org-wide" genuinely org-wide so effective access (`getProjectTeamIds → []`, `isOrgWideProject → true`) matches the displayed visibility — exactly the resolution option (1) the issue called for.
- Server `updateProjectSharing` resolves requested shares (falling back to stored shares when the arg is omitted), normalizes, then writes a single declarative patch and logs accurate `nowOrgWide` audit metadata.
- Client `handleOwningTeamChange` mirrors the rule: clears all shares on the Org-wide transition, otherwise filters out the new owning team.

Walked every arg combination (teamId omitted/null/set × shares omitted/empty/non-empty). The reproduction path (Team A → Org-wide with `sharedWithTeamIds=[B]`) now persists `{teamId:undefined, sharedWithTeamIds:undefined}`. `validateSharing` still runs before normalization; `db.patch` is a shallow merge so always-writing the two fields does not disturb others. No regression: changing only the owner preserves existing shares; updating only shares preserves the owner.

1. **`createProject` / `duplicateProject` do not call `normalizeSharing`** (`mutations.ts:338`, `:1470`). A direct programmatic API call to `createProject` with `teamId` omitted + non-empty `sharedWithTeamIds` could persist the same orphaned shape, and `duplicateProject` copies source sharing verbatim. Neither is reachable from the create UI and both are pre-existing — outside this issue's scope — but applying `normalizeSharing` there too would close the class entirely (defense-in-depth).
2. **Legacy already-orphaned rows** (created before this fix) are not back-filled; they self-heal on the next sharing write but the multiselect guard (`teamId && …`) still hides their shares until then. No migration is included; the issue did not require one.
3. **No mutation-level or component test.** Only the pure helper is unit-tested; the mutation's arg-omitted fallback / audit-metadata branches and the client `handleOwningTeamChange` are exercised only indirectly. Acceptable given the semantic rule lives entirely in the tested helper, but a `convex-test` mutation test would guard against future regressions.

These are improvements, not blockers. The change is minimal, follows repo conventions, comments are accurate, and it genuinely fixes the reported bug.

PR #2303, opened 2026-07-01
feat(platform): add IMAP/SMTP mailbox integration for two-way email

Israeltheminer:
## What

Adds a generic **`imap_smtp`** integration so a private, self-hosted (non-Gmail) mailbox flows into **Conversations two-way**:

- **Receive** over IMAP (`imapflow` + `mailparser`) → Conversations
- **Send** replies over SMTP (`nodemailer`)

IMAP/SMTP are raw-TCP, and the connector sandbox is HTTP-only — so, following the **SQL integration precedent**, the protocol work lives in `convex/node_only/imap_smtp/*` node actions and dispatches through the existing integration action helpers.

## Highlights

- **Split credentials** — IMAP reads the mailbox (`basicAuth`); SMTP can send through a *separate* login (`smtpAuth ?? basicAuth`). Enables "read a private mailbox, **send via Resend**." `smtpAuth` is an **optional encrypted** field that mirrors the `basicAuth` pipeline end-to-end (validators → encrypt → action → mutation → load → resolve). Being optional, it is **data-safe growth — no migration**.
- **Multi-address** — one integration serves `support@`, `billing@`, `sales@`, `hiring@`, `privacy@`, `terms@`… Replies go out from whichever address the customer wrote to, *when it's on the configured From domain*; otherwise fall back to the configured From / SMTP login. (`conversations/reply_from.ts`, unit-tested.)
- **Scheduled sync** — a sync workflow on a cron trigger; **does not run on install**.
- **UI** — SMTP username/password fields and an inline From-address hint in the manage dialog, localized in **en/de/fr**.
- **Docs** — integrations + conversations overview pages updated in all three locales.

Secrets go through the same JWE `encryptString` path as every other integration credential — never in `connectionConfig` (plaintext), never in git.

## Testing

- `tsc --noEmit`: **0 errors**
- New unit tests: `reply_from` (multi-address / domain fallback), `imap_smtp_config` (split-cred resolution), `is_imap_smtp_integration` guard, sync-workflow shape, credentials-form component
- Regression: full `convex/integrations`, `convex/conversations`, integration action-defs suites green (157 tests); i18n parity (23); docs structural suite (144)
- `oxfmt --check`, `oxlint --type-aware`, opengrep SAST: clean
- **Observed end-to-end**: a real email to `billing@` synced into Conversations after a manual sync run.


Israeltheminer:
Resolving the four `github-code-quality` "Unknown directive: 'use node'" threads as false positives.

`'use node';` is the **required Convex runtime directive** — it marks a module to run in Convex's Node.js runtime instead of the default V8 isolate, which the IMAP/SMTP libraries (imapflow/nodemailer/mailparser) depend on. It's used in 228 existing files across `convex/` (e.g. the `node_only/integration_sandbox` and SQL-integration precedents). Removing it, as the bot suggests, would break these node actions. No code change.

PR #2324, opened 2026-07-02
feat(platform,sandbox): default sandbox Claude Code to Fable 5

yannickmonney:
## What

Defaults the in-sandbox Claude Code agent to **Claude Fable 5** and keeps it on Anthropic's most capable model automatically, completing the max-tuning started in #2271 (effort=max, 1M context, workflows, ultrathink).

- **Agent pin** (`builtin-configs/agents/chat/claude-code.json`): `supportedModels` moves from the static Opus 4.8 pin to `openrouter:~anthropic/claude-fable-latest` (OpenRouter's rolling alias — auto-tracks the newest Fable-class model) first, concrete `openrouter:anthropic/claude-fable-5` second.
- **Catalog** (`builtin-configs/providers/openrouter.json`): both entries added (1M context, 128K output, $10/$50 per M, chat+vision, qualityScore 0.98) with de/fr descriptions. The weekly sync would auto-add the concrete id eventually but can never add the `~` alias (its vendor prefix is outside the curated frontier set) — so both ship here.
- **Migration `0.2.89/03`** (node, reversible, snapshot-free): per org, appends the two catalog entries when absent and retargets every `agentKind: 'claude-code'` agent whose pin still equals the old shipped default. Operator-edited pins/entries survive both `up` and `down`; `down` removes only entries structurally identical to what `up` wrote.
- **Sandbox image**: new overridable env floor `CLAUDE_CODE_MAX_OUTPUT_TOKENS=64000` — Claude Code gives *recognised* flagship ids 64K output under `[1m]` but drops *unrecognised* ids (the rolling alias) to 32K; the floor restores parity. Deliberately not >64K: unverified against live upstreams at 1M context.
- **Docs**: models tables (en/de/fr) regenerated via `update-model-catalog.ts --docs-only`; the generator gains a `~anthropic` → Anthropic vendor label.

No Claude Code version bump: the pinned 2.1.173 binary already knows `claude-fable-5` (verified via `strings` + live smoke), so the daemon's verified stream-json/steer-hook behaviors stay untouched.

## Verification

Smoke-tested the **pinned 2.1.173 binary** against a local capture server posing as the API, for both `anthropic/claude-fable-5[1m]` and `~anthropic/claude-fable-latest[1m]`:

- request `model` arrives **bare** ([1m] stripped) → cannot 404 the per-turn VK allowlist
- `output_config: {"effort": "max"}`, `thinking: {"type": "adaptive"}`
- `anthropic-beta` carries `context-1m-2025-08-07` + `effort-2025-11-24` + interleaved thinking
- CC `modelUsage.contextWindow = 1,000,000`; `CLAUDE_CODE_MAX_OUTPUT_TOKENS` lands verbatim as `max_tokens`

Suites: migration tests (6, incl. idempotency + operator-edit guards + exact inverse), `lib/agent-adapters` + `convex/migrations` (137), `lib/shared` + `convex/providers` + `convex/agents` (1447), docs suite (144), `migrations:check`, typecheck, oxlint, oxfmt — all green. (Local `knip` is red on a clean `main` checkout in this environment too — pre-existing, untouched by this diff.)

## Notes

- Known pre-existing limitation: bifrost's openai-format hop clamps `effort max→high` toward OpenRouter (v1.5.13). Anthropic-direct orgs get true max. Not addressed here.
- Possible follow-up: raise the output floor to 128K for Fable once verified against the live upstream (OpenRouter advertises 128K completion at 1M context for fable-5).


yannickmonney:
Two follow-up commits on this branch address BYO routing and Fable usage rationing:

**BYO / OAuth sessions never see OpenRouter ids.** A BYO session talks directly to Anthropic, where the catalog refs (`openrouter:~anthropic/claude-fable-latest`, …) don't exist. The adapter maps exactly the shipped Fable default refs to Anthropic's native `claude-fable-5`; any explicitly specified model passes through byte-identical (no override).

**Fable usage is rationed → automatic Opus 4.8 fallback.** Per Anthropic's docs, Fable 5 has [limited included usage moving to usage credits](https://www.anthropic.com/news/redeploying-fable-5), and its safety classifiers reroute flagged requests. Claude Code ships two fallback mechanisms ([model-config docs](https://code.claude.com/docs/en/model-config)): the `--fallback-model` availability chain, and content-based classifier fallback that re-runs on the `ANTHROPIC_DEFAULT_OPUS_MODEL` slot — which we previously pinned to the primary, so the documented fallback would have looped onto the rationed model. Now: the Fable catalog entries carry `fallbackModelId: anthropic/claude-opus-4.8` (the existing failover field), the run scopes the VK to allow it, the adapter arms both mechanisms (FABLE slot stays on the primary — that's how CC identifies a gateway id as Fable), and Opus 4.8 is in `supportedModels` as the manual switch.

Smoke-verified on the pinned 2.1.173 binary: a failing Fable request flips the turn onto `anthropic/claude-opus-4.8` mid-turn and succeeds. Note for BYO subscription users: Fable usage surfaces in Claude Code via `/usage-credits` and limit notices — CC's native ids already enable its built-in fallback there.

yannickmonney:
Third commit (`3027d603d`) removes the hardcoding from the BYO mapping, per review feedback:

- New `nativeModelId` field on provider catalog entries (next to the existing `fallbackModelId`) — the vendor-native id behind a gateway-shaped id. All shipped Anthropic entries declare it; the rolling alias maps to `claude-fable-5` by data, not code.
- `run_external_agent` resolves it from the org's routing catalog for BYO sessions; a raw user-typed id matches no catalog entry and passes through unchanged. The adapter's hardcoded ref set is deleted — it now has zero model knowledge.
- The weekly model sync auto-derives the field for newly added Anthropic models (`anthropic/claude-x-y.z` → `claude-x-y-z`), so a future Fable 6 gets BYO support with no code change; org admins can also set/override it per entry like any other catalog field.

Everything model-specific now lives in config: the default pin (`supportedModels`), the fallback (`fallbackModelId`), and the BYO id (`nativeModelId`). Suites green (adapters 71, shared+providers+agents+migrations 1592, docs 144, `migrations:check` reports the schema addition as data-safe).

yannickmonney:
Fourth commit (`7133a99c2`): the weekly model sync now maintains the last piece of static data itself.

The sync engine gains a **RETRACK** step (after ADD): each rolling `~vendor/…-latest` alias entry's `nativeModelId` is re-pointed at the newest same-family concrete release in the config. Ordering matters — it runs after ADD, so the flagship added in the same run is already a candidate: when Fable 6 appears on OpenRouter, one sync run adds `anthropic/claude-fable-6` (native id derived), hides Fable 5, and moves the alias's BYO target to `claude-fable-6`. Managed sessions were already covered (OpenRouter resolves the alias upstream). Same 3-way rule as every synced field: an operator-edited value is never touched.

With this, nothing about the model line requires a code change anymore — managed tracking (gateway alias), BYO tracking (sync-maintained `nativeModelId`), fallback (`fallbackModelId`), and the pin itself (`supportedModels`) are all config the sync keeps fresh or org admins can edit.

PR #2397, opened 2026-07-02
fix(cli): unbreak the macOS install and first-run tale dev experience

yannickmonney:
## Why

A live walkthrough of the documented quickstart (tale.dev → docs → install → `tale init` → `tale dev`) found the self-serve path broken at two independent points:

1. **Every released macOS binary since the Bun 1.3.12 pin is killed by the kernel** (`Killed: 9`). Bun 1.3.12's `--compile` emits a truncated code signature (oven-sh/bun#29120, fixed in 1.3.13). CI's smoke test passes because GitHub's macOS runners run with SIP disabled; user machines SIGKILL the binary. The installer then reported **“Successfully installed tale”** anyway, because its `--version` probe swallowed the failure.
2. **`tale dev` never announces readiness.** The health probe fetches `https://localhost/health` with default TLS verification, which rejects the stack's self-signed certificate — so the READY block, browser auto-open, and admin key were unreachable on every run. On first runs the announcer also gave up after ~2 minutes (mid image-pull, which the docs budget at 5–10 minutes) with a false *“Services did not become healthy in time — check the log above”* — while pull progress was classified away, so the log above was empty.

## What

- **ci(cli):** Bun 1.3.12 → 1.3.13; add `codesign --verify --strict` for the macOS artifact so signature breakage fails the build instead of shipping.
- **fix(cli) installers:** `verify_installation` now fails with recovery hints when the binary can't run (both bash and PowerShell); unsupported architectures (Intel macOS, non-x86_64 Linux) are rejected up front with a build-from-source pointer; the Windows installer's next-step line no longer advertises the non-existent `tale setup`.
- **fix(cli) dev:** the health probe accepts the stack's own self-signed localhost certificate; the readiness watcher polls until compose exits instead of giving up, with a one-time first-run note.
- **fix(shared):** compose per-image pull milestones (`db Pulling` / `db Pulled`) classify as info so a first run shows signs of life; layer churn stays collapsed.
- **fix(cli) logs:** `tale logs platform` falls back to the colour-less dev container when no deployment exists, so the quickstart's troubleshooting command works on a dev-only stack.

## Testing

- `tools/cli`: `bun run typecheck` clean, `bun test` 244 pass / 0 fail.
- `packages/shared`: classifier suite 12/12 (new milestone cases included).
- Verified live on macOS 26 (arm64): released v0.2.97 binary is SIGKILLed with 'code object is not signed at all'; after `codesign --remove-signature && codesign -s -` the same binary runs — confirming the signature (not the code) is the killer, and that 1.3.13's fixed signer resolves it.
- Reproduced the silent first run + false health warning twice against ghcr with `tale dev` v0.2.97; confirmed `fetch` of the self-signed endpoint fails with `DEPTH_ZERO_SELF_SIGNED_CERT` under default TLS and succeeds with `tls: { rejectUnauthorized: false }`.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -92,15 +92,19 @@ jobs:
       - name: Setup Bun
         uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
         with:
-          bun-version: '1.3.12'
+          # 1.3.12's `bun build --compile` emits a truncated macOS code
+          # signature (oven-sh/bun#29120): users' kernels SIGKILL the binary
+          # while CI's SIP-disabled runners run it fine. Keep >= 1.3.13 and
+          # keep the "Verify macOS code signature" step below.
+          bun-version: '1.3.13'
 
       - name: Restore Bun install cache
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: ~/.bun/install/cache
-          key: bun-install-${{ runner.os }}-1.3.12-${{ hashFiles('bun.lock') }}
+          key: bun-install-${{ runner.os }}-1.3.13-${{ hashFiles('bun.lock') }}
           restore-keys: |
-            bun-install-${{ runner.os }}-1.3.12-
+            bun-install-${{ runner.os }}-1.3.13-
 
       - name: Install dependencies
         working-directory: tools/cli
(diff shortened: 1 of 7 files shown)


yannickmonney:
Closing as superseded. This branch was the original single-PR bundle of the getting-started fixes; it was later split into focused PRs, and two fixes were improved in the process, so the tip here is now behind `main` and its net diff would revert ~14k lines:

| This bundle's commit | Landed as | Notes |
|---|---|---|
| ci(cli): bump Bun to 1.3.13 + codesign verify | #2399 | plus a **codesign normalize** step — Bun signs *before* finalizing the binary, so the bump alone still failed `codesign --verify --strict` (the `Build (macos)` failure here) |
| fix(cli): fail installers on dead binary / unsupported arch | #2400 | patch-identical |
| fix(shared): surface per-image pull milestones | #2401 | patch-identical |
| fix(cli): dev readiness + trust local TLS | #2402 | **reworked** — gated on the platform container's Docker healthcheck instead of `rejectUnauthorized: false`, which the Opengrep gate rejected (the `Opengrep` failure here) and which also fixed a deeper flaw (`/health` is answered by the proxy alone) |
| fix(cli): tale logs dev fallback | #2403 | patch-identical |

All five behaviours are on `main` in equal-or-better form (`git cherry` confirms three are patch-identical; the other two are superseded by the improvements above). Nothing here is lost by closing.

PR #2513, opened 2026-07-06
fix(platform): SSO login for title-less Entra users + login-failure observability

larryro:
## Why

An Entra user with **no job title** could not sign in via SSO — they hit a generic red *"Single sign-on could not complete due to a server error"* page. The real cause was invisible: it only appeared as a `console.error` in the Convex logs (the callback catches the error and returns a 302, so it logs as a *success*), and never reached our error dashboard.

This PR fixes the bug and closes the observability gaps that hid it.

## What

**1. Root fix — `fix(platform): allow SSO login for Entra users without a job title`**
Microsoft Graph returns every `$select`ed field even when empty, so a title-less user comes back as `jobTitle: null` (not an omitted key). That `null` leaked through the Entra adapter into `handleSsoLogin`, whose `jobTitle: v.optional(v.string())` validator **rejects `null`** (Convex `optional` accepts `undefined`, not `null`) — throwing an `ArgumentValidationError` before the handler ran and failing the whole login. Normalised `null`/`""` → `undefined` at the adapter boundary, with a regression test through `entraIdAdapter.getUserInfo`.

**2. Frontend — `feat(platform): report error-level browser console logs to Sentry`**
Sentry captures thrown/unhandled errors by default but treats a plain `console.error(...)` as a *breadcrumb*, not an issue. Added `captureConsoleIntegration({ levels: ['error'] })` so deliberately-logged frontend failures surface as issues too. Scoped to `error` only to bound event volume (the GlitchTip org is quota-throttled).

**3. Backend — `feat(platform): audit SSO login failures for observability`**
The OIDC authorize/callback handlers catch server-side failures and redirect, so a failure never surfaces as a failed request. Both catch blocks now write a durable, org-scoped `auth`/`failure` audit row (`action: sso_login_failed`, real error + stage/provider in metadata) via the existing audit-log writer. Best-effort: no-ops when the org isn't resolved yet and never throws, so it can't mask the login error or break the redirect.

## Notes
- No new i18n keys — reuses `sso.errors.serverError`. The `sso_login_failed` audit action renders via the audit table's `defaultValue` fallback, consistent with the existing `sso_configure`/`sso_enabled` actions (none carry explicit labels).
- No schema/migration change — reuses the `createAuditLog` internal mutation and existing `auditLogs` table.
- Complementary (not in this PR): enabling the Convex dashboard **Sentry exception-reporting integration** would surface uncaught function throws (like this one) in GlitchTip with `func` tags.

## Testing
- `entra_id/adapter.test.ts`: `jobTitle: null` → `undefined`, empty string → `undefined`, real title preserved (would fail pre-fix).
- SSO authorize/callback handler suites: green (13 tests).
- `bunx tsc --noEmit` clean; `oxlint --type-aware` clean; SAST clean.

### DIFF
--- a/services/platform/app/router.tsx
+++ b/services/platform/app/router.tsx
@@ -80,7 +80,14 @@ if (sentryDsn) {
   Sentry.init({
     dsn: sentryDsn,
     release: getEnv('TALE_VERSION'),
-    integrations: [Sentry.tanstackRouterBrowserTracingIntegration(router)],
+    integrations: [
+      Sentry.tanstackRouterBrowserTracingIntegration(router),
+      // Sentry captures thrown/unhandled errors by default but treats a plain
+      // `console.error(...)` as a breadcrumb, not an issue. Promote error-level
+      // console calls to issues so deliberately-logged failures are visible too.
+      // Kept to `error` only (not `warn`) to bound event volume.
+      Sentry.captureConsoleIntegration({ levels: ['error'] }),
+    ],
     tracesSampleRate: getEnv('SENTRY_TRACES_SAMPLE_RATE'),
   });
 }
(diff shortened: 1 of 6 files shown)


larryro:
Superseded by #2514 — changes cherry-picked onto fix/platform-chat-nav-rail-flash.

PR #2539, opened 2026-07-07
test(platform): add docker:dev seeded-login path to manual harness

larryro:
## Summary

Makes the manual-test harness (`services/platform/tests/manual/`) runnable by an unattended AI tester. Follow-up to the AI-manual-testing discussion: the tester agent only tests (fix/PR/review stays with the issue desk), and its environment is the containerized `docker:dev` stack with the seeded dev login.

## Changes

- **`tests/manual/SETUP.md`** — new **mode C (`docker:dev`)**: app at `https://localhost` via the Caddy proxy (self-signed-cert handling), `curl -skf https://localhost/api/health` readiness probe, the seeded `dev@tale.test` / `TaleDev!Passw0rd` login (loopback-only guard referenced), and the `TALE_PROVIDER_KEY_OPENROUTER` forwarding contract — explicitly framed as operator prep, so "No API key configured" during a run is a reportable defect, not an environment note.
- **`tests/manual/scripts/save-auth-state.ts`** — `QA_AUTH_EMAIL` / `QA_AUTH_PASSWORD` sign in as an existing account (the mode C seeded login) instead of minting a fresh owner+org; skips the fixture-only `waitForSeededOrg` gate; `ignoreHTTPSErrors` for the self-signed docker:dev cert. Default sign-up behaviour unchanged.
- **`tests/manual/README.md`** — declares the user docs under `docs/en/platform/` the behaviour oracle for manual runs: app-vs-docs mismatch is a reportable defect, never a silent judgment call.
- **`tests/e2e/AGENTS.md`** — fixes the pointer to the deleted `qa-browser` skill (now `browse-web`) and the broken relative link to `tests/manual/`.

## Verification

- oxlint: 0 warnings / 0 errors (whole platform workspace); platform typecheck green; `format:check` green; pre-commit SAST 0 findings.
- All relative links in the touched docs resolved by hand.
- **Not verified live:** the new sign-in mode of `save-auth-state.ts` was not run against a booted docker:dev stack (image build too heavy for this session). It reuses the file's proven in-page-fetch pattern and the same Better Auth endpoint/org-id regex as `tests/e2e/helpers/auth.ts`; worth a one-off run on the next docker:dev boot.
- Known limitation (accepted for a QA script): `signInExisting` assumes the account owns an org — true for the seeder; an org-less account times out at `waitForURL`.

## Definition of done

- [x] Green gate — format / lint / typecheck pass locally
- [x] Security gate — pre-commit SAST clean
- [x] Tests — N/A: docs + a QA convenience script outside the vitest projects; no runtime code touched
- [x] Migration — N/A: no data-model or config-schema change
- [x] Locales — N/A: internal English-only test docs
- [x] Docs — the change is the docs; stale instruction pointers fixed (`tests/e2e/AGENTS.md`)
- [x] Accessibility — N/A: no UI
- [x] Sweep — `qa-browser` has zero remaining references; `save-auth-state` referenced only from SETUP.md (updated)
- [x] Observed — static gates observed; live-run caveat stated above
- [x] Commits — one atomic conventional commit, branched off main

### DIFF
--- a/services/docs/app/content/frontmatter.json
+++ b/services/docs/app/content/frontmatter.json
@@ -277,7 +277,7 @@
     "locale": "fr",
     "frontmatter": {
       "title": "Agents externes",
-      "description": "Agents de code intégrés (Claude Code, Cursor, Hermes Agent, Gemini CLI) qui s'exécutent dans un bac à sable isolé ; tu discutes directement avec eux pendant qu'ils modifient des fichiers, lancent des commandes et poursuivent le travail sur plusieurs tours."
+      "description": "Agents de code intégrés (Claude Code, Cursor, OpenCode, Hermes Agent, Gemini CLI) qui s'exécutent dans un bac à sable isolé ; tu discutes directement avec eux pendant qu'ils modifient des fichiers, lancent des commandes et poursuivent le travail sur plusieurs tours."
     }
   },
   "fr:platform/agents/delegation": {
@@ -1413,7 +1413,7 @@
     "locale": "en",
     "frontmatter": {
       "title": "External agents",
-      "description": "Built-in coding agents (Claude Code, Cursor, Hermes Agent, Gemini CLI) that run inside an isolated sandbox; you chat with them directly as they edit files, run commands, and continue the work across turns."
+      "description": "Built-in coding agents (Claude Code, Cursor, OpenCode, Hermes Agent, Gemini CLI) that run inside an isolated sandbox; you chat with them directly as they edit files, run commands, and continue the work across turns."
     }
   },
   "en:platform/agents/delegation": {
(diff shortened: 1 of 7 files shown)


larryro:
The failing required checks (Opengrep, Unit) were inherited from main at d0bb0fec7 — both broke there, not in this branch. Root causes and fixes, landed here to get CI green:

- **Unit** — `@tale/docs` `content-manifest.test.ts`: the external-agent doc edits in d0bb0fec7 landed without rerunning `build-search-index`, so the committed `frontmatter.json` no longer matches the on-disk frontmatter. Fixed by regenerating (`17231a700`).
- **Opengrep** — two stacked causes (`3d590410c`, `f139f881c`):
  1. With `--sarif-output` (CI always sets it for the Security tab), opengrep counts **nosemgrep-suppressed** findings toward `--error`'s exit code — verified locally on v1.22.0 and v1.25.0: the same tree scans to 1 finding without SARIF and 11 with it, and the SARIF marks 10 of them `suppressions`. The gate now runs sarif-free (suppressions honored); the SARIF report is produced by a separate non-blocking pass, run first so the Security tab is fed even when the gate fails.
  2. The one genuinely unsuppressed finding: `tale-gemini-run:137` (new in d0bb0fec7) — `subprocess.Popen(cmd)` with adapter-provided flags. List-form exec, no shell, launching the agent CLI inside the sandbox is the runner's purpose; annotated with the repo's inline-nosemgrep convention like the sibling launchers.

Verified locally: `bash tools/opengrep/run.sh` → 0 findings, exit 0; with `OPENGREP_SARIF_OUTPUT` → exit 0 and a SARIF containing all 11 findings, 11/11 marked suppressed; `bun run --filter @tale/docs test` → 166/166.

These two fixes also unblock every other PR based on current main.

PR #2584, opened 2026-07-08
feat(platform): explicit run_code modes with packages-only install

larryro:
## Summary

Replaces `run_code`'s `entryPath`/`steps`/`code` mutex with an explicit `mode` discriminator — `script | inline | install` — as strict schema variants, and ships the P0 reliability/governance fixes that came out of the plan review.

- **`mode: "install"`** (new): packages-only provisioning — no script, skips the empty-workspace gate and the output harvest, 120s default timeout. Installed packages persist for later `run_code` calls in the turn's session.
- **`steps` removed** (breaking for the tool schema only): multi-step work is sequential `run_code` calls against the turn-scoped session; the mode error message carries copy-paste examples for repair.
- **Strict variants are load-bearing:** the provider-facing schema is flattened for OpenAI (`flattenUnionSchema`), which advertises every field on every mode — non-strict objects would silently strip cross-mode keys (e.g. `code` alongside `mode: "script"`) and run something other than what the model asked. Strict variants reject with a repairable `unrecognized_keys` error. Snapshot test locks `required === ['mode']`.
- **P0 — install failures propagate:** `runStepsInSession` previously ignored pip/npm exit codes (a failed install was invisible on successful runs since stderr only surfaces on failure). Now: non-zero install → short-circuit, `INSTALL_FAILED` with the failing tool + exit code; installs get their own budget (`min(max(timeoutMs, 120s), 300s)`); `errorCode`/`errorMessage` threaded through `SessionExecResultShape` and the action's returns validator. Workflow + crawler callers never pass packages — regression-tested unaffected.
- **P0 — node packages now hit the policy gate:** `checkPackagesAgainstPolicy` only ever checked the python bucket. Runtime and the admin policy tester now share one `evaluatePackageAgainstPolicy` (case-insensitive base-name match, deny wins even in allowlist mode) — this also fixes two pre-existing admin/runtime semantic divergences.
- **Inline pre-flight (fail-open):** new `inline_language.ts` scores per-declared-language signals over literal-stripped code — `LANGUAGE_MISMATCH` for interpreter/syntax contradictions (the `import: command not found` classic, plus a stderr backstop hint), `PREFER_PACKAGES` for ad-hoc `pip install` / `npm install -g` (route through the policy-gated `packages` param; `-r`/`-e`/project-local npm stay allowed).

Second commit (bundled intentionally): `pendingBackgroundTasks` ledger on `sandboxSessionOps` + `isAgentActivelyWorking`/`isAgentLingeringSteerReady` chat helpers, so the Working pill / stop / streaming affordances persist while background tasks run and stdin-steer engages only on true idle.

## Definition of done

- [x] Green gate — format, oxlint (type-aware), tsc, full platform vitest suite (75k pass; 4 convexTest files are the known starvation flake — pass in isolation)
- [x] Security gate — opengrep SAST 0 findings (repo scan + pre-commit staged scans)
- [x] Tests — 76 targeted: strict-union schema (accept/reject/steps/cross-mode), policy node×python×allow/deny/case, install semantics incl. packages-only + timeout floor + short-circuit, inline heuristic TP/FP matrix, flatten snapshot, tool-description path/mode contract
- [x] Migration — N/A for the run_code commit (no persisted schema change); the sessions field is optional + `migrations:check` classifies it data-safe
- [x] Locales — N/A: all new strings are LLM-facing; `INSTALL_FAILED` canvas keys already exist in en/de/fr
- [x] Docs — sandbox design doc updated (`mode: "install"` shipped); docs/ has no run_code pages (verified); builtin-configs carry no steps/pip guidance (verified)
- [x] Accessibility — N/A: no UI structure change (admin route swaps internal logic only)
- [x] Sweep — error-code renames grep-verified single-file; `runCodeArgs`/`executeCodeInSession` consumers enumerated; e2e/examples/skills carry no old-arg references
- [x] Observed — schema + heuristic behavior exercised via live bun scripts; session semantics via mocked-wire tests; **browser E2E running now, evidence follows as a PR comment**
- [x] Commits — atomic conventional pair, branched off main

## Test plan

Browser E2E on the dev stack immediately after opening (scenarios: inline happy path, install-only success + failure, script+packages, PREFER_PACKAGES rejection + model adaptation, language-mismatch rejection, org policy deny incl. node bucket, admin tester parity). Results land as a comment on this PR.

### DIFF
--- a/services/platform/app/features/chat/components/chat-interface.tsx
+++ b/services/platform/app/features/chat/components/chat-interface.tsx
@@ -53,6 +53,8 @@ import {
   useChatAgents,
   useResolvedHumanInputRequests,
   useSessionProgress,
+  isAgentActivelyWorking,
+  isAgentLingeringSteerReady,
 } from '../hooks/queries';
 import { useArenaThreadSetup } from '../hooks/use-arena-thread-setup';
 import { useChatScroll } from '../hooks/use-chat-scroll';
@@ -339,8 +341,11 @@ export function ChatInterface({
   useReportServerNow(threadMeta?.serverNow);
   const { toClientEpoch } = useClockOffset();
   const isGenerating = threadMeta?.isGenerating;
-  const agentLingering = sessionProgress?.agentIdleAt != null;
-  const agentActivelyWorking = (isGenerating ?? false) && !agentLingering;
+  const agentActivelyWorking = isAgentActivelyWorking(
+    isGenerating,
+    sessionProgress,
+  );
+  const agentLingeringSteerReady = isAgentLingeringSteerReady(sessionProgress);
 
   // Bridge the active message's live `file_write` tool calls into the
   // StreamingToolContext for the canvas pane. See hook.
@@ -1458,7 +1463,7 @@ export function ChatInterface({
                 className="mx-auto w-full max-w-(--chat-max-width)"
                 placeholder={
                   queueModeActive
-                    ? agentLingering
+                    ? agentLingeringSteerReady
                       ? t('queue.placeholderIdle')
                       : t('queue.placeholder')
                     : isImageGenAgent
--- a/services/platform/app/features/chat/components/sandbox-chip.test.tsx
+++ b/services/platform/app/features/chat/components/sandbox-chip.test.tsx
@@ -18,7 +18,11 @@ vi.mock('../context/chat-layout-context', () => ({
   }),
 }));
 
-vi.mock('../hooks/queries', () => ({
+// Partial mock: hooks are stubbed, but pure helpers (isAgentActivelyWorking
+// & co.) stay real — the component's working-state derivation is behaviour
+// under test, not test plumbing.
+vi.mock('../hooks/queries', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../hooks/queries')>()),
   useChatAgents: () => ({
     agents: [{ name: 'claude-code', primaryBehavior: 'external-agent' }],
   }),
--- a/services/platform/app/features/chat/components/sandbox-chip.tsx
+++ b/services/platform/app/features/chat/components/sandbox-chip.tsx
@@ -26,6 +26,7 @@ import { cn } from '@/lib/utils/cn';
 import { useChatLayout } from '../context/chat-layout-context';
 import {
   useChatAgents,
+  isAgentActivelyWorking,
   useSessionProgress,
   useThreadSandboxState,
 } from '../hooks/queries';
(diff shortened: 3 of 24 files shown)


larryro:
## Browser E2E results (dev stack, live sandbox)

Drove the real chat UI at localhost:3000 through every mode scenario. **Two defects found and fixed during the run** (`ef0593dbc`, `6764af0ae`) — both are now locked by unit tests.

| # | Scenario | Result | Evidence (observed tool output) |
|---|----------|--------|--------------------------------|
| 1 | `mode:"inline"` happy path (python `print(6*7)`) | ✅ | stdout `42` in the result message, 1 tool call |
| 2 | `mode:"install"` packages-only + same-turn persistence | ✅ after fix | `run_code installed 1 package(s) in 1973ms. They persist for later run_code calls in this turn.` + pip's `Successfully installed cowsay-6.1` surfaced; follow-up inline `import cowsay` in the same turn succeeded (ASCII cow) |
| 3 | `INSTALL_FAILED` propagation (nonexistent package) | ✅ | `runErrorCode:"INSTALL_FAILED"`, `runErrorMessage:"pip install failed (exit 1)"`, `success:false`, fenced pip stderr (`No matching distribution found`) — errorCode flows through the returns validator live |
| 4 | `PREFER_PACKAGES` (inline bash `pip install requests`) | ✅ | Rejected pre-sandbox with the structured hint (mode-install example, `-r`/`-e`/local-npm carve-outs) |
| 5 | `LANGUAGE_MISMATCH` (python source declared bash) | ✅ | `\`language\` is "bash" but the snippet looks like python — it contains a Python import statement, \`print(…)\`…` — no auto-correction, no sandbox round-trip |
| 6 | `mode:"script"` + `packages` + harvest | ✅ | `/user/output/tux.txt` harvested, file card + canvas tree show it, stdout clean of pip noise |
| 7 | Org policy: admin tester + runtime node bucket | ✅ after fix | Tester: `Sharp@1.2.3` → `Base name: Sharp` → **Denied**; runtime: `PACKAGE_NOT_ALLOWED` / `node package "sharp" is explicitly denied by org policy` (this bucket was never checked before this PR) |
| 8 | Strict cross-mode rejection (`mode:"script"` + `code` + `language`) | ✅ | Tool-call layer returned `unrecognized_keys: ["language","code"]` as a repairable validation error — not silently stripped |

### E2E-found defects (fixed in this PR)

1. **Install-only on a fresh session hit `INVALID_CWD`** — nothing is staged, so `/user/code` didn't exist and runnerd rejected the install exec. Installs now run from `/user` (always exists); steps keep `/user/code`. Notably, the failure surfaced *as designed* — `INSTALL_FAILED` with runnerd's message — which is exactly the propagation this PR adds; pre-PR this would have been a silent fake success.
2. **Unscoped npm `name@version` bypassed the policy** — `packageBaseName` only stripped versions off scoped specs, so `Sharp@1.2.3` never matched a `sharp` deny entry (runtime gate and admin tester alike). The delimiter set now includes `@` for non-scoped specs.

Incidental observations: mid-turn model failover (gemini→deepseek, provider availability) recovered the turn and re-ran the tool cleanly; org policy restored to its pristine empty state after the test.

PR #2629, opened 2026-07-10
feat(platform): migrations framework revamp — version truth, re-homing, full-chain proofs

yannickmonney:
## Summary

Complete revamp of the versioned Convex migrations framework, plus a **version-truth layer** that proves data survives `0.2.84 → newest → 0.2.84` byte-for-byte — and that every migration (and every corpus row/file) lives at the version whose release actually shipped its change.

**Framework rewrite.** One `migration.ts` per folder via `define<Kind>Migration` factories; identity derived from the folder path; registries generated by `migrations:sync` (byte-drift-checked, `api.d.ts` module map maintained); runner-injected snapshot APIs; `downTable` for table moves; org-fleet/limits seams; deploy failures surfaced (grep-stable `[migrations][deploy-failure]`, entrypoint banner, `tale migrate status` FAILED section); sentinel-framed CLI transport; `gen:migration` plop scaffold; the dormant `@convex-dev/migrations` component removed. All 49 migrations ported verbatim.

**Testing.** Every migration carries a `defineMigrationTest` ritual through the real runner (real org fleet, real betterAuth component, true handler idempotency, digest-equal down). The chain suite runs all 37 runnable migrations up and back down per PR with frontier-by-frontier digest equality and re-up convergence. A container e2e replays the operator surface against the real compose stack (tier 1 per-PR path-filtered; tier 2 nightly real-old-image upgrade via `MIGRATIONS_E2E_FROM`).

**Version truth.** `testing/versions/` stores every released tag's ground truth (db schema fingerprint, org-config Zod schemas as JSON Schema, initialized-project scaffold; content-addressed, ~1.8 MB). The versions suite validates the seed corpus as a real 0.2.84 deployment, validates rows **and config files** after every migration version against that release's actual schemas, and — for each version — inits a fresh project `up --to X`, validates it as a release-X deployment, migrates to newest, and proves `down --to X` restores the at-X world byte-identically.

**What the truth layer forced/found:**

- **25 migrations re-homed** (0.2.88–0.2.93 blocks → 0.2.96 / 0.2.98 / 0.3.4, per schema-diff + first-shipping-tag evidence). A new `formerIds` mechanism keeps live deployments safe: apply actions adopt old-id ledger rows, read-only queries fold them, snapshot restores (table-rows, component, fs-tree) fall back to former-id captures. Persisted data markers stay frozen at their original names.
- **Corpus made era-true**: rows/files born after 0.2.84 moved to version-boundary injections (`world/injections.testkit.ts`, db + fs halves); chains stay injection-free and baseline-anchored.
- **A shipped bug in `config_fingerprint.ts`**: the annotation stripper deleted schema _properties_ named `description`/`title`/`default`/`id`, blinding the config drift gate to those fields; fixed, baselines refreshed, and the checkpoint store re-derived from each tag's real Zod schemas (`--slow-config`).
- Placement audit (`scripts/audit-migration-versions.ts`): **48 OK / 0 MISMATCH / 1 justified NO-SIGNAL**.

Docs: new `convex-migrations` skill (authoring, reversibility doctrine, version truth, corpus duty, operator surface), AGENTS.md, scripts README, and `docs/{en,de,fr}` upgrades pages for `tale migrate`.

## Checklist

- [x] `bun run check` passes (format, lint, typecheck, all tests). Caveat: on my machine the full-parallel run trips 5s-timeout flakes in a handful of long-standing platform tests on untouched code paths (each passes in isolation, 37/37; platform-only run: 75,924/75,925 with one known-flaky conversations timeout, 12/12 isolated).
- [x] `bun run lint:sast` passes (Opengrep) — 232 rules, 0 findings (it caught a script-injection surface in the new `migrations-e2e.yml`, fixed in `ci(e2e)`).
- [x] Translations — **N/A** for `messages/*.json` (no UI strings); the `docs/{en,de,fr}` upgrades pages ship all three locales.
- [x] Docs updated (`docs/{en,de,fr}/self-hosted/operate/upgrades.md`, convex-migrations skill, AGENTS.md, `services/platform/scripts/README.md`).
- [x] `README.md` / `README.de.md` / `README.fr.md` — **N/A** (no root-README-documented behaviour changed).

## Test plan

- `bun run --filter @tale/platform migrations:check` — registries current, corpus + checkpoint guard, both snapshot gates.
- `cd services/platform && bunx vitest --run --project server convex/migrations` — 67 files / 280 tests: framework unit suite, all 49 per-migration rituals, chains A/B/C, the versions suite (boundary walk incl. config files, init-at-version round trips, scaffold integrity), corpus smoke.
- `bun services/platform/scripts/audit-migration-versions.ts` — placement verdicts with schema-diff + ship-tag evidence.
- Container e2e: runs in CI only (`migrations-e2e.yml`; **never run locally beside a dev stack** — compose pins `tale-*` container names). Tier 2 (real old-image upgrade incl. push-time validation over old rows) is nightly/dispatch and has not yet had a run — first nightly is the watch item.
- Review starting points: `framework/ledger.ts` (`reconcileAliases`), `framework/runner.ts` (`snapshotPageFor`), `framework/snapshot_store.ts` (former-id sidecar fallback), `testing/versions.test.ts`.


yannickmonney:
**Rebased onto main** (the customers+vendors → contacts merge, #2622/#2618) — beyond mechanical conflicts, the rebase absorbed the contacts work into the framework:

- Main's `0.3.4/06_customers_vendors_to_contacts_teardown` (old `meta.ts`/`index.ts` shape + hand-registry entry) is ported to the define API as **`0.3.4/29`**.
- **New runnable migrations `0.3.4/27` + `0.3.4/28`**: the release drops `conversations.customerId` / `supportCases.customerId` from the schema, and Convex validates existing rows at push time — a `reference` migration cannot execute that clear. The new migrations unset `customerId` on repointed rows; their `down` restores the original id **byte-for-byte from the contact `metadata.__migratedFrom` stamp** (proven by the harness digest ritual + the chain). Rows whose customer vanished before the backfills are left untouched rather than destroyed.
  - ⚠️ Release-process note: on a real upgrade the schema push happens **before** `migrations:runAll`, so deployments still carrying `customerId` rows would reject the push — the field-drop wants an expand→migrate→contract release split, or these clears shipped one release ahead of the schema drop. The tier-2 e2e (real old-image upgrade) is the place this shows up.
- The versions suite now distinguishes **dropped-table residue** (customers/vendors rows persist as undeclared-table data — legitimate on real deployments) from never-declared tables (mis-homing evidence), via the cumulative declared-set over the checkpoint history.
- `migrations:sync` now owns the **entire `migrations/**` section of `api.d.ts`** (framework + testing + versions): this rebase's conflict resolution silently dropped main's `contacts` entries and the branch's `testing/support` entry from the generated file — the maintainer now self-heals it.
- Registries, both snapshot baselines, and the dev checkpoint were regenerated on the merged tree.

Post-rebase verification: `migrations:check` green (52 folders, 39 runnable corpus-covered), migrations tree **70 files / 287 tests green**, placement audit **51 OK / 0 MISMATCH / 1 justified NO-SIGNAL**, lint + typecheck green.

Issue #2018, opened 2026-06-23
Bug: organizations/delete_cleanup.ts + users/create_member.ts throw raw Error — org/user lifecycle mutations are opaque


yannickmonney:
## Area

`services/platform/convex/organizations/delete_cleanup.ts`, `users/create_member.ts`, `users/create_user_without_session.ts`.

## Steps to reproduce

1. Attempt to delete an organization as a non-owner.
2. Attempt to delete the default organization.
3. Try to create a member who already exists in the org, or provide an already-used email.
4. All rejections surface as opaque \"Server Error\".

## Expected

All user-facing rejections should be `ConvexError({ code: ... })` with codes like `UNAUTHENTICATED`, `FORBIDDEN`, `DUPLICATE_MEMBER`, etc.

## Actual

`organizations/delete_cleanup.ts`:
- `line 30` — Unauthenticated
- `line 37` — Only owners can delete organizations
- `line 42` — The default organization cannot be deleted

`users/create_member.ts`:
- `line 49` — Unauthenticated
- `line 72` — Only admins can create members
- `line 123` — User is already a member of this organization
- `line 185` — Failed to create user account

`users/create_user_without_session.ts`:
- `line 64` — A user with this email already exists
- `line 81` — Failed to create user account

## Evidence

- `services/platform/convex/organizations/delete_cleanup.ts:28-43`
- `services/platform/convex/users/create_member.ts:47-73`, `:121-123`, `:183-186`
- `services/platform/convex/users/create_user_without_session.ts:64-81`

## Notes

\"User is already a member of this organization\" and \"A user with this email already exists\" are particularly important — these are validation errors the frontend member-invite form should display inline.

Issue #1470, opened 2026-04-12
Bug: The password field is not marked required but the form is failed to submit when that field is empty


MohRamadan24:
### What happened?

When the user trying to add new member manually form the tale platform, the password field is not marked required, but the form is failed to submit when that field is empty.

### Steps to reproduce

1. Login to tale
2. Navigate to setting page
3. Open organization tab
4. Add new member
5. Leave the password field empty, and then submit the field

### Expected behavior

The password should be required, and when all the required field is filled correctly, the submit feature should be works without any unclear error information.

### Affected service

Platform

### Version or commit

0.2.26

### Relevant log output

```shell

```

### Screenshots or recordings

<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/94aadc95-e7e9-42ef-9f92-7286bd2f3b1f" />

### Before submitting

- [x] I searched existing issues and this hasn't been reported yet.

PR #2697, opened 2026-07-13
fix(platform): disable add-member submit until password is set

### LINKED ISSUE #2687: Improvement: Add-member dialog should disable submit while Password is empty for a new user (validation parity)
## Problem

The Add-member dialog (Settings → Organization) keeps its submit button **enabled** while Password — required for a brand-new email — is empty, surfacing "Password is required to create a new user" only after the click. Every other validated form in the area (create-org wizard's Next, change-password's submit) proactively disables submit until valid, so this dialog is an inconsistency that costs users an extra failed round trip.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e): Add member with Name + new Email + empty Password → button clickable → inline error only after click.

### Code pointers

- Add-member dialog under `services/platform/app/features/settings/organization/`

## Proposed change

Disable submit while a required field is empty for the new-user case, matching the sibling forms' proactive validation.

## Acceptance criteria

- [ ] Submit disabled until Name/Email/Password valid for a new user; existing-user path unaffected

Related: #1943 (closed — unify validation timing), #1966 (closed — settings & forms UX polish epic).

Israeltheminer:
## Summary
- Disable the Add member submit button for new users until a non-empty password is entered, even when `formState.isValid` is true (password is optional in zod until submit).
- Add regression test (#2687) asserting submit stays disabled when name and email are filled but password is empty.
- Update related tests (#1470, #2018) to fill a policy-compliant password so they still exercise backend error handling.

Fixes #2687

## Test plan
- [x] `cd services/platform && bunx vitest --run --config vitest.ui.config.ts member-add-dialog.test`
- [ ] Open Settings > Members > Add member; fill name and email only — submit stays disabled until password is entered
- [ ] Existing-user email path still allows submit without password

### DIFF
--- a/services/platform/app/features/settings/organization/components/member-add-dialog.test.tsx
+++ b/services/platform/app/features/settings/organization/components/member-add-dialog.test.tsx
@@ -6,6 +6,8 @@ import { render, screen, waitFor } from '@/tests/utils/render';
 
 import { AddMemberDialog } from './member-add-dialog';
 
+const VALID_NEW_USER_PASSWORD = 'ValidPass12!';
+
 const { createMemberMock, userExistsMock } = vi.hoisted(() => ({
   createMemberMock: vi.fn(),
   userExistsMock: vi.fn(() => false),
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Closing as superseded: the #2687 fix landed on `main` via #2705 with a stricter implementation (`isPasswordValid` policy gating + expanded tests). This branch would regress that work on merge.

PR #2702, opened 2026-07-13
fix(platform): skip covering-gate checks in container catalog validation

### LINKED ISSUE #2675: Bug: validateBuiltinCatalog covering-gate check always fails inside the container
## Problem

The post-deploy builtin-catalog sanity check logs two ERROR lines and returns `ok: false` on **every** container boot, even when the catalog is perfectly healthy. A permanently red safety net masks real catalog regressions and trains operators to ignore the log line.

## Current behaviour

`validateBuiltinCatalog` (`services/platform/convex/lib/config_store/validate_builtin_catalog.ts`) calls `checkValidatorRegistryComplete()` from `services/platform/lib/shared/config/catalog_validator.ts`, which verifies that every `external-gate` domain's covering test files still exist under `REPO_ROOT` (derived from `import.meta.url`). Inside the shipped image there is no repo checkout and test files are never bundled, so the `automations` domain's check always reports:

```
[validateBuiltinCatalog] 2 issue(s) in the builtin catalog at /app/builtin:
  - automations: covering gate services/platform/convex/workflow_engine/helpers/validation/builtin_apps.test.ts is gone — restore it or fold the domain into this suite
  - automations: covering gate services/platform/convex/automations/fixture_bundle_drift.test.ts is gone — restore it or fold the domain into this suite
```

Repro: boot any container stack (e.g. the Migrations E2E job — see the `tale-platform` service log in https://github.com/tale-project/tale/actions/runs/29173794639/job/86599207511, also present on `main`) and read the entrypoint's "Validating builtin config catalog" section. Both test files exist in the repo; the check is meaningful only where a repo checkout exists (the vitest gate `builtin_configs_validation.test.ts` and the build-time `scripts/validate-builtin-configs.ts` call sites), never at runtime.

## Proposed change

Skip the covering-gate existence check in the runtime call site — e.g. only run it when the resolved `REPO_ROOT` actually looks like a checkout (`existsSync(join(REPO_ROOT, 'package.json'))`), or give `checkValidatorRegistryComplete()` an option the runtime action sets. Registry completeness (domain ↔ validator mapping) should still be checked everywhere; only the file-existence half is checkout-bound.

## Acceptance criteria

- [ ] A healthy container boot logs `builtin catalog OK` and returns `ok: true` from `validateBuiltinCatalog`.
- [ ] The vitest and build-time gates still fail when a covering-gate test file is deleted or moved without updating `DOMAIN_VALIDATORS`.
- [ ] A genuinely broken builtin catalog still turns the runtime check red.

Israeltheminer:
## Summary
- Add optional `checkCoveringGates` parameter to `checkValidatorRegistryComplete` (default `true`) so callers can skip repo-relative existence checks for externally-gated domains like `automations`.
- Call `checkValidatorRegistryComplete({ checkCoveringGates: false })` from the runtime `validateBuiltinCatalog` action — container images ship no vitest sources, so covering-gate paths always appeared missing.
- Add regression test proving `checkCoveringGates: false` skips automations gate file checks while the default still reports them.

Fixes #2675

## Test plan
- [x] `bun run --filter @tale/platform test -- catalog_validator builtin_configs_validation`
- [ ] Post-deploy `validateBuiltinCatalog` in a container no longer logs automations covering-gate false positives

### DIFF
--- a/services/platform/convex/lib/config_store/validate_builtin_catalog.ts
+++ b/services/platform/convex/lib/config_store/validate_builtin_catalog.ts
@@ -43,7 +43,9 @@ export const validateBuiltinCatalog = internalAction({
       return { ok: false, issueCount: 0, filesValidated: 0 };
     }
 
-    const registryIssues = checkValidatorRegistryComplete();
+    const registryIssues = checkValidatorRegistryComplete({
+      checkCoveringGates: false,
+    });
     const { issues: catalogIssues, filesValidated } = validateConfigDir(
       builtinDir,
       'catalog',
--- a/services/platform/lib/shared/config/catalog_validator.test.ts
+++ b/services/platform/lib/shared/config/catalog_validator.test.ts
@@ -1,0 +1,45 @@
+/**
+ * Regression for #2675 — runtime container images have no vitest sources, so
+ * `validateBuiltinCatalog` must skip repo-relative covering-gate existence
+ * checks while CI/build gates keep them enabled.
+ */
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+const AUTOMATIONS_GATE_FRAGMENTS = [
+  'builtin_apps.test.ts',
+  'fixture_bundle_drift.test.ts',
+] as const;
+
+vi.mock('node:fs', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:fs')>();
+  return {
+    ...actual,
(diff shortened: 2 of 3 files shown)


Israeltheminer:
Closing as superseded: the #2675 `checkCoveringGates: false` runtime fix landed on `main` via #2705 (with `repoRoot` injectable for tests). Rebase onto current `main` conflicts because the fix is already there.

Issue #2693, opened 2026-07-13
Bug: workflow step side panel "Save" always fails — handleSave is a TODO stub


yannickmonney:
## Problem

The workflow canvas step-config side panel renders a functional-looking **Save** button, but its handler never persists anything — every click fires a destructive "save failed" toast.

## Current behaviour

- `services/platform/app/features/workflows/components/workflow-sidepanel.tsx` (~line 220): `handleSave` is a stub behind `// TODO: Replace with file-based workflow save` — it unconditionally toasts `sidePanel.stepSaveFailed` / `sidePanel.fileBasedStepSaveUnsupported`.
- `isSaving` is a `const false`, and the footer `<Button><Save/>Save</Button>` (~line 134) sits outside the unified `useActiveEditor`/`EditorActions` save model every other config surface uses.
- The step's `JsonInput` config editor renders without a label/id (~line 95), unlike every other JsonInput in the workflow/automation config forms.

Repro: open a workflow's Editor tab → click a step node → edit the step config JSON → **Save** → destructive toast, nothing persisted.

## Proposed change

Implement file-based step-config persistence (write the step's config through the same workflow-file save path the Configuration tab uses, with `expectedHash` conflict handling), route it through an `EditorController` so Save/Discard docks in the header cluster like the Configuration tab, and label the JsonInput. If step-level editing is intentionally unsupported for file-based workflows, the panel should render the config read-only instead of a Save button that always fails.

Related: #2045 (canvas add-step/connect/delete no-ops — same read-only-canvas family; this is the remaining step-CONFIG stub).

## Acceptance criteria

- [ ] Editing a step's config and saving persists to the workflow file (or the panel is read-only with no Save affordance).
- [ ] Save/Discard for the panel participates in the unified editor model (header cluster), not a bespoke footer button.
- [ ] The step JsonInput has an accessible label.
- [ ] A failed save surfaces the real error; a successful one updates the canvas without a reload.

Issue #2684, opened 2026-07-12
Bug: docs agents/webhook-triggers.md still documents the removed publish gate


yannickmonney:
## Problem

`docs/en/platform/agents/webhook-triggers.md` (lines ~20–24) still says webhooks require a **published** agent ("an unpublished one shows *Publish this agent to enable webhook access*"). The publish gate was removed as dead code in #2359 — every agent is live in the file-based architecture, and no publish control exists anywhere in the UI.

## Current behaviour

Verified (2026-07-12, main @ 5605bf56e): repo-wide grep for `isPublished` / "Publish this agent" in the agents feature returns nothing; the manual guide (`tests/manual/agents.md` F11) itself documents "publish gate (none exists — by design)". The docs page contradicts both.

### Code pointers

- `docs/en/platform/agents/webhook-triggers.md` (+ `fr`/`de` locales)

## Proposed change

Delete the publish-gate paragraph; describe the actual behaviour (webhook available once created/enabled). Sweep locales.

## Acceptance criteria

- [ ] No mention of publishing in webhook-triggers.md across en/de/fr; described flow matches the live UI

Related: #2359 (closed — removed the dead publish gate).

Issue #2683, opened 2026-07-12
Bug: docs agents/create.md worked example "marketing/seo-writer" is rejected by the create dialog slug validation


yannickmonney:
## Problem

`docs/en/platform/agents/create.md` Step 1 tells users to "add a `/` to file the agent in a folder, e.g. `marketing/seo-writer`" — but the create dialog now **deliberately rejects `/`** in Name ("Must start with a letter or number and use only lowercase letters, numbers, hyphens, and underscores"), a guard added to close the broken-folder bug (#2337). The docs' own worked example cannot be typed.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e): Create agent → Blank → Name `marketing/seo-writer` → Continue stays disabled with the pattern error.

### Code pointers

- `services/platform/app/features/agents/components/agent-create-dialog.tsx` (~line 195) — zod schema forbids `/`, with a comment referencing the folder-creation bug
- `docs/en/platform/agents/create.md` — Step 1 folder guidance (also `fr`/`de` locales of the page)

## Proposed change

Update the docs to the supported way of organizing agents (or state that folders aren't user-creatable from the dialog); sweep all locales.

## Acceptance criteria

- [ ] create.md no longer instructs a slash-named agent; example is creatable as written, in en/de/fr

Related: #2337 (closed — slash-named agents unreachable; the validation this collides with).

Issue #2681, opened 2026-07-12
Bug: logout confirmation dialog does not return focus to its trigger on close


yannickmonney:
## Problem

Closing the logout confirmation dialog (Escape or Cancel) drops keyboard focus on `<body>` instead of returning it to the triggering control — a keyboard user loses their place after declining to log out. The sibling dialogs (Change password, Add member) return focus to their triggers correctly, so this one dialog is the outlier (WCAG 2.4.3 focus order).

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e):

1. Open the user menu ("Manage account") → **Log out** → confirm dialog appears.
2. Press Escape (or click Cancel).
3. `document.activeElement` is `<body>`. Contrasted in the same session: Change-password and Add-member dialogs both restore focus to their trigger buttons.

### Code pointers

- The logout confirm dialog in the user-menu feature (`auth.userButton.*`) — likely missing the focus-return wiring the other dialogs get from the shared dialog component because the trigger (menu item) unmounts with the menu

## Proposed change

Return focus to the user-menu trigger ("Manage account" button) on close, matching the other dialogs.

## Acceptance criteria

- [ ] Escape and Cancel both land focus back on the user-menu trigger
- [ ] Covered by the keyboard e2e or a component test

Evidence: `services/platform/tests/screenshots/2026-07-12_r1/auth/a5-after-cancel-focus-on-body.png` (local QA run).

PR #2705, opened 2026-07-13
fix(platform): resolve issues #2675-#2688 + unify settings/metrics ui

### LINKED ISSUE #2675: Bug: validateBuiltinCatalog covering-gate check always fails inside the container

### LINKED ISSUE #2676: Bug: provisioning probe counts .gitkeep-only catalog domains — every hermetic org shows a permanent "didn't finish setting up" banner and Retry falsely reports success
## Problem

Every organization — including a freshly created, fully seeded one — permanently shows the owner-facing warning banner **"This workspace didn't finish setting up — Part of its built-in configuration is missing, so agents and chat may not work yet."** in any environment whose builtin catalog contains domains with only hidden files (the e2e/manual fixture catalog's `branding/`, `governance/`, `skills/` hold only `.gitkeep`). **Retry setup** toasts success but never clears the banner, because the repair can't seed files that don't exist.

The probe and the scaffold disagree about what "seedable content" means: the probe counts `.gitkeep`, the scaffold copies nothing.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e, mode A per `tests/manual/SETUP.md`):

1. Mint any org (wizard or `save-auth-state.ts`) — scaffold succeeds, the seeded `E2E Assistant` appears, chat works.
2. Open `/dashboard/{org}/chat` as owner → the provisioning banner renders on every page.
3. Click **Retry setup** → success toast; banner still present after refetch and after reload. Reproduced on 3 pre-existing orgs and 1 brand-new org.
4. Filesystem: `default/branding|governance|skills/` contain only `.gitkeep`; every scaffolded org dir has no `branding/`, `governance/`, `skills/` at all.

### Code pointers

- `services/platform/convex/organizations/scaffold.ts:873` — `listMissingScaffoldDomains` treats a domain as seedable when `dirHasFiles(sourceDir)` is true
- `services/platform/convex/organizations/scaffold.ts:118` — `dirHasFiles` counts **any** non-transient entry, including dotfiles like `.gitkeep`
- `services/platform/convex/organizations/actions.ts` — `retryProvisioning` returns `ok: true` (nothing failed) even though the missing domains remain missing, so the UI toasts `toast.retrySucceeded`
- Fixture catalog: `services/platform/tests/e2e/fixtures/config/default/{branding,governance,skills}/.gitkeep`

## Proposed change

Make the probe and the scaffold agree on "seedable": `listMissingScaffoldDomains` (or `dirHasFiles` for the source side) should ignore hidden/dot entries so a `.gitkeep`-only domain counts as "nothing to seed from". Additionally, `retryProvisioning` should not report `ok` when the post-retry probe still lists missing domains — the success toast is currently unearned.

## Acceptance criteria

- [ ] A catalog domain containing only dotfiles (`.gitkeep`) is not counted as seedable; orgs scaffolded from the e2e fixture catalog show no provisioning banner
- [ ] After a Retry that cannot repair, the UI does not toast success
- [ ] A regression test covers the `.gitkeep`-only-domain case for `listMissingScaffoldDomains`

Related: #2636 (closed — introduced the banner + retry), #2675 (open — adjacent builtin-catalog gate issue in containers).

### LINKED ISSUE #2677: Bug: uploaded Canvas files are always download-only — renderHint hardcoded to attachment overrides the viewer type inference
## Problem

Any non-image file a **user uploads** in chat is permanently rendered in the Canvas pane as the download-only "can't preview" view — no Source/Preview toggle, no rendered markdown/Mermaid/SVG/HTML, no Copy — even for tiny, correctly-typed text files. The docs promise the opposite ("toggle between Source and Preview", `docs/en/platform/chat/canvas-pane.md`), and agent-written files of the same types render fine.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e, mode A):

1. Attach `notes.md` (or `.mmd`, `.svg`, `.html`, `.csv`, `.py`) to a chat message and send.
2. Open the Canvas pane and select the uploaded file.
3. Viewer shows only the path, `{size} · {contentType}` and a bare **Download** button. Reproduced for md/csv/mmd/svg/html/py, including a fresh correctly-typed `text/markdown` file.

### Code pointers

- `services/platform/convex/lib/attachments/workspace_uploads.ts:139` — every `user_upload` row hardcodes `renderHint: contentType.startsWith('image/') ? 'image' : 'attachment'`
- `services/platform/app/features/workspace/components/file-viewer-router.tsx:63` — `resolveKind` returns a server hint unconditionally when present, so `'attachment'` bypasses the extension/contentType inference that agent-written files (hint undefined) correctly fall through to

## Proposed change

Stop forcing `'attachment'` for user uploads: either leave `renderHint` undefined for non-image uploads (let `resolveKind`'s inference decide, as for agent files), or compute the same inference server-side at filing time.

## Acceptance criteria

- [ ] An uploaded `.md` shows Source/Preview with rendered markdown; `.mmd`/`.svg` render; `.html` renders in the sandboxed iframe; code/text files show source with Copy
- [ ] Image uploads keep rendering as images; genuinely binary uploads keep the download-only view
- [ ] Covered by a component/e2e test for at least one uploaded text type

Evidence: screenshots under `services/platform/tests/screenshots/2026-07-12_r1/workspace/` (local QA run).

### LINKED ISSUE #2678: Bug: agent editor Display name/Description snap back to the legacy value on clear and concatenate retyped text
## Problem

In the agent editor's General tab, clearing **Display name** (select-all + Backspace) instantly snaps the field back to the stale legacy value, and anything typed afterwards is **appended to the old name** instead of replacing it. The obvious rename gesture produces garbage like `E2E AssistantQAX`. **Description** shares the identical code path.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e):

1. Open a seeded agent's General tab (`/dashboard/{org}/agents/assistant`).
2. Click Display name, select all, Backspace → field immediately shows `E2E Assistant` again.
3. Type `QAX` → field shows `E2E AssistantQAX`.

### Code pointers

- `services/platform/app/routes/dashboard/$id/agents/$agentId/index.tsx:129` — `writeFields` sets `next.displayName = v ? patch.displayName : undefined;` — an empty input becomes `undefined`, and the rendered value then falls back to the legacy top-level `config.displayName` for the default-locale tab, so the cleared state can never be represented while typing

## Proposed change

Track the in-progress edit as the field's live value (empty string included) and only normalize empty→undefined at save/commit time, so clear-then-retype behaves like a normal input.

## Acceptance criteria

- [ ] Select-all + delete leaves the field visibly empty; typing replaces rather than appends
- [ ] Same for Description
- [ ] Regression test for the clear-then-retype flow

Related: #1941 (closed — Display name rename/auto-derive work in the same form).

### LINKED ISSUE #2679: Bug: BYO "Provider model id" typed but not added is silently discarded — Save commits an empty model list
## Problem

On the agent Credentials form (Bring your own credentials → Models), a model id **typed into "Provider model id" but not explicitly committed** (Enter or **+ Add model**) is silently thrown away: typing alone never dirties the form, and when Save is enabled by an unrelated pending change, saving commits an **empty model list** — no error, no warning. The manual guide's own F19 steps ("fill Provider model id → Save") lose the value.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e):

1. Switch an external/coding agent's Credentials to BYO.
2. Type `claude-opus-4-20250514` into Provider model id; do not press Enter.
3. Click Save (enabled by the credentials-mode change) → config persists with no models; the typed text vanishes. Committing via Enter/+ Add model works and persists correctly.

### Code pointers

- Agent editor Instructions/Credentials form — the BYO models list input (free-text + "+ Add model") under `services/platform/app/features/agents/` (the pending input value is not treated as part of the form state)

## Proposed change

Standard combobox hygiene: commit the pending text on blur/Save (or block Save with an inline "press Add model" hint while uncommitted text is present). Silently dropping typed input is the one wrong option.

## Acceptance criteria

- [ ] Typed-but-uncommitted model id is either auto-committed on Save/blur or Save is blocked with an inline message
- [ ] Guide F19's literal steps (type → Save) persist the model after reload

Related: #2427 (closed — BYO model resolution).

### LINKED ISSUE #2680: Bug: locked discussion composer shows chat archived copy instead of the locked notice
## Problem

Locking a project discussion disables the composer with the **chat feature's archived copy** — "This chat is archived. Unarchive it to continue the conversation." — directly under a status pill that says **Locked** and a button that says **Unlock**. The discussion-specific locked notice (`discussions.reply.lockedPlaceholder`, "This discussion is locked") exists and is set, but is never visible.

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e): lock any discussion with a reply → composer overlay shows the archived-chat text. Reproduced on two discussions.

### Code pointers

- `services/platform/app/features/discussions/components/discussion-thread-view.tsx:335` — passes `disabledReason={isLocked ? 'archived' : undefined}` to the shared ChatInput (line 328 sets the correct `lockedPlaceholder`, which the overlay then covers)
- `services/platform/app/features/chat/components/chat-input.tsx:1336` — the overlay renders `tChat('archivedDisabled')` for `disabledReason === 'archived'`

## Proposed change

Add a `'locked'` variant to `disabledReason` (or a caller-supplied disabled message) and have the discussion view use it with the discussions copy, all locales included.

## Acceptance criteria

- [ ] Locked discussion composer shows the discussions locked notice, not chat's archived copy, in en/de/fr
- [ ] Chat's archived state unchanged
- [ ] Component test asserting the locked copy

Evidence: `services/platform/tests/screenshots/2026-07-12_r1/discussions/b2-locked-wrong-placeholder.png` (local QA run).

### LINKED ISSUE #2682: Bug: agent-type radio group has no accessible name
## Problem

The **Agent type** radio group on the agent editor's General tab (Agent / Coding agent / Image generation) exposes no accessible name: both `aria-label` and `aria-labelledby` are null on the `role=radiogroup` element. A screen-reader user tabbing into the group hears only the bare option labels with no grouping context (WCAG 1.3.1 / 4.1.2). Keyboard operation itself is fine (roving focus, Space/Enter both work).

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e) by DOM inspection of the radiogroup on `/dashboard/{org}/agents/{slug}`.

### Code pointers

- Agent editor General tab, agent-type `RadioGroup` (under `services/platform/app/features/agents/`) — the visible "Agent type" heading is not linked via `aria-labelledby`

## Proposed change

Point `aria-labelledby` at the existing visible heading (or add an `aria-label`).

## Acceptance criteria

- [ ] `role=radiogroup` exposes an accessible name matching the visible label
- [ ] vitest-axe (or the component test) asserts it

### LINKED ISSUE #2685: Bug: docs admin/members-and-roles.md describes an email-invite flow the app does not have
## Problem

`docs/en/platform/admin/members-and-roles.md` ("A worked invite") documents an **email-invite flow that does not exist**: "The new member receives an email link valid for 24 hours … change the role on the form before sending." The shipped Add-member dialog has no invite concept — the admin must set Name / Email / Role / **Password** directly, and the success screen prints the credentials once with "Save these credentials now. They won't be shown again."

## Current behaviour

Verified live (2026-07-12, main @ 5605bf56e): Settings → Organization → Add member with a new email and no password → inline "Password is required to create a new user"; with a password → shown-once credentials screen. No email is sent, no link, no 24h token. In-app copy ("Users who forget their password should contact an admin") and the auth manual guide (F14) both match the app, not the docs.

### Code pointers

- `docs/en/platform/admin/members-and-roles.md` (+ locales)
- Add-member dialog under `services/platform/app/features/settings/organization/`

## Proposed change

Rewrite the section to the real flow (admin-set password, shown-once credential handoff, out-of-band relay), or — if an invite flow is planned — mark it as such and file it as a feature. Sweep locales.

## Acceptance criteria

- [ ] members-and-roles.md describes the shipped Add-member flow in en/de/fr; no email-invite claims remain

### LINKED ISSUE #2686: Bug: manual guides drift — workspace.md describes a removed tablist and deleted test; agents.md describes removed List/Catalog/Metrics and Performance tabs
## Problem

Two manual test guides have drifted from the shipped UI badly enough to send testers chasing controls that no longer exist:

- `tests/manual/workspace.md` — the Automated-coverage table cites `workspace-file-tabs.test.tsx`, which was deleted (superseded by `canvas-file-tree.test.tsx`), and F3/A2 describe a horizontal **tablist** with ArrowLeft/Right + `aria-selected`, while the shipped control is a vertical `role="tree"` (aria-label "Canvas") grouped by source folder with ArrowDown-style movement (`canvas-file-tree.tsx:115`).
- `tests/manual/agents.md` — the "Scope & routes" table describes the removed List/Catalog/Metrics top tabs (`/agents` → `/agents/all`), a per-agent **Performance** tab (F3/F12), and a workforce metrics dashboard at `/agents/metrics` (F15); today `/agents` renders the roster directly, `/agents/all|catalog` resolve as pseudo-slugs to "Agent not found", `/agents/metrics` statically redirects to `/settings/metrics/usage` (#2382 removal), and `agent-navigation.tsx` has no Performance item. F17's claim that Coding agents lose the Skills tab is also wrong — Skills intentionally stays for external agents (code comment: sandbox skill staging), only Knowledge is chat-only.

## Current behaviour

Verified (2026-07-12, main @ 5605bf56e) by route probes and code reads during a QA round; the drift cost real testing time on nonexistent controls.

## Proposed change

Update both guides to the shipped structure (tree semantics + real test file; current agents routes/tabs and the corrected F17 expectation).

## Acceptance criteria

- [ ] workspace.md cites the real component/test and describes tree semantics for F3/A2
- [ ] agents.md routes/tabs match the app; F1/F3/F12/F15/F17 are executable as written

Related: #2382 (closed — workforce metrics removal), #2654 (closed — same guide-drift class for theme/language controls).

### LINKED ISSUE #2687: Improvement: Add-member dialog should disable submit while Password is empty for a new user (validation parity)

### LINKED ISSUE #2688: Improvement: mock gateway cannot synthesize workspace-panel tool events — Canvas/Files/Live-browser/Plan panes are untestable offline
## Problem

The workspace side panel (Canvas viewers, Workspace files, Live browser + takeover, Research plan) is the platform's only ⛔ manual-only area **and** is untestable offline: the mock gateway's four scripted scenarios (`e2e:reasoning`, `e2e:nextsteps`, `e2e:humaninput`, `e2e:error`) never emit `file_write` / `run_code` / `update_todos` / browser-stream events, so 30+ of the guide's 35 cases need mode B + a live sandbox. A QA round in mode A can exercise exactly F1 (panel absence) and the upload-driven fraction of the Canvas pane.

## Current behaviour

Verified (2026-07-12, main @ 5605bf56e): `services/platform/lib/mocks/overrides/canned.ts` contains only the four triggers above; `tests/manual/workspace.md`'s own coverage table marks the area manual-only, and its prerequisites require mode B/B+sandbox for F2–F20, B1–B8, A1–A5, P1–P4.

## Proposed change

Add deterministic mock scenarios that synthesize workspace tool events (e.g. `e2e:filewrite` emitting a small `file_write` for a .md/.mmd/.html, `e2e:runcode` with an output file, `e2e:plan` driving `update_todos`) so the Canvas/Files/Plan panes get offline manual coverage and become e2e-automatable, the same way the existing four scenarios cover reasoning/next-steps/human-input/error.

## Acceptance criteria

- [ ] Mock scenarios exist for at least file_write, run_code output, and update_todos
- [ ] workspace.md maps the newly coverable rows to mode A; at least one e2e spec exercises a pane via the new scenarios

yannickmonney:
## Summary

One sweep landing the 2026-07-13 QA round end to end — all fourteen issues #2675–#2688 close on merge — plus the settings and metrics UI unification built alongside them:

- **Issue fixes #2675–#2688** — per-issue table below.
- **Settings UI unification** — one settings design system across the org/account/agent/automation/workflow surfaces: env editors move to the header save cluster (`EnvVarListEditor` external save + `useEnvEditorController`), a shared `AgentTabContent` tab shell (8 hand-copied wrappers folded), hand-rolled boxes → `Card`, notices → `Alert`, unconfirmed destructive flows → `ConfirmDialog`/`DeleteDialog` (SSO disable/remove, branding reset, memory delete), ~17 raw palette sites → tokens, aria-label sweep, `SettingsToggleRow` adoption, `LocaleTabs` dedupe, `json-input` i18n + themed viewer.
- **Metrics UI unification** — restores the `automations.metrics.*` catalog the #2414 Apps→Automations rework dropped (Settings → Metrics → Automations rendered raw i18n keys in every locale), one `MetricsPeriodSelect` (was 4 hand-rolled copies with duplicate keys), one `MetricsSection` header (was 11 copies), `StatCardGrid cols=3`, locale-aware number/duration formatting (`useFormatNumber`, `lib/utils/format/{number,duration}`), feedback emerald/rose → chart tokens, and an e2e raw-key leak guard (`tests/e2e/specs/metrics.spec.ts`).
- **New i18n reverse guard** — a `usage-missing` check (keys referenced in source but defined in no locale) in the `packages/ui` i18n test framework, rolled out in report mode with a flip-to-enforce note; the 13 real dangling keys it found are fixed in en/de/fr.

Integration hardening found while landing (the workstreams were built in parallel sessions and never gated as one tree until now):

- The rebase over #2704 left a **duplicate `automations.metrics` block** in every locale — upstream restored the same catalog independently; `JSON.parse` keeps the last block, so every file-level gate was blind to it. Kept this branch's block (matches the code: shared `metrics.period.*`, no `link`), deleted upstream's; verified zero references to the dropped keys.
- The six new MCP form option keys were defined under `settings.mcpServers.*` while `mcp-server-form.tsx` reads the top-level `mcpServers` namespace — raw keys at runtime and six orphans in the gate. Moved next to their parent fields (`transportType`/`authType`/`oauth2`) in all three locales.

## Per-issue status

Every row closes on merge.

| Issue | Fix                                                                                                                                                                                                                                               |
| ----- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| #2675 | container boot skips the checkout-bound covering-gate half of registry completeness (`checkCoveringGates: false`); the vitest/build-time gates keep enforcing it                                                                                  |
| #2676 | provisioning probe counts only seedable entries (mirrors `copyTree`'s filters — `.gitkeep`/dotfiles/secrets never count); `retryProvisioning` re-probes after repair so `ok` can't toast success while the banner persists; tests for both layers |
| #2677 | uploads stamp `renderHint` only for images — everything else stays unhinted so viewer inference decides (`resolve-render-kind` extracted + tested); agent-written and uploaded twins of the same file now render alike                            |
| #2678 | agent Display name/Description no longer snap back on clear or concatenate retyped text (`general-tab-fields.test.tsx` regression tests)                                                                                                          |
| #2679 | BYO "Provider model id" typed but not added is committed on save instead of silently discarded (`byo-model-editor.test.tsx`)                                                                                                                      |
| #2680 | locked discussion composer shows its own locked notice via `ChatInput disabledReason='locked'` with caller-owned copy (`discussion-thread-view.test.tsx`)                                                                                         |
| #2681 | ~~logout confirmation returns focus to its trigger~~ — **already fixed on `main` by #2700**; my divergent copy dropped on rebase (upstream's shared `useRestoreFocus`/`restoreFocusRef` approach kept)                                            |
| #2682 | agent-type radiogroup gets an accessible name from the same i18n key as its visible section title                                                                                                                                                 |
| #2683 | ~~docs slug worked example~~ — **already fixed on `main` by #2698**; my divergent copy dropped on rebase                                                                                                                                          |
| #2684 | ~~webhook-triggers publish gate~~ — **already fixed on `main` by #2698**; my divergent copy dropped on rebase                                                                                                                                     |
| #2685 | members-and-roles docs describe the real add-member flow — no phantom email invite (en/de/fr)                                                                                                                                                     |
| #2686 | manual guides re-aligned: `workspace.md` (removed tablist/deleted test) and `agents.md` (removed List/Catalog/Metrics + Performance tabs)                                                                                                         |
| #2687 | add-member submit disabled while a new user's password is empty (validation parity; `member-add-dialog.test.tsx`)                                                                                                                                 |
| #2688 | mock gateway synthesizes workspace-panel tool events — canned Canvas/Files/live-browser/Plan panes are testable offline (`chat-scenarios.spec.ts` coverage)                                                                                       |

Closes #2675, closes #2676, closes #2677, closes #2678, closes #2679, closes #2680, closes #2682, closes #2685, closes #2686, closes #2687, closes #2688. (#2681/#2683/#2684 were already closed by #2700/#2698 landing on `main`; this branch was rebased onto them and my now-redundant copies dropped.)

Known deferrals stay open on purpose: workflow sidepanel save stub (#2693), provider drawer restructure, members-settings toolbar, `Badge` dark-mode colored variants, TrendIndicator "vs previous period" label, workflow display-name resolution in top-workflows, and flipping `usage-missing` to enforce once the remaining repo-wide dangling refs are swept.

## Verification

- `bun run check` — green across the monorepo: format clean, lint + typecheck 47/47 tasks, platform tests 76 283 passing. Two caveats, both machine-local: `tale-vision-read-hook > PDF when pdftotext missing` fails on this laptop because Homebrew's `pdftotext` defeats the "missing" simulation (the branch touches zero files in that workspace), and 5 convex-test timeouts under full parallel load pass 34/34 when run in isolation.
- `bun run lint:sast` — 232 rules on 5 022 files, 0 findings.
- `bun run migrations:check` — all four gates OK; the retention-bounds validator widening is a returns-validator change classified data-safe (no stored shape, no migration).
- Platform i18n suite 24/24 (parity + orphans + report-mode usage-missing); UI i18n framework tests green; e2e `metrics.spec.ts` 4/4 in the authoring session.
- Honest gap: no browser visual pass over the settings surfaces in this landing session (local Convex auth belongs to the developer's live account); component + e2e tests and the authoring sessions' verification carry it.

## Done checklist

- [x] Green gate — format/lint/typecheck/tests locally (caveats above are machine-local and re-verified in isolation)
- [x] Security gate — SAST 0 findings
- [x] Tests — every issue fix carries a regression test; new units for format utils, render-kind, retention diff, org actions/scaffold; e2e guards for metrics + workspace panels
- [x] Migration — N/A: no schema/table change; `migrations:check` green confirms (validator widening data-safe)
- [x] Locales — every user-visible string shipped in en/de/fr in lockstep; parity + orphan + usage-missing gates green
- [x] Docs — issue-driven docs updates (#2683–#2686) + stale brand-color claims in `pptx` SKILL.md and `design/docs/branding.md`
- [x] Accessibility — net-positive: focus return (#2681), radiogroup name (#2682), aria-label sweep, RadioGroup for tokenize/mask/block; documented WCAG exceptions (DSAR badges) left as-is
- [x] Sweep — blast radius walked per workstream note; integration defects from the parallel-session union found and fixed (duplicate catalog block, MCP namespace)
- [x] Observed — behaviour verified in the authoring sessions (tests + e2e); this landing session re-verified gates on the unified tree; settings visual pass deferred (auth-blocked) and stated above
- [x] Commits — one sweep commit (house precedent #2674), conventional header, branched off main, no attribution trailers


yannickmonney:
## CI round 1 diagnosis (6 failing Playwright jobs)

Four distinct causes; the two real ones were this sweep's own UI decisions outrunning the e2e suite:

1. **agent-editor (shard 1) — spec drift, fixed.** The sweep removes the per-tab page titles in the agent editor; four spec anchors still asserted those `<h2>`s, three via now-deleted i18n keys (the strict `t()` helper throws, and the serial suite skipped the sibling tests after the first throw, hiding three more). Anchors now use the surviving description lead-ins. Audited every `t()` reference across the e2e suite — all resolve.
2. **chat-scenarios #2688 (shards 4+5) — fixture gap, fixed.** Reproduced locally; the failure screenshot says it outright: *"Model tried to call unavailable tool 'update_todos'"*. The e2e assistant fixture only carried `request_human_input`, so the mock's `file_write`/`update_todos` calls killed the turn server-side ("Stream is not streaming" in the shard logs). Fixture now carries both tools, `maxSteps` 1→4 for the tool→ack loop. Both specs green locally (file-write in the parallel run; plan solo-verified — its one local re-fail was machine-load keystroke loss in the shared `fillComposer` retry loop, a mode CI never exhibited).
3. **project-files-desk (shard 10) — spec drift, fixed.** The sweep removes the post-install redirect (`handleFinish` only closes the wizard); the spec's `installIntoProject` still waited on the old URL before its own explicit navigation. Dropped the wait (`walkInstallWizard` already asserts the close). No other spec expects the redirect.
4. **settings-depth branding (shard 13) + docs — infra flakes, no change.** Branding never ran: its worker's org bootstrap timed out ("Seeded agent never appeared") on a backend logging 1s function timeouts; the seeding path this polls is untouched here. Docs was a 240 s webServer boot timeout and passed on the identical previous run.

Local verification of the fix round: 11/12 in one loaded parallel run + plan spec 1/1 solo; fixture-validation suite 6/6; pre-push SAST clean.

yannickmonney:
Rebased onto latest `main` (through #2700 + #2698). Those two PRs independently fixed **#2681** (logout focus) and **#2683/#2684** (agent-create slug docs + webhook publish gate) — the exact issues three of my commits targeted, and all three are now CLOSED. Resolved the five resulting conflicts by taking upstream's versions and dropping my divergent copies (notably #2700's shared `useRestoreFocus`/`restoreFocusRef` approach over my bespoke effect in `user-button.tsx`). My commit is 185→180 files; `closes` list trimmed accordingly. Post-rebase gate: typecheck + lint (platform, ui) clean, knip clean, i18n 24/24.

PR #2739, opened 2026-07-15
test(platform): live multi-link YouTube transcript extraction

yannickmonney:
Adds a gated live test proving the shipped video-link ingest transcript path yields a real, de-duplicated transcript across multiple link **shapes** and content types — extending `ytdlp_live.test.ts` (one video) to the multi-link case the trial feedback asked about.

## What it proves (observed locally, 3/3)
- Full `watch?v=` link → real transcript (music/audio-only video via auto-captions)
- `youtu.be/` short link → same transcript (URL-shape parity)
- A separate spoken-word video ("Me at the zoo") → real transcript (elephant/trunk)

Runs the ACTUAL `yt-dlp` + the SHIPPED parser (`parseVtt` + `rollingWindowDedup`) — the same code the ingest pipeline runs. Gated `YOUTUBE_LIVE_TEST=1` (self-provisions the toolchain, like `ytdlp_live.test.ts`); skips cleanly in the ordinary suite.

## Deliberately NO LLM / API key in the test
Per direction, a stored test must not call a paid LLM API. Whether the AI then answers correctly *from* a transcript is the RAG pipeline's job, proved hermetically in `tests/external-db-retrieval`. (The full paste→transcript→AI-answer flow was additionally verified once by hand against the real model — 5/5, answers correct — but that manual check is not committed.)


yannickmonney:
Superseded by #fix/live-youtube-test-resilience — the multi-link cases are folded into ytdlp_live.test.ts with bot-wall-resilient skipping (a bare live test would keep flaking the required Unit job).

Issue #2767, opened 2026-07-17
bug(platform): pausing human-input turn with no content text triggers a spurious model-fallback banner


yannickmonney:
## Problem

When an assistant turn emits a `request_human_input` tool call **without any content text** (reasoning + tool call only — a shape real models produce), the run pauses correctly and the approval card renders, but the fallback layer judges the generation empty/failed: it records a provider failure and injects a **"<model> was unavailable — switched to <model>"** system line into the chat. The resume then runs on the switched model.

## Current behaviour (repro on the hermetic mock stack)

1. Boot the Mode-A docs stack (mock gateway :4141, app :3000).
2. In chat, send a prompt whose mocked turn streams `reasoning_content` deltas followed by a `request_human_input` tool call and `finish_reason: tool_calls`, with **no** `content` deltas (the e2e `e2e:humaninput` scenario has this exact shape).
3. Approve the card.

The thread shows the approval card **plus** an amber `MODEL_FALLBACK` line (e.g. "claude-sonnet-4.6 was unavailable — switched to claude-haiku-4.5"), and the circuit breaker records a failure for a model that answered fine. The gateway log shows the tool-call request served with HTTP 200 — no provider error occurred.

With the same tool call preceded by a single content sentence, no banner appears — that is the workaround the docs video mocks now use (`DocsReply.toolIntro`, `services/platform/lib/mocks/overrides/docs-replies.ts`).

## Proposed change

Treat a turn that ends in an executed pausing tool call (`request_human_input`) as a successful generation regardless of whether it produced content text — neither record a breaker failure nor emit the `MODEL_FALLBACK` system line (`convex/lib/agent_chat/internal_actions.ts`).

## Acceptance criteria

- A reasoning+tool-call-only `request_human_input` turn pauses without a fallback banner and without a recorded provider failure.
- The resume after approval keeps the originally routed model.
- Regression coverage for the no-content pausing turn.

Issue #2716, opened 2026-07-14
bug(platform): chat composer's agent and model pickers paint blank on a thread route


yannickmonney:
## Problem

On a chat **thread** route, the composer's agent picker and model picker render as blank grey pills — the leading icon and the trailing chevron are drawn, but the label area is empty. They should read the agent name (`Assistant`) and the model (`Auto`).

## Current behaviour

Reproduced against `/dashboard/<orgId>/chat/<threadId>`. The label text is **present in the DOM** the whole time — a Playwright locator gating on the button's text (`getByRole('button', { name: 'Select agent' }).filter({ hasText: 'Assistant' })`) resolves and reports visible — yet the painted frame still shows an empty pill. So this is not "the data has not arrived"; the text is there and simply is not painted at that moment (a hydration/paint race, or a label that fades in and is caught at opacity 0).

The same composer on the empty-chat route (`/chat`) paints both labels correctly, so it is specific to the thread route's first paint.

## Impact

Cosmetic but user-visible: for a beat, the composer looks like it has lost its agent and model. It also made the docs screenshot of a code reply capture two blank pills, which is why it was found.

## Proposed change

Find why the picker's label is unpainted on first paint of a thread route (suspect: a skeleton/placeholder that stays mounted, or a fade-in transition) and let it paint with the value that is already in the DOM.

## Acceptance criteria

- [ ] Opening a chat thread never shows an empty agent/model pill.
- [ ] A regression test (or a docs-capture shot) covers the thread route's composer.

## Pointers

- Composer pickers: `chat.agentSelector.label` / `chat.modelSelector.label`
- Workaround in the docs pipeline: `services/platform/tests/docs-screenshots/manifest.ts` (`chat-code-reply` now crops to the message log to avoid the flaky element).

Issue #2661, opened 2026-07-11
Bug: chat attachment caps (count/size/type) are client-side only — chatWithAgentTurn re-enforces none of them


yannickmonney:
## Problem

The chat composer enforces four attachment limits — max 10 files, 200 MB total, 100 MB per file, and an allowed-MIME allowlist — entirely on the client. The server mutation that accepts a turn (`chatWithAgentTurn`) takes an unbounded `attachments[]` array with client-supplied `fileType`/`fileSize` and never re-validates any of it. The task-attachment path *does* re-check server-side (`convex/tasks/attachments.ts`), so this is an inconsistency as well as an abuse vector: a scripted client (or DevTools) can attach arbitrarily many/large/disallowed files to a chat turn.

## Current behaviour

Code-grounded (2026-07-11, main @ 94ef84e6f):

- `convex/agents/chat_turn.ts:55-64,153,367` — `chatWithAgentTurn` accepts `attachments` with no count/size/type ceiling; `checkUploadPolicy` returns "allowed" when the org has no upload policy configured (the default), so nothing bounds it.
- Contrast `convex/tasks/attachments.ts:28,36` — the task path re-enforces size/type server-side.

Repro (authenticated): in DevTools, call the Convex mutation for a chat turn with `attachments` of 30 entries and a `fileSize` of `5e8` — accepted, unlike the UI's 10-file / 100 MB guard.

### Code pointers

- `convex/agents/chat_turn.ts:55-64,153,367`
- client caps: chat composer upload gate (`app/features/chat/...` upload hook) — the only place the limits live today
- `convex/tasks/attachments.ts:28,36` — the correct server-side pattern to mirror

## Proposed change

Re-enforce the four caps server-side in `chatWithAgentTurn` (count, total size, per-file size, MIME allowlist), mirroring the task path, independent of whether an org upload policy exists.

## Acceptance criteria

- [ ] `chatWithAgentTurn` rejects over-count / over-size / disallowed-type attachments server-side with a ConvexError
- [ ] Limits match the client caps and the task-path enforcement
- [ ] A test drives the bypass (oversized/over-count) and asserts rejection

Distinct from #1620/#2026/#2048/#2078 (closed — those were client-path bypasses; this is the missing server mirror).

PR #2797, opened 2026-07-18
docs(platform): add chat seam map and PR seam-checklist

Israeltheminer:
## What

Adds `services/platform/convex/agents/CHAT_SEAM_MAP.md` — a reference doc that maps the **seams** of the chat feature: the boundaries where state crosses from one subsystem to another. It fixes no bug and changes no behaviour; it captures where the bugs have historically lived so the next chat change can be reasoned about before it merges.

Two parts:

- **Chat Seam Map** — 11 seams, each with _what state crosses_, the enforcing `path:symbol`, the invariant, the historical failure (PR/issue, open/closed), and the _hidden condition_ that keeps it from reproducing locally.
- **PR seam-checklist** — one yes/no question per seam + how to exercise it under its hidden condition.

The 11 seams: active-org context · external-agent lock · branch file listing · optimistic-bubble swap · provider turn-shape/retry · @-mention re-resolution · auto-route → UI signal · projectId binding · model governance RBAC · in-flight supersede · deploy-drain.

## Why

Across recent chat PRs/issues the bugs almost never live _inside_ a subsystem (streaming, routing, branching, RAG, org-scoping — each well-tested); they live on the seams, under a condition the author's machine never reproduces. Five conditions recur, so the same class of seam bug keeps shipping:

1. **Zero-latency localhost** — flicker/paint/optimistic-swap
2. **The single-org session** — cross-org coherence needs a user in 2+ orgs who switches org mid-session
3. **Provider output variance** — missing/unlabelled `finish_reason`, text-less reasoning/tool-only turns
4. **The shipped happy path** — client-only caps, uniformly-bundled tools, bundled backend hide the general bug
5. **Adversarial code-reading, not usage, finds them**

The checklist turns those into a repeatable pre-merge pass (a two-org session · a deployed/latency instance · a real provider · a custom/BYO agent · an adversarial sibling read).

## Where it lives / why

Co-located with the chat backend, matching the repo's existing co-located architecture-doc precedent (`convex/lib/rls/MEMBERSHIP_MIRROR_DESIGN.md`, `convex/conversations/README.md`, `lib/agent-adapters/README.md`). Named `CHAT_SEAM_MAP.md` per the `*_DESIGN.md`/README convention — in this repo `AGENTS.md` is reserved for authoring contracts. Internal English reference: no `nav.json` entry, no locale sync, no `skills:check` obligation.

## Notes for reviewers / follow-ups

- **Anchors verified** against the base branch: all 36 referenced file paths resolve and every symbol was confirmed. Anchored on `path:symbol` (not line numbers) so they survive drift.
- **Stale comment flagged — not fixed here (separate change):** `convex/threads/branch_schema.ts` says the branch file cut is on `threadFiles.createdAt`, but `convex/thread_files/queries.ts:listThreadFilesForUser` deliberately cuts on `updatedAt`. The prose is out of date; the code is correct.
- **Open issues referenced in the doc:** #2716 (composer pickers paint blank on a thread route) and #2767 (spurious model-fallback banner on a text-less human-input pause).
- **Candidates for later mechanical enforcement (noted, not built here):** a SAST rule for by-id thread reads that bypass the active-org guard; a test that `validateChatAttachmentCaps` rejects over-cap server-side (#2661); a provider-variance matrix over `shouldRetryGeneration`; a `BACKEND_DRAINING` new-turn gate test.

### DIFF
--- a/services/platform/convex/agents/CHAT_SEAM_MAP.md
+++ b/services/platform/convex/agents/CHAT_SEAM_MAP.md
@@ -1,0 +1,345 @@
+# Chat Seam Map & PR seam-checklist
+
+> Status: REFERENCE. This maps the **seams** of the chat feature — the boundaries
+> where state crosses from one subsystem to another — plus a **PR seam-checklist**
+> a reviewer runs against any chat change. It fixes no bug and changes no
+> behaviour; it captures where the bugs have historically lived so the next change
+> can be reasoned about before it merges. **Keep it truthful:** a change that moves
+> a chat seam updates the matching entry here in the same PR (root `AGENTS.md`:
+> "Instructions are docs too").
+>
+> Anchors are `path:symbol` relative to `services/platform/` (the common root of
+> the backend `convex/` and the SPA `app/`). Symbols are stable; line numbers are
+> not, so none are cited — grep the symbol.
+
+## Why this doc exists
+
+The chat subsystems — streaming, routing, branching, RAG, org-scoping — are each
+individually well-tested. The bugs almost never live _inside_ one of them. They
+live on the **seams**: a value carried across a boundary under a condition the
+author's machine never reproduces. Those conditions recur, so the same _class_ of
+seam bug keeps shipping. This map names the seams; the checklist forces the
+conditions.
+
+### The turn spine (the path every anchor hangs off)
+
+A chat turn flows: composer `app/features/chat/components/chat-input.tsx:ChatInput`
+→ send hook `app/features/chat/hooks/use-send-message.ts:useSendMessage` (paints
+the optimistic bubble via `pendingMessage`) → mutation wrapper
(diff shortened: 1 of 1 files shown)


Israeltheminer:
Closing without merging.

This doc is a set of precise, falsifiable claims about code (path:symbol anchors, invariants, specific enforcement sites) with nothing coupling those claims to the code they describe. A test fails when it drifts from reality; prose doesn't — so it can assert something false today or silently rot tomorrow with no signal, and it already shipped with one stale claim (branch_schema.ts createdAt vs updatedAt). Wiring an unverifiable reference into the workflow would propagate possibly-wrong statements to everyone who consults it — worse than not having it.

The value here was the one-time analysis, which is captured. Ongoing effort goes to measured observability (#2798), where the statements are computed from live data rather than asserted. Worth revisiting only if the seams get backed by executable checks that fail on drift — at which point those checks, not the prose, are the source of truth.

PR #2821, opened 2026-07-20
feat(platform): onedrive sync engine installs on first sync import

larryro:
## Summary

`onedrive/sync-files` — the hidden org-wide engine cron that drives OneDrive knowledge sync — was the only integration sync automation with `autoInstall: true`. Every org (including ones that never touch OneDrive) got it preinstalled at creation, ran its no-op `*/15` cron forever, and saw an unexplained **Onedrive** section on the Automations → Installed tab of fresh orgs.

The engine now installs **when a member actually sets up a Sync import** — the org's real opt-in moment — and never before:

- `builtin-configs/automations/onedrive/sync-files/automation.json` drops `autoInstall` (stays `hidden`).
- `importFiles`' sync branch schedules a targeted `installAutomationInternal` instead of re-running the whole defaults sweep. Idempotent end-to-end: `ensureOrgResources` upserts and `reconcileAutomationSchedules` collapses the schedule set, so repeat or concurrent imports can never duplicate the cron. A new sync import deliberately reinstalls after an org-level uninstall — creating a sync IS the org asking for the engine back.
- Manifest contract locked by tests: `hidden: true`, no `autoInstall`, **exactly one** schedule trigger (guards the duplicate-schedule regression at the manifest level).
- Docs (en/de/fr `platform/automations/builtin.md`): OneDrive moves out of "installed automatically at creation" to install-on-first-sync.

Second commit closes a pre-existing cross-tenant gap the change made more visible: `importFiles` trusted the caller-supplied `organizationId` with authentication only. It now runs the canonical `requireOrgMembershipById` gate before any work (same pattern as the rest of the unified public action surface).

Existing orgs keep their already-installed engine — no uninstall sweep, no schema change, no migration.

## Verification

- `configs:validate` (89 files), automations + onedrive server suites (287 tests), docs suite incl. i18n checks, `oxlint --type-aware`, `lint:sast` (467 rules, 0 findings) — all green.
- Live on the dev backend: ran the targeted install 3× against an org — `ok: true` every time, `wfSchedules` stayed at exactly one row; unauthenticated `importFiles` now rejects with `UNAUTHENTICATED` before any write.
- Dev DB survey that motivated this: 9/9 orgs carried a firing `onedrive/sync-files` cron; none had ever synced OneDrive.

## Definition of done

- [x] Green gate — format/lint/typecheck/tests pass for the touched surfaces (known pre-existing failure: `packages/ui/src/components/data-display/chart-card.tsx:112` type error from #2820, unrelated to this change)
- [x] Security gate — `lint:sast` clean; membership gate added at the touched boundary
- [x] Tests — happy path (sync installs), edge (one-time doesn't), errors (non-member rejected, failed token bails); manifest contract suite
- [x] Migration — N/A: no data-model or config-schema change; existing installs persist by design
- [x] Locales — docs updated natively in en/de/fr; no UI strings changed
- [x] Docs — `platform/automations/builtin.md` ×3 locales
- [x] Accessibility — N/A: no UI change
- [x] Sweep — builtin manifest + fixture copies (gitignored, synced on disk), CLI embedded files regenerated (gitignored), stale comments updated, frozen v0_3_4 migrations ruled out, e2e specs checked (no OneDrive dependency)
- [x] Observed — idempotency and the gate exercised live against the dev backend

## Notes for reviewers

- Riskiest spot: the semantics change that a **new** sync import reinstalls after an org uninstall (the defaults provisioner used to respect uninstall forever). Rationale in the `importFiles` comment.
- Reactivating an existing sync config does *not* install the engine — only a new Sync import does. Pre-existing footgun (engine uninstalled while configs stay active → silent no-sync) is unchanged and out of scope.
- Long-term follow-up (separate issue if wanted): converge OneDrive onto the sibling model (Google Drive / Gmail style — visible catalog automation, explicit install) and retire the parallel knowledge-picker architecture.


larryro:
Pushed two additional commits fixing the pre-existing main breakage from #2820 that was failing this PR's CI:

- **Type check**: `chart-card.tsx:112` fed `string | undefined` into `EmptyState`'s required heading after `title` became optional. The empty-state now renders only when a heading (emptyTitle ?? title) exists — contract documented on the prop, locked by a test.
- **Unit**: two orphan i18n keys (`settings.integrations.panel.{addIntegration,integrationDetails}`) left behind when the sheet title switched to the integration's identity — removed from en/de/fr.

PR #2845, opened 2026-07-22
feat(platform): multi-account foundation for IMAP/SMTP mailboxes

Israeltheminer:
## What & why

First slice toward reworking `imap_smtp` from a single catch-all mailbox into **several real mailboxes under one Email integration**, with provider-agnostic outbound. **Foundation only** — purely additive and behavior-preserving; the sync, per-account routing, and settings/compose UI that consume it land in follow-up PRs.

## Changes
- **Account shape** — `imapSmtpAccountValidator` / `imapSmtpAccountEncryptedValidator` (+ inferred types). Each account carries its own From address, IMAP + SMTP/relay host/port/TLS (typed — not the old string-parsed booleans), `sentMailbox`, `isDefault`, per-account `routing`, and encrypted `imapAuth`/`smtpAuth`. SMTP is generic (the mailbox's own SMTP, any relay, or a self-hosted submission host) — no relay baked in.
- **Storage** — optional `imapSmtpAccounts` array on `integrationCredentials` (data-safe growth; the schema snapshot classifies it safe, so no migration), surfaced on `LoadedIntegration`.
- **Resolver** — `resolveImapSmtpConnection` accepts an optional `accountId`: selects by id → the `isDefault` account → the first; SMTP reuses the account's IMAP login unless it carries a distinct `smtpAuth`. A pre-save dry-run keeps using the legacy inline path.

## Behavior
No user-visible change. With no `imapSmtpAccounts` stored, resolution is byte-for-byte today's single-mailbox path (proven by the retained legacy tests).

## Tests / gates
`imap_smtp_config.test.ts`: 5 retained legacy + 5 new multi-account cases (account selection, default/first fallback, distinct SMTP relay, unknown-id error, legacy dry-run precedence). Green: `tsc --noEmit`, oxlint, oxfmt, `migrations:check`.

### DIFF
--- a/services/platform/convex/integrations/credentials_schema.ts
+++ b/services/platform/convex/integrations/credentials_schema.ts
@@ -2,6 +2,7 @@ import { defineTable } from 'convex/server';
 import { v } from 'convex/values';
 
 import { jsonRecordValidator } from '../lib/validators/json';
+import { imapSmtpAccountEncryptedValidator } from './validators';
 
 /**
  * Slim credentials table for installed integrations.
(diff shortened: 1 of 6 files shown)


Israeltheminer:
Superseded: replacing the imap_smtp-specific multi-account model with a general "duplicate integration as a new instance" capability — each mailbox becomes a normal single-mailbox install under its own slug, so multi-instance works for any integration without a bespoke sub-model.

Issue #2842, opened 2026-07-22
perf(platform): first-load static assets bottlenecked by single origin (~3.6 KB/s, up to 20s TTFB)


Israeltheminer:
## Problem

Dashboard first paint is dominated by **static-asset delivery, not payload size**. A HAR of `/dashboard/<org>/conversations/open` (148 requests) shows the ~870 KB JS/CSS bundle taking ~4 minutes to arrive: the router preloads ~147 route chunks in one burst over a **single HTTP/2 connection to a single origin**, which queues them (TTFB climbs with queue position, up to **20.7 s**, median ~16 s) and delivers at an aggregate **~3.6 KB/s**. Render-critical chunks (`markdown-renderer`, `chat-input`) finish last (~3.9 min).

This is downstream of the HTTP-caching fix (immutable headers on hashed assets): caching removes the *repeat*-visit cost, but the *first* uncached load is still gated by origin throughput.

## Current behaviour

- All static assets are proxied to the platform app process: `services/proxy/Caddyfile` default `reverse_proxy platform:3000` ("static files + HTML env injection only"). No `file_server` for `/assets/*`, no CDN.
- The proxy container has **no access to `dist/`** (proxy Dockerfile copies only the Caddyfile + `maintenance.html`; no shared `dist` volume), so Caddy cannot serve the assets directly today.
- Single origin (self-hosted, single box), one HTTP/2 connection for all 147 requests; TTFB 0.2 s → 20.7 s with queue depth; aggregate ~3.6 KB/s.

## Proposed change (options for ops/infra)

- **(a)** Serve `/assets/*` from Caddy `file_server` (or a proxy cache) instead of the Node process. **Prerequisite:** export the built `dist/` from the platform image to a volume the proxy can read.
- **(b)** Investigate origin saturation on the box (CPU/IO/bandwidth) — a static server should have sub-100 ms TTFB, not 12–20 s.
- **(c)** Optionally shrink the first-load burst: coarsen `manualChunks` and defer katex/shiki until a message needs them (fewer concurrent requests on the shared connection).
- A third-party CDN is intentionally **out of scope**: assets are served same-origin for GDPR/air-gap (CSP policy in `services/platform/server.ts`).

## Acceptance criteria

- First uncached dashboard load completes in seconds, not minutes, on a normal connection.
- Static assets have sub-second TTFB.
- Same-origin / air-gap constraint preserved (no third-party CDN).

## Evidence

[...]

PR #2847, opened 2026-07-22
perf(platform): skip CSP/nonce middleware for hashed assets

Israeltheminer:
## Problem

Follow-up to the caching fix (#2843): repeat visits are now fast, but the *first* (cold) visit is still slow. Part of that is self-inflicted — on a cold load the ~150 content-hashed JS/CSS chunks each pass through the global `secureHeaders` middleware, which builds a full CSP string and a **fresh crypto nonce per request**, on the single Bun event loop. A static script/style byte stream never uses that nonce'd CSP, so it's wasted work paid ~150× during the burst. (Confirmed: on `main`, a request to `/assets/vendor-katex-….js` comes back with `content-security-policy: default-src 'self'; script-src 'nonce-…'`.)

## Change

Bypass `secureHeaders` for content-hashed `/assets/*` in the `app.use('*')` guard — mirroring the existing `/canvas-preview` and `/dav/*` bypasses — and re-assert `X-Content-Type-Options: nosniff` on the static-file response (the one header that matters for a script/style stream; CSP/X-Frame/Referrer are document-level). Uses the same `IMMUTABLE_ASSET` pattern as the cache rule, so only hashed JS/CSS bypass; un-hashed images under `/assets/` still get the full middleware.

## Verification

- `bun run --filter @tale/platform test server.test.ts` → green (57), incl. a new test asserting a hashed-asset response carries **no** CSP while `/api/health` still does. Verified it **fails on `main`** (asset gets the full CSP) and passes here.
- `typecheck` clean; `oxfmt --check` + `oxlint` clean.

## Honest caveat — this is measure-after

I can't prove from the HAR alone that this middleware is *the* cold-load bottleneck (the box's raw bandwidth and Caddy's on-the-fly gzip/zstd are also candidates). It's a correct optimization regardless (don't build a per-request nonce for static bytes). Next step: deploy, capture a fresh **cold** HAR, and measure first-load TTFB/throughput. If it's not enough, the documented escalation is moving `/assets/*` to Caddy `file_server` (bigger infra change; see #2842).

### DIFF
--- a/services/platform/server.test.ts
+++ b/services/platform/server.test.ts
@@ -264,6 +264,20 @@ describe('security headers', () => {
       "connect-src 'self' https://acc.r2.cloudflarestorage.com",
     );
   });
+
+  test('content-hashed assets bypass the CSP/nonce middleware', async () => {
+    const app = createApp(baseEnv);
+    // A content-hashed asset path skips secureHeaders (no per-request CSP +
+    // crypto nonce), so its response carries no CSP — while a normal path still
+    // does (control). Independent of whether the file exists on disk.
+    const asset = await app.fetch(
+      new Request('http://localhost/assets/vendor-katex-DUoGyCxW.js'),
+    );
+    expect(asset.headers.get('content-security-policy')).toBeNull();
+
+    const normal = await app.fetch(new Request('http://localhost/api/health'));
+    expect(normal.headers.get('content-security-policy')).not.toBeNull();
+  });
 });
 
 describe('POST /canvas-preview', () => {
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Closing — net negative on review.

**Security:** this introduces a clickjacking hole. Skipping the security-headers middleware by URL *pattern* means a forged `/assets/<hash>.js` path (matches the pattern, no real file) falls through to `index.html`, served as `200 text/html` with **no CSP and no X-Frame-Options**. Reproduced against a simulated production `dist`.

**Benefit doesn't hold up:** the per-request CSP nonce it avoids costs microseconds — it can't explain the ~20s cold-load TTFB. That bottleneck is delivery/infra (#2842), not this middleware.

Keeping the security headers intact on every path. Cold-load will be pursued via the Caddy `file_server` offload (#2842) with real measurement. The merged wins (#2843 caching, #2846 hash-length) are unaffected.

Issue #2803, opened 2026-07-19
Bug: Toasts (including "update available") must render at the outer top-right


yannickmonney:
## Problem

Toasts such as the changelog **“update available”** prompt are not painted at the outer top-right of the viewport. They should all render in that corner, outermost (above other chrome), consistently.

## Current behaviour

- Global `Toaster` (`app/components/ui/feedback/toaster.tsx`) defaults to `top-right` and portals the viewport to `document.body` at `z-100`.
- The viewport intentionally adds large top padding — `pt-[calc(4rem+var(--safe-top))]` — to clear the app/sheet header (see comment citing #1986). That pushes toasts **down** from the outer top-right rather than flush to the corner.
- Changelog toast: `ChangelogToastTrigger` → `toast({…})` via `use-changelog-notification.ts` (no custom `position`; inherits default).
- Some auth flows still force `position: 'top-center'` (log-in / 2FA routes), so not every toast is top-right.

## Proposed change

- Place the toast stack at the **outer top-right** of the viewport (flush with safe-area insets), without the header-clearing offset that pulls them inward — or replace that offset with a solution that keeps outer placement while avoiding the #1986 overlap regression (e.g. higher z-index / avoid covering interactive header controls another way).
- Ensure changelog “update available” and ordinary success/error toasts share that same outer top-right placement.
- Decide whether auth `top-center` toasts should stay an exception; default product toasts must be top-right outer.

## Acceptance criteria

- [ ] “Update available” and standard toasts appear in the outer top-right (safe-area aware), not mid-header-band.
- [ ] Toasts remain clickable above sheets/dialogs (`z-index` / portal behaviour from the #1986 fix preserved or replaced with an equivalent).
- [ ] No regression where a toast permanently covers an untouchable header control without a dismiss path.
- [ ] Spot-check: dashboard with changelog toast; settings sheet open; ordinary save toast.

## Affected service

Platform (web app)

## Related

Closed: #1986 (toast under sheet title), #2552 (changelog toast re-fire), #2361 (toast duration).

Israeltheminer:
Fixed on `main` via #2857 (`2743ee48e`): product toasts use the outer top-right gutter `pt-[calc(0.75rem+var(--safe-top))]` (same as the side edges). Supersedes #2874.

PR #2874, opened 2026-07-28
fix(platform): place product toasts at the outer top-right

AdeolaAdekoya:
## Summary
- Product toasts (including changelog “update available”) now sit at the **outer top-right**, using `0.75rem` + safe-area padding instead of the old `4rem` header-clearing offset that pushed them into the mid-header band ([#2803](https://github.com/tale-project/tale/issues/2803)).
- Body portal + `z-100` unchanged so toasts still paint above sheets/dialogs.
- Auth `top-center` toasts left as an intentional exception.

## Test plan
- [x] `toaster.test.tsx` asserts `pt-[calc(0.75rem+var(--safe-top))]`
- [ ] Dashboard: trigger changelog “update available” — toast at outer top-right
- [ ] Open a settings sheet; fire a save toast — still clickable above chrome
- [ ] Spot-check auth login toast still top-center

## Definition of done
- [x] Green gate — focused toaster tests
- [x] Tests — regression for viewport padding class
- [x] Migration — N/A
- [x] Locales — N/A (no new strings)
- [x] Docs — N/A
- [x] Accessibility — N/A (placement only)
- [x] Sweep — toaster viewport only; auth top-center left alone
- [ ] Observed — visual spot-check pending in browser
- [x] Commits — one focused fix on a branch off `main`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/services/platform/app/components/ui/feedback/toaster.test.tsx
+++ b/services/platform/app/components/ui/feedback/toaster.test.tsx
@@ -37,15 +37,13 @@ describe('Toaster', () => {
     expect(viewport?.textContent).toContain('Update available');
   });
 
-  // Regression (#1986): the viewport must clear the top header band, not sit
-  // flush against the top edge. A right-anchored Sheet's header lands in the
-  // top-right corner — the same spot as the toast — so a flush toast covered
-  // the panel title. The `4rem` top padding drops the first toast below that
-  // header band.
-  it('offsets the viewport below the top header band', () => {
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Closing as superseded — the same toast inset change (`4rem` → `0.75rem` symmetric top-right gutter) already landed on `main` in #2857 (`2743ee48e`), so this branch now only conflicts on comments/test naming with no remaining code delta.

PR #2878, opened 2026-07-30
fix(platform): collect a connector's declared config fields

Israeltheminer:
Adding an IMAP / SMTP mailbox credential fails with

```
Connector "imap-smtp" needs "IMAP server"
```

— naming a field the form never offers. `createCredential` validates the submitted
config against the connector's declared `configFields` and refuses a missing
required one, but the add dialog installed `noExtras()`, so it renders none of
them. Any connector declaring a required config field is unreachable through the
UI; `imap-smtp` requires both `imapHost` and `smtpHost`.

The backend half already existed — `normalizeConfig` coerces types, honours enums,
applies defaults and rejects undeclared keys. This wires the form to it.

## Changes

- **`listConnectors` returns `configFields`.** The settings catalog dropped them,
  so the UI could not have known what to ask for.
- **A config extra module renders one control per declared field**, typed from the
  declaration: `boolean` → switch, `string` + `enum` → select over exactly those
  values, otherwise an input. Numbers stay strings in form state and are coerced
  by the server. A declared default renders as a **placeholder, not a prefilled value**.
- **Blank entries are dropped rather than sent as `''`.**

## Completeness and edit round-trip

`CredentialExtraModule` gains optional vendor-aware `isComplete`; both dialogs gate on it.
`listCredentials` returns `config` so edit cannot wipe hosts/ports.

## Follow-ups

- Split SMTP auth + live imapflow/nodemailer + credentials table polish → #2892 (stacked on this PR).
- Pack model-id dialect matching + Run-live dialog close → #2891 (independent).

## Note

Field labels come from the shipped connector YAML rather than the locale catalogs.


Israeltheminer:
Rebased onto current `main` after the credential catalog became a table (`credential-add-dialog` instead of `credential-create-dialog`). Scope unchanged: configFields collection only — split SMTP auth and credentials UI polish land in a follow-up stacked on this PR.

PR #2894, opened 2026-08-03
feat(platform): task agent run details, live transcript, and output attachments

larryro:
## What

Four related pieces around the task-agent run lane, driven by watching a real run (the panda-deck task) end to end:

1. **Details entry on the task's agent run card.** The run card only said "Agent working…" with no way to see what the agent was doing, and nothing at all after it settled. Every run (live, settled, failed) now offers **Details** — a dialog with the run's sandbox transcript (agent prose + tool rows, unfoldable payloads). The transcript view is extracted from the automation lane's `AgentExecutionLog` so both lanes render through one component; the new `getTaskAgentRunSandboxOp` query is project-ACL-gated like the run card itself and keys to the run's own exec inside the agent's standing session.

2. **The task lane now streams live progress at all.** Root cause of the empty log: `tasks/agent_run_host.ts` ran `drainHarnessWindow` with **no progress sink**, so the op row never carried `progressText`/`liveTimeline` — live or settled. `liveProgressSink` moves out of the automation host (parameterized by op kind) and both task-agent drain sites flush exactly like workflow agent nodes.

3. **Agent deliverables attach to the task (Output zone, overwrite-by-name).** Harvested `/user/output` files previously stranded in blob storage with only their *names* in the report comment — and, unclaimed, the agent temp-retention GC would sweep the blobs. The settle now (a) claims each blob's metadata out of the GC lane, (b) merges the files into a new `outputs` field on the task — **same fileName replaces the entry and purges the superseded blob; new names append** — and (c) the task modal renders them as a read-only **Output** zone (en/de/fr). The standing session's `/user/output` is swept at turn start; without that, a previous task's leftovers re-harvested onto whatever task ran next (observed live: a prior deck landed on an unrelated task).

4. **Vision polyfill: never auto-select a `:free` model.** The cheapest-first pick chose `google/gemma-4-26b-a4b-it:free` (price 0) and **every vision call of the observed run 401'd at the provider** (35/35 in the gateway log) — free tiers sit behind per-account data-policy gates and hard rate caps. Auto-selection now skips `:free` variants; an explicit allowlist pin can still name one.

## Verification

- Unit: new convex-test suites for the op query (exec matching, sibling-run isolation, fail-closed access) and the outputs merge (append / replace+purge / identical-blob no-op / empty no-op); a `:free`-exclusion case for the vision resolver; UI tests for the Details dialog (transcript + empty state). Touched-area suites green: 619 server + 232 UI tests; `migrations:check` green (`outputs` is a data-safe additive field); locale parity suites green.
- Live, against the dev stack: started a fresh agent run — Details shows the transcript streaming while the run works and keeps it after settle; the Output zone appears with the produced file; a second run producing the same fileName **replaced** the entry (new blob id, still one row); the leftover-sweep kept the prior task's stale deck out of the second harvest.

## Notes

- The Output zone reuses the attachments renderer under its own label — no new file-display code path.
- `sessionOpTimelinePartValidator` replaces three inline copies of the timeline-part shape.
- The wrongly-attached deck on my local smoke task predates the sweep fix and stays as dev data; new runs can't reproduce it.

### DIFF
--- a/services/platform/app/features/automations/components/agent-execution-log.tsx
+++ b/services/platform/app/features/automations/components/agent-execution-log.tsx
@@ -208,38 +208,41 @@ export function AgentActivityLine({
   );
 }
 
+/** The op fields the transcript view reads — the common projection of the
+ * automation (`getAgentNodeSandboxOp`) and task (`getTaskAgentRunSandboxOp`)
+ * queries, so both lanes render through ONE view. */
+export interface AgentSandboxOpView {
+  execId: string;
+  status: string;
+  progressText?: string;
+  liveTimeline?: TimelinePart[];
+}
+
 /**
  * What the agent did INSIDE the sandbox, as a chronological transcript: its
  * own words as flowing prose, each tool call as one compact row that unfolds
  * into the full payload. Streamed onto the run's session op by the agent host
- * and read back here — the only window into a sandbox turn, which would
- * otherwise be an opaque "working…" spinner until it settles.
+ * and read back by the caller's op query — the only window into a sandbox
+ * turn, which would otherwise be an opaque "working…" spinner until it
+ * settles.
  *
  * The transcript reads DOWNWARD like the turn itself: the pane pins to its
  * bottom while the reader stays there (new entries just appear, chat-style)
  * and stops following the moment they scroll up, with a jump-back pill.
  * Entries accumulate client-side across the op's bounded flushes, so rows
- * never vanish mid-read. Renders nothing when the run never ran an agent
- * node.
+ * never vanish mid-read. Renders nothing while `op` is null.
  */
-export function AgentExecutionLog({
-  organizationId,
-  runId,
+export function ExecutionLogView({
+  op,
   className,
 }: {
-  organizationId: string;
-  runId: Id<'automationRuns'>;
+  op: AgentSandboxOpView | null;
   /** Sizes the scroll pane — the run dialog stretches it, the run page caps
    * it. The pane must have a bounded height for the pinning to mean anything. */
   className?: string;
 }) {
   const { t } = useT('automations');
   const { t: tChat } = useT('chat');
-  const opQuery = useConvexQuery(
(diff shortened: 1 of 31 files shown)


larryro:
Added on top: **@mention puts a project agent to work** (fc6eaf8).

Commenting `@alice 把第3页的图换成真实照片` on a task now (re)assigns it to the mentioned project-agent instance and kicks a run whose brief carries the comment as reviewer feedback — the review-loop verb the settled run card lacked. Instances resolve by display name (member handle convention, shadowing same-named legacy slugs); the trigger is write-gated, quiet on refusal, and never preempts a live engine. `assignTask`'s core and `startTaskAgentRun`'s admission moved into shared helpers so the comment path is the same code as the picker and the board verb; runs record `trigger` + `feedback`.

Verified live end-to-end: an `@alice` comment on a settled task kicked a rerun within a second, and the produced file's content matched the FEEDBACK (the task description still said v1 — the v2 content could only have come from the comment). Same-name overwrite replaced the Output entry again.

larryro:
Two more on top:

**Run card → property panel** (c692e26). The run card read as a second subject in the task body; the run is the ASSIGNEE's state. It's now a compact **Run** property under Assignee — status caption (harness · model on hover) + the same small verbs (Start / Cancel / Retry / Details) + the same transcript dialog.

**Per-task delivery box** (33365a2). The standing session serves every task of one agent while the live-run mutex is per TASK — so two runs could share one `/user/output` and cross-attach or sweep each other's deliverables. The box is now `/user/output/<taskId>/`: named in the turn's instructions AND in the user-prompt brief (observed live: a resumed conversation reused last turn's remembered root path when only the system addendum named the new one — the deliverable went uncollected), swept per task at start (plus legacy loose root files), harvested per task at settle.

Live-verified: the rerun's transcript wrote `/user/output/wh7ajp…/hello.txt`, the Output zone replaced the entry (new blob, same fileName), and the root-path run before the prompt fix correctly collected nothing rather than mis-attaching.

PR #2901, opened 2026-08-03
fix(platform): restore the log-out confirm copy and unbreak the gates

Israeltheminer:
The copy rework in #2897 left `main` red on two gates. Verified on a pristine worktree at `121a11b3d` with nothing else applied.

## `auth.userButton.logOutConfirm.description` — user-visible

The key was dropped from en, de, and fr, but `app/components/user-button.tsx:654` still renders it:

```tsx
description={t('userButton.logOutConfirm.description')}
```

So the log-out confirmation dialog currently shows the raw key path instead of a sentence. `lib/i18n/messages.test.ts` catches this as `[referenced-key-missing]` in enforce mode. Restored in all three catalogs with the wording they had before the rework.

## `products.create.description` — orphan

The create dialog became a wizard with per-step hints (`createWizard.basicsHint` / `pricingHint` / `reviewHint`) and stopped using the flat description. The key stayed behind in all three catalogs and trips the orphan check. Removed.

## Stale delete-confirm assertion

`products-crud.test.tsx` still expected `Are you sure you want to delete`, which #2897 reworded to `Delete "{name}"?`. The assertion now checks the current copy via `toHaveTextContent` rather than `getByText`: the confirmation interpolates the product name into its own element, so the sentence spans several nodes and an element-scoped matcher can't see it whole.

## Verification

On `main` at `121a11b3d`: `messages.test.ts` fails 2/48, `products-crud.test.tsx` fails 1/8. With this branch both suites pass — 48/48 and 20/20 including `user-button`.

### DIFF
--- a/services/platform/app/features/analytics/usage/usage-metrics-page.test.tsx
+++ b/services/platform/app/features/analytics/usage/usage-metrics-page.test.tsx
@@ -57,8 +57,8 @@ describe('UsageMetricsPage', () => {
     expect(screen.getByRole('button', { name: 'Filter' })).toBeInTheDocument();
 
     // Static summary-card labels asserted by the E2E.
-    expect(screen.getByText('Total Requests')).toBeInTheDocument();
-    expect(screen.getByText('Active Users')).toBeInTheDocument();
+    expect(screen.getByText('Total requests')).toBeInTheDocument();
+    expect(screen.getByText('Active users')).toBeInTheDocument();
   });
 
   it('passes axe audit in its loaded state', async () => {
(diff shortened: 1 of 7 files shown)


Israeltheminer:
Superseded by #2909, which landed the same repairs independently: `auth.userButton.logOutConfirm.description` is restored (with better wording), `products.create.description` is gone, and the stale assertions in `products-crud`, `usage-metrics-page`, `new-automation-dialog`, and `node-inspector` are updated.

Verified on `main` at 35c0548e8: full typecheck 14/14 tasks, i18n gate 48/48, and the previously red `chat-surface` / `composer` / `providers-settings` suites all pass. Nothing left here to land.

PR #2947, opened 2026-08-08
feat(platform): add compliance provenance layer

larryro:
## What

The compliance provenance layer for regulated (IEC 62304 / ISO 13485-style) agentic development processes — four platform capabilities, all strictly additive (safe growth: new tables, optional fields, policy-gated enforcement; default behavior unchanged except one deliberate bug-fix tightening, called out below).

### 1. Provenance ledger (`convex/audit_logs/agent_run_ledger.ts`)
One immutable audit-chain entry (`category: 'agent'`, action `agent.run_settled`) per settled agent run, written **inside the terminal mutation** — atomic with the settle, exactly-once by the settle election's own guards. Covers all 7 terminal doors (task settle/fail/cancel ×2 lanes, automation finish/cancel/one-piece). Each entry carries: model identity (requested / served / vision), the gateway virtual-key capability snapshot (model allowlist, budget, connector+tool grants), spend, produced outputs with `_storage` sha256, the knowledge read-set (exec-pinned; sibling-turn reads excluded — false provenance is worse than omission), approvals raised, and the named-reviewer linkage (review decisions join on `runId`). Rides the existing per-org hash chain, so integrity verification, export, and up-to-10-year retention already apply.

### 2. Knowledge retrieval ACL + read-set capture
Corpus documents now carry `team_ids[]` / `project_id` (DDL migrations 06+07 + a versioned backfill migration). Search, `rag_fetch`, and the chat inline-row fallback all enforce the caller's document visibility, derived server-side from the **same rules as listing** (`documents/access.ts`) — this closes the documented-but-unenforced scoping gap where team/project documents were retrievable org-wide. Multi-team documents match by array overlap. Denied fetch is indistinguishable from missing (no existence leak). Scoped queries bypass the semantic cache. `rag_search`/`rag_fetch` dispatches record their refs + exec onto `sandboxToolCalls` — the ledger's read-set source. Admin-keyed lanes (REST, MCP `get_knowledge`) stay org-wide by design, with comments; `get_knowledge` now returns durable refs.

⚠️ **Behavior change (deliberate, release-note worthy):** after the backfill runs, retrieval tightens to match documented document visibility. Users who relied on the org-wide leak will see fewer search results.

### 3. Autonomy tiers + competence records
`autonomyTier: a1|a2|a3` (optional) on project agents and automation nodes. **Unset = today's behavior bit-identical** (regression-tested). Tiers only tighten: `a3` = write effects refused outright at the approval gate (no approval row — nothing a human could grant); `a2` = every outbound write parks even where `approval_policy` says auto-approve; `a1` = declared posture, runtime-identical to unset. Tier is stamped into the ledger. `competenceRecords` (admin grant/revoke, never hard-deleted, audited) + a `review_policy` governance policy (`requireIndependentReviewer`, `requiredCompetences`) enforced in task review — the responder must differ from the run's driver and hold the required competences; check results are recorded into the approval metadata on the audit chain.

### 4. Controlled document records
Opt-in per document (`markControlled`, upload/agent-sourced only): draft → in_review (**frozen** — reviews evaluate a fixed artifact) → approved (**immutable**) → new revision (version+1). Approved snapshots are sha256-addressable via `record.approvedVersions` and retained in `historyFiles`. Freeze/trash guards wired into **every** content path: public/internal update, connector upsert, WebDAV PUT/DELETE/cascade, REST, trash, folder-delete cascade, knowledge-entry delete. The generic `updateApprovalStatus` mutation now refuses review-type approvals (`task_review`, `document_record_review`) — decisions must go through the dedicated respond mutations.

## How it was built & verified

Five phases implemented by parallel subagents with disjoint file ownership, then an independent adversarial review of the full diff. The review surfaced 2 guard bypasses (folder-delete cascade, knowledge-entry delete), the multi-team retrieval gap, and a ledger robustness issue (malformed output ref could wedge a settle) — all fixed with regression tests (the bypass tests were falsified against the disabled guard).

- Full `bun run check`: **44/44 tasks green** (format, lint, typecheck, all workspace tests)
- `migrations:check`: all four gates green (new schema fields are data-safe growth; corpus DDL + backfill ship as proper migrations)
- SAST (opengrep): 0 findings
- New/extended suites: ledger (13), knowledge scoping (~40 across corpus/fetch/access/bridge), tier lattice + competence + review policy, controlled-records state machine + freeze matrix (21+), bypass regressions
- i18n: en/de/fr parity script-verified for every new key (native translations per the write-translations doctrine)

## Known boundaries (deliberate, documented in code)

- Governance retention & GDPR erasure can still remove approved records (erasure is a legal right; retention is explicit org governance) — flag if the compliance posture should exempt approved records from retention.
- `review_policy` applies to task reviews in v1 (not document-record reviews); document reviews have no bell notification yet (badge + approval row carry visibility).
- `a2`/`a3` on automation **agent** nodes is declaration-only until the sandbox connector bridge gains a write lane (documented in `lib/engine/core/types.ts`); enforcement is live on connector nodes.
- UI verified by tests only — a manual pass over the tier selector and controlled-record flows on a dev stack is recommended before release.
- End-user docs pages for the four features are a follow-up (journey-first docs need driven UI + screenshots).


larryro:
## Live E2E verification (local dev stack) + one fix

Drove the four features in a real browser against a running dev stack. Flows A–D all pass; two things came out of it.

**Verified end to end**
- **Controlled records**: upload → mark controlled (`v1 · Draft`) → submit for review (`v1 · In review`, "the document freezes while it is in review") → delete replaced by a disabled *Protected controlled record* item → approve (`v1 · Approved`) → new revision (`v2 · Draft`). All four `document.record_*` actions land in the audit log with the right actor; the governance page reports **Chain integrity — Verified**.
- **Autonomy tier**: selector defaults to *Unset (default) — "No declared tier — the agent behaves exactly as before."*, A1/A2/A3 each with a one-line description; the choice persists across dialog reopen.
- **Provenance ledger**: a real run started (`Working…`) and was cancelled; the audit log gained an `agent.run_settled` entry whose metadata carries `finalStatus: cancelled`, `surface: task`, **`autonomyTier: "a2"`** (the tier set in the agent dialog flows through), harness, trigger, gateway `{allowedModels, budgetCents, keyId}`, model `{requested, servedRef, visionRef}`, run/task/exec ids, and duration.
- **Knowledge**: after indexing, chat retrieval found the seeded phrase and cited the source document; the RAG tool calls were themselves audited.

**Fixed (5144638e7): approved history could be deleted via a revision**

Trash protection keyed off the record's *current* state, so `New revision` (which returns the record to `draft`) re-enabled deletion — and deleting then destroyed the document **including its approved v1 snapshot**, with no audit entry. An approved controlled record could be made to vanish without a trace.

Protection now follows the record's **evidence, not its current state**: undeletable while `in_review`/`approved`, and for the rest of its life once any version has been approved (verified refused on all three delete paths — public mutation, WebDAV soft delete, REST hard delete). A never-approved draft still deletes as before, but now writes a `document.record_deleted` audit entry. Regression tests added; the affected suites (601 tests) and full `bun run check` (44/44) are green.

**⚠️ Upgrade note for deployments**

This branch adds two knowledge-corpus DDL migrations (`06_knowledge_private_document_scope`, `07_knowledge_private_team_ids`). The bundled knowledge database applies them via dbmate **at container start**; a BYO corpus gets them from `convex/knowledge/ddl.ts` on first touch. A deployment that upgrades the platform **without** restarting/migrating its knowledge database will fail every document ingest with:

```
Indexing failed: column "team_ids" of relation "documents" does not exist
```

Hit exactly this on the dev stack; applying the two migrations fixed it with no code change. Worth calling out in the release notes alongside the retrieval-scoping tightening.

PR #2961, opened 2026-08-11
feat(platform): add clarifying questions to chat and automation runs

Israeltheminer:
Chat can ask a clarifying question again, and this time the shape stops it becoming a form.

## Why

The AI-backend rewrite (#2857) drew a real boundary — chat retrieves, tasks produce deliverables — but it took "chat does not DO things" to mean "chat does not ASK things". The assistant lost every way to ask, so a vague request produced a confident guess. The seams for the answer were left standing and unused: a `human-input` message part, a `human_input_request` approval type with a thread index, a `waiting-input` generation status, and a manual test plan (`chat.md` F18) still specifying behaviour nothing could produce.

The shape that shipped before the rewrite asked badly. Four required fields stacked at once, each a blank textarea with its examples buried in a subtitle, mounted inline in the transcript so the reader had to scroll back to find it and then compose four answers before anything happened.

That format already supported `single_select` with options, and its tool description said, in capitals, never to present choices as plain text. Models wrote `textarea` for every field anyway.

## What changed

**Options are required; free text is the escape hatch.** Ten field types collapse to one shape with no free-text question type at all, so the old failure mode is unrepresentable rather than discouraged. `Other…` is injected by the client — always present, always localised, never model-authored.

**One question at a time.** Sets cap at four and options at four per question. A single question renders with no progress chrome: one-at-a-time lifts completion, but the gain lives at five-to-ten fields and stepping fewer costs more than it returns.

**It takes the input area, and never traps.** The panel occupies the composer's own frame. Esc collapses it to a one-line bar over a usable composer; Skip retires it outright; and simply saying something else retires it too. Picking advances, and the panel says so first (WCAG 3.2.2), with Back always available.

**The turn stops on a question.** `ask_question` is the fourth chat tool and returns no data — it pauses. The release notes for v0.2.91 record the previous gate being auto-retried straight past into acting on a guess, so the loop breaks on a *valid* call while a rejected one comes back as a correctable error the model can fix.

**Both lanes share one schema and one component**, so the chat composer and the automation-run card cannot drift. On the automation lane options stay optional: a run's blocker is often genuinely open ("what is the staging URL?"), and forcing four invented options onto that is worse than one honest box.

## Notable decisions

- **A question's lifetime is derived, not recorded.** It is outstanding only while the conversation has not moved past it, and "the person said something since" is already in the `messages` table. Nothing the reader sees depends on a bookkeeping write landing.
- **The answer is the person's next message**, so the transcript reads as a conversation. No hidden system message, no second resume path.
- **The transcript keeps one marker per ask**, not three. A tool whose purpose is to put something in front of someone is drawn as that thing, not as a step announcing it happened.
- **Only the exception is labelled.** "Answered" restated the message directly below it; a skip can be followed by nothing at all.

## Compatibility

No migration — the schema gate classifies the one new field (`automationHumanAsks.questions`) as data-safe. Existing asks carry no `questions` and render exactly the card that ships today. Pre-rewrite `human_input_request` approval rows do not parse as the new shape and are hidden rather than crashed on.

## Verification

`lint --type-aware` 0 · `typecheck` 0 · server + pii **74,778 tests** · ui **3,262 tests** · `migrations:check` clean · SAST 0 findings.

Behaviour is mutation-tested throughout — each guard was verified to fail when removed, after several passed against broken code.

Known flake, present on `main` too: `turbo run test test:ui` concurrently fails `test:ui`; both suites pass when run alone.


Israeltheminer:
Agreed — closing this.

The repositioning is the deciding point: if chat is Q&A rather than an agent, then asking the user questions is out of scope regardless of how the asking is done, and holding the tool count hard is the only way that line survives contact with plausible-sounding additions. This was one of those.

Two things from this branch are worth keeping separately, since neither is the chat tool:

- **#2962** — the turn loop writes a round's text twice when a turn stops mid-tool-execution. Pausing made it reproducible, but two cancel paths hit the same double-append, so it is live on `main` today. A verified fix and a regression test exist on this branch if useful.
- **#2963** — `canAccessThread` reads `threadMetadata`, which the rewritten chat never writes, so it denies every chat thread. The failure mode is "no data" rather than "no access", which is why it is invisible.

The tester complaint that prompted this (answering 3–5 questions at once with no support) still applies to the **automation-run** lane, where `ask_human` is an agent asking its operator. That half of the branch touches nothing in `lib/chat` or `convex/chat`. Happy to bring it back as a small PR if that is wanted — say the word, otherwise I will leave it.

Issue #2985, opened 2026-08-17
Bug: imap-smtp re-ingests the same attachment forever because it never binds it to a conversation


Israeltheminer:
## What happens

On a self-hosted deployment, the imap-smtp sync ingests an email attachment, writes the bytes to storage, and creates a `fileMetadata` row — but never binds the attachment to the conversation the email became. Two consequences, one of them silent data loss:

1. **The attachment is invisible in the inbox.** The conversation shows no file. An operator reasonably concludes the sender didn't attach one. On this deployment that meant a job applicant's CV was treated as never sent.
2. **Every poll re-ingests it.** With no binding there is no idempotency key, so the next sync sees the same attachment as new. A fresh blob and a fresh `fileMetadata` row, every cycle, forever.

## Scale observed

One deployment, roughly 11 days of running:

- **3,295 `fileMetadata` rows across 5 distinct filenames**
- a new storage blob every 5 minutes, each with its own `ObjectKey`, nothing consuming them
- most rows `failed`; ~8 permanently `queued`

That works out to ~288 orphaned blobs a day indefinitely.

## Why it is worse than duplicate rows

Those ~8 `queued` rows are **not** parked, so `countRagInFlight` (`convex/file_metadata/rag_dispatch.ts`) counts every one of them against `MAX_CONCURRENT_RAG_INDEXING_PER_ORG = 3`. The cap is therefore permanently saturated by a background job, and **every user-uploaded document parks instead of indexing**.

On the deployment where this was found, the oldest parked user upload was **7 days old**. Nothing a person uploaded had been indexed for a week, with no error shown anywhere. See #2986 for that half.

It is also self-sustaining: the RAG watchdog does clear stale rows every 5 minutes, but the sync mints a replacement on the same cadence, so the in-flight count never drops below the cap and `promoteQueuedRagJobs` never fires.

## Expected

- An ingested attachment is bound to its conversation and visible on it.
- Binding (or a stable external key derived from the message id + attachment) makes re-ingestion idempotent — a second poll of the same message must not create a second blob.
- Attachment ingestion must not consume the interactive indexing budget.

## Also worth checking

The re-ingested PDFs on this deployment were **4,445 bytes** — implausibly small for the document they claimed to be. The extracted attachment bytes may themselves be truncated or a stub, which would mean the content is wrong even once binding is fixed. Worth verifying extraction independently of this fix.

## Where to look

[...]

PR #3009, opened 2026-08-19
fix(platform): stop hiding an email attachment that carries a Content-ID

Israeltheminer:
An email attachment that carries a Content-ID now appears in the Inbox. Before this, it was stored, given a download URL, written into its message metadata, and then filtered out before render — so the file looked like it had never arrived.

## Why

The list hid anything carrying **both** a `contentId` and a `url`:

```ts
.filter((att) => !(att.contentId && att.url))
```

That treats a Content-ID as proof the part is an inline image already drawn in the body. It is not. Many mail clients stamp Content-ID on ordinary file parts, and `materializeEmailAttachments` gives every stored attachment a `url` — so both conditions were met by a perfectly ordinary attachment.

This is why a self-sent test worked while real mail did not: the test client omits Content-ID.

Reported from a live deployment. The `fileMetadata` row existed with a `storageId`, which is what narrowed it to display rather than to ingest.

## What changed

An attachment is hidden only when the body actually draws it — its cid appears in a `src="cid:…"` or `href="cid:…"` reference in the HTML.

- an unreferenced part stays visible whatever headers it carries
- a referenced inline image with no `url` yet stays visible, so it can be downloaded by hand when the auto-download fails
- `normalizeCid` moves to module scope, so the cid map and the reference test compare values the same way
- a reference is percent-decoded before comparison, because it can be escaped while the header value is not

## Tests

Six cases: an unreferenced attachment with a cid shows, a drawn inline image hides, both together behave correctly, no cid at all shows, a referenced image with no url shows, and an escaped reference still matches.

Mutations, all red: restoring the old `contentId && url` filter, hiding nothing at all, dropping the percent-decode, and no longer stripping angle brackets from a Content-ID.

## Scope

Display only. Ingest, storage and the `fileMetadata` row are unchanged.

Does not make email attachments searchable. They are still skipped for RAG by #2995, which was deliberate. Making them searchable needs the attachment bound to its conversation first (#2985), so visibility can be derived from the conversation's current assignment rather than stamped at ingest and left to drift.

### DIFF
--- a/services/platform/app/features/conversations/components/message.test.tsx
+++ b/services/platform/app/features/conversations/components/message.test.tsx
@@ -148,3 +148,126 @@ describe('Message — settled outbound', () => {
     expect(screen.queryByText(/message\.notDelivered/)).not.toBeInTheDocument();
   });
 });
+
+// A Content-ID is not proof that an attachment is inline. Many mail clients
+// stamp one on ordinary file parts, and the previous filter hid anything
+// carrying both a contentId and a url — so a real PDF from such a sender
+// vanished from the Inbox while a self-sent test from a client that omits
+// Content-ID displayed fine.
+describe('Message — attachment list vs inline images', () => {
+  function attachment(over: Record<string, unknown> = {}) {
+    return {
+      id: 'a1',
+      filename: 'CV.pdf',
+      contentType: 'application/pdf',
+      size: 23_359,
+      url: '/storage/blob_1/CV.pdf',
+      ...over,
+    } as NonNullable<MessageType['attachments']>[number];
+  }
+
+  it('shows an attachment whose cid the body never references', () => {
+    render(
+      <Message
+        message={makeMessage({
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Confirmed against the reported message on a live deployment. Its `metadata.attachments[0]` carries all three of `contentId`, `url` and `storageId`, so the old `contentId && url` filter hid it. The attachment was stored and reachable the whole time.

Two details worth recording.

The sender's client set a real Content-ID on an ordinary PDF part. `mailAttachmentsFromParsed` reads it from what the sender sent and does not fabricate one, so the connector is faithful and needs no change. Attachments sent from the platform's own compose path carry no Content-ID, which is why self-sent tests displayed correctly and inbound mail did not.

That message's body is empty — `content`, `html` and `text` are each a single newline. There is no HTML and therefore no `cid:` reference, so under this change the attachment shows. The old filter hid it on the header alone, with nothing in the body to justify treating it as inline.

Issue #3020, opened 2026-08-20
Improvement: browser Sentry events never group — convex/react's "[Request ID: …]" in console.error makes every failure a new GlitchTip issue


larryro:
## Problem

Any Convex call that fails repeatedly in the browser floods GlitchTip with **one issue per event** instead of one issue with N events. Concrete case: the deleted-org incident #3019 minted **100+ single-event issues** between 2026-07-18 and 08-11 — roughly a third of the demo project's total issue count at the time (100+ of 327) — all shaped like:

```
[CONVEX A(agents/file_actions:listAgents)] [Request ID: 5dabf348f1164725] Server Error
[CONVEX A(providers/file_actions:listProviders)] [Request ID: 91c0…] Server Error
[CONVEX A(branding/file_actions:readBranding)] [Request ID: …] Server Error
```

Noise at that volume buries real regressions — the #3019 incident itself sat unnoticed in that pile for a month.

## Why

1. convex/react logs every failed query/action via `console.error('[CONVEX A(module:fn)] [Request ID: xxxx] Server Error …')`;
2. `Sentry.captureConsoleIntegration({ levels: ['error'] })` promotes each such call to an event (`app/router.tsx:90`);
3. `Sentry.init` (`app/router.tsx:81-94`) has no `beforeSend`, no `ignoreErrors`, no fingerprint configuration;
4. the `[Request ID: …]` embedded in the message is unique per call, so message-based grouping never matches → a brand-new GlitchTip issue every time.

Nothing about this is specific to deleted orgs: **any** recurring browser-side Convex failure keeps minting issues at event rate until this is fixed.

## Suggested

Add a `beforeSend` that fingerprints console-captured Convex errors on the message minus the Request ID, and optionally drops codes the client already handles terminally (the server-side report for those exists anyway). Sketch to adjust:

```ts
beforeSend(event) {
  const msg = event.message ?? event.exception?.values?.[0]?.value ?? '';
  const convex = msg.match(/^\[CONVEX [QMA]\([^)]+\)\]/);
  if (convex) {
    // The Request ID is unique per call — strip it or grouping never matches.
    event.fingerprint = [msg.replace(/\[Request ID: [^\]]+\]\s*/, '').slice(0, 200)];
    // Optional: expected, client-handled outcomes add no signal as browser events.
    if (/ORG_NOT_FOUND|ORG_FORBIDDEN|UNAUTHENTICATED/.test(msg)) return null;
  }
  return event;
}
```

[...]

Issue #3019, opened 2026-08-20
Bug: every visit to /dashboard/<deleted org id> fires a burst of uncaught server errors — OrgSlugUnresolvableError / RLSError / ORG_NOT_FOUND


larryro:
## Problem

A browser session still holding a bookmark/deep link to a **deleted** organization has been hitting demo.tale.dev since 2026-07-18. Every visit fires the dashboard's initial round of org-scoped requests with the dead id, and each request dies on the server as an **uncaught** exception. This is live, not historical: the latest burst was **2026-08-19 17:30:51 UTC** — all four grouped GlitchTip issues below gained +3 events within that same second (one visit = one burst).

Every event carries the same deleted org id (`jh7csd7ks8740bza6qsxbz6sph7yegh2`) and culprit `/dashboard/`; the latest event is from an authenticated user session — a returning visitor with a stale bookmark or pinned tab, not a crawler and not a cron.

| GlitchTip issue | Error | Events |
|---|---|---|
| [7489511](https://app.glitchtip.com/tale/issues/7489511), [7489509](https://app.glitchtip.com/tale/issues/7489509) | `Uncaught OrgSlugUnresolvableError: [orgSlugFromId] no organization row found for id "jh7csd…"` | 16 each |
| [7489512](https://app.glitchtip.com/tale/issues/7489512) | `Uncaught ConvexError: {"code":"ORG_NOT_FOUND", …}` via `agents/file_actions:listAgents` | 16 |
| [7489510](https://app.glitchtip.com/tale/issues/7489510) | `Uncaught RLSError: Not a member of organization jh7csd…` | 16 |
| ~100+ single-event issues | `[CONVEX A(agents/file_actions:listAgents)] [Request ID: …] Server Error`, same for `listProviders`, `readBranding`, … (browser-side console capture; the per-call Request ID keeps them from ever grouping — separate ticket, see comments) | 1 each |

The single-event pile alone was roughly a third of the demo project's total issue count (100+ of 327 at the 2026-08-12 review) — enough noise to bury a real regression, which is why this ran for a month.

## Why every visit produces a burst (frontend recovery exists, but fires too late)

[...]

Issue #2013, opened 2026-06-23
Bug: workflow execution mutations/actions throw raw Error — execution management rejections are opaque Server Errors


yannickmonney:
## Area

`services/platform/convex/workflow_executions/mutations.ts`, `workflow_executions/actions.ts`, `workflows/executions/resume_debug_step.ts`, `workflows/triggers/slug_mutations.ts`.

## Steps to reproduce

1. Cancel an execution that is already in a terminal state.
2. Re-run an execution while unauthenticated.
3. Resume a debug step on a non-paused execution.
4. Create a schedule for a workflow that is not installed.
5. All rejections surface as opaque \"Server Error\".

## Expected

User-facing errors (not-found, wrong-state, unauthenticated, workflow-not-installed) should be `ConvexError({ code: ... })`.

## Actual

`workflow_executions/mutations.ts`:
- `line 36` — Execution not found
- `line 40` — Cannot cancel execution with status \"...\"

`workflow_executions/actions.ts`:
- `line 32/100/190` — Unauthenticated
- `line 108/198` — Execution not found

`workflows/executions/resume_debug_step.ts`:
- `line 31` — Execution not found
- `line 35-38` — Cannot resume a debug step on execution with status \"...\"
- `line 42` — Execution is not paused in debug mode
- `line 46` — Execution is missing its component workflow ID

`workflows/triggers/slug_mutations.ts`:
- `line 37` — Workflow is not installed
- `line 52/83/113/134/160/193/218/274/346/370/389` — Unauthenticated
- `line 55` — Invalid workflow slug
- `line 86/116/137/196` — Schedule/webhook/subscription not found

## Evidence

- `services/platform/convex/workflow_executions/mutations.ts:34-42`
- `services/platform/convex/workflow_executions/actions.ts:30-34`, `:98-101`, `:188-191`
- `services/platform/convex/workflows/executions/resume_debug_step.ts:30-47`
- `services/platform/convex/workflows/triggers/slug_mutations.ts:37-55`, `:83-86`, `:113-116`, `:160-163`, `:193-196`

## Notes

These are user-initiated operations from the workflow builder/debug UI. The client has no way to differentiate the error cause without `ConvexError`.

PR #3024, opened 2026-08-20
fix(platform): type RLS errors as ConvexError and name a deleted org

### LINKED ISSUE #3021: Bug: getOrganizationMember throws a bare RLSError — uncaught by Convex, and it says "Not a member" even when the organization was deleted
## Problem

Org-scoped queries and mutations guarded by the RLS helper crash with

```
Uncaught RLSError: Not a member of organization jh7csd7ks8740bza6qsxbz6sph7yegh2
```

— 16 events in [GlitchTip 7489510](https://app.glitchtip.com/tale/issues/7489510), from the deleted-org incident #3019. Two distinct defects:

1. **Uncaught.** `RLSError` and its subclasses (`UnauthorizedError`, …) extend plain `Error`, not `ConvexError` (`convex/lib/rls/errors.ts:8-24`). Convex therefore surfaces them to clients as an opaque redacted "Server Error" and records an uncaught exception server-side. The throw site: `convex/lib/rls/organization/get_organization_member.ts:140-144`.
2. **Wrong diagnosis when the org is gone.** If the organization row was deleted, the member lookup naturally finds nothing — and the error still claims "Not a member of organization …". Triage reads a permissions bug where the real state is "organization no longer exists". Exactly this misleading string sat in GlitchTip for a month during #3019.

## Where

`getOrganizationMember` is referenced across ~88 non-test files under `convex/` — public entry points like `provider_credentials/queries.ts:66-77` (`listCredentials`) route through it. The blast radius is why this is its own ticket rather than part of #3019.

## Suggested

At public boundaries, convert to typed `ConvexError`s distinguishing `ORG_NOT_FOUND` (org row gone) from `ORG_FORBIDDEN` (org exists, caller is not a member). Both house patterns already exist:

- `getCurrentMemberContext` catches `UnauthorizedError`, re-checks whether the org exists, and returns `{ status: 'not_found' | 'not_member' }` without throwing (`convex/members/queries.ts:73-112`);
- `requireOrgMembershipById` already emits `ConvexError` with `ORG_NOT_FOUND` / `ORG_FORBIDDEN` codes (`convex/lib/auth/require_org_membership.ts`).

Heads-up for the coordinated frontend pass: the layout error boundary decides retry behavior by string-matching error messages, including `Unauthenticated`/`UNAUTHENTICATED` (`app/components/error-boundaries/boundaries/layout-error-boundary.tsx:16-39`) — changing thrown types/messages needs a matching check there.

Related precedent: #2018 (same problem class for org/user lifecycle mutations — closed).

larryro:
Org-scoped queries and mutations guarded by the RLS helper crashed with `Uncaught RLSError: Not a member of organization jh7csd…` — 16 events in GlitchTip 7489510 from the deleted-org incident. Two defects in one throw site: the error was **uncaught** (a plain `Error`, redacted to "Server Error" for clients), and it **misdiagnosed** a deleted organization as a membership refusal, which is exactly what sent triage down the permissions path for a month.

Fixes #3021.

## What changed

1. **`RLSError` now extends `ConvexError<{ code, message }>`** (`convex/lib/rls/errors.ts`). The Convex runtime detects application errors via the `Symbol.for('ConvexError')` marker field, which subclasses inherit — so every RLS refusal that escapes a public function now reaches the client as structured `data` with a stable `code`, exactly like `requireOrgMembershipById`'s errors, instead of a redacted "Server Error" plus an uncaught-exception record. `message` is kept as the human sentence (ConvexError would stringify the data into it), so server logs and GlitchTip titles keep reading naturally.
2. **`getOrganizationMember` distinguishes the two failure states** (`convex/lib/rls/organization/get_organization_member.ts`). On an empty member lookup — and only then, so the happy path costs nothing — it re-checks the organization row and throws `UnauthorizedError` with `code: 'ORG_NOT_FOUND'` ("Organization … not found") when the row is gone, vs `code: 'ORG_FORBIDDEN'` ("Not a member of organization …") when the org exists. The disabled-member throw also carries `ORG_FORBIDDEN`, matching `requireOrgMembershipById`. The re-check is guarded by `looksLikeConvexDocumentId` (a sentinel/slug would throw inside the betterAuth component) and its own failure falls back to `ORG_NOT_FOUND` with a warn, mirroring `getCurrentMemberContext`'s catch.

## Why the blast radius is contained

`getOrganizationMember` is referenced across ~88 non-test files, so the design constraint was: don't change what existing catches see.

- All swallow sites catch by `instanceof UnauthorizedError` (approvals ×6, sandbox ×7, members ×3, team_members, tasks, governance/erasure) — the class hierarchy is unchanged, both new codes are still `UnauthorizedError`, so "non-member → return empty/null" behavior is byte-identical, including for deleted orgs (which already took this path).
- `getCurrentMemberContext` (`members/queries.ts`) keeps working unchanged — it catches `UnauthorizedError` and does its own org re-check, so the dashboard's `not_found` bounce is unaffected.
- `http.ts` boundaries catch broadly and map to 403 without echoing the error — unchanged.
- Frontend: `isStructuredConvexError` is a structural `data` check (no `instanceof`), so `useActionQuery` now correctly treats RLS refusals as terminal (no 3×-retry delay); the layout error boundary's transient matcher already deliberately matches the structured `{"code":"UNAUTHENTICATED"}` payload shape (see the #2013 comment in `layout-error-boundary.tsx`) — this PR makes prod behave like that comment always assumed.
- Callers constructing these errors are source-compatible: the `(message, code?)` signatures are supersets of the old ones.

One deliberate information tradeoff: a client can now distinguish "org gone" from "org exists, not yours" — the same distinction `requireOrgMembershipById` (and thus the whole agents/credentials surface) already exposes.

## Tests

`get_organization_member.test.ts` rewritten against the real error classes (it previously mocked the errors module): existing hot-path/mirror/email-fallback cases kept; new cases pin `ORG_FORBIDDEN` (org exists / disabled member, incl. the wire `data` payload), `ORG_NOT_FOUND` (row gone, with the `findOne` call asserted), the shape-guard short-circuit (no component read for a non-id-shaped org id), and the re-check-failure fallback.

Targeted suites over the catch sites: `convex/lib/rls` (84), plus members, team_members, sandbox, governance, tasks error-codes/stats, auth, http_jwks_cache — 584 tests / 50 files, green.

A **full `--project server` sweep** (8,057 tests / 657 files, run twice with identical results) then surfaced the two public-boundary tests that assert the refusal **message** end-to-end: `connector_credentials/queries.test.ts` and `provider_credentials/queries.test.ts` — their fixture org strings never existed as organization rows, so under the new distinction they correctly read "Organization … not found" where they expected `/Not a member/`. Adapted in the second commit: the membership-refusal case now seeds a real betterAuth organization row (the org must exist for "not a member" to be the true state) and keeps asserting `/Not a member/`; the unknown-org flavor is pinned as its own case (`/not found/`). Both credential suites green after: 109 tests / 9 files. The sweep's only other failure is `lib/mocks/contract/openai-compat.test.ts` (Prism-mock embeddings length) — pre-existing, fails identically on untouched `main` in this environment.

## Verifying after deploy

GlitchTip: `RLSError: Not a member of organization jh7csd…` (7489510) stops recurring for the deleted org; new occurrences of the same visit pattern surface as `RLSError: Organization … not found` application errors instead. See the cleanup checklist in #3019.

Related: #3019 (incident), #3020 (browser event grouping), #2018 (same problem class for lifecycle mutations, closed).

Gate: `tsc --noEmit` clean, `oxlint --type-aware` clean, oxfmt, targeted vitest 584 + full server sweep (sole remaining failure pre-exists on `main`).

### DIFF
--- a/services/platform/convex/connector_credentials/queries.test.ts
+++ b/services/platform/convex/connector_credentials/queries.test.ts
@@ -12,7 +12,7 @@
 import { convexTest, type TestConvex } from 'convex-test';
 import { describe, expect, it } from 'vitest';
 
-import { api, internal } from '../_generated/api';
+import { api, components, internal } from '../_generated/api';
 import type { Id } from '../_generated/dataModel';
 import betterAuthSchema from '../betterAuth/schema';
 import schema from '../schema';
@@ -100,6 +100,26 @@ async function seedMember(
       role: 'member',
       createdAt: 0,
     });
+  });
+}
+
+/** A real betterAuth organization row: a membership REFUSAL requires the org
+ * to exist — an unknown org id reports ORG_NOT_FOUND instead (#3021). The
+ * fixture ORG strings above deliberately have no row. Returns the created
+ * component-side document id. */
+async function seedOrganization(t: T, slug: string): Promise<string> {
+  return t.run(async (ctx) => {
+    const created = await ctx.runMutation(
+      components.betterAuth.adapter.create,
+      {
+        input: {
+          model: 'organization',
+          data: { name: slug, slug, createdAt: 0 },
+        },
+      },
+    );
+    // oxlint-disable-next-line typescript/no-unsafe-type-assertion -- adapter returns the created record as unknown
+    return (created as { _id: string })._id;
   });
 }
 
@@ -278,14 +298,30 @@ describe('listCredentials', () => {
 
   it('refuses a caller who is not a member of the organization', async () => {
     const t = newWorld();
+    const realOrg = await seedOrganization(t, 'ic-queries-real');
     await seedMember(t, MEMBER, OTHER_ORG);
+    await expect(
+      t
+        .withIdentity({ subject: MEMBER })
+        .query(api.connector_credentials.queries.listCredentials, {
+          organizationId: realOrg,
+        }),
+    ).rejects.toThrowError(/Not a member/);
+  });
+
+  it('reports an organization that does not exist as not found', async () => {
+    const t = newWorld();
+    await seedMember(t, MEMBER, OTHER_ORG);
+    // The fixture ORG string never had an organization row — the exact state
+    // a deleted org's stale id leaves behind. It must not read as a
+    // permissions bug (#3021).
     await expect(
       t
         .withIdentity({ subject: MEMBER })
         .query(api.connector_credentials.queries.listCredentials, {
           organizationId: ORG,
         }),
-    ).rejects.toThrowError(/Not a member/);
+    ).rejects.toThrowError(/not found/);
   });
 });
 
--- a/services/platform/convex/lib/rls/errors.ts
+++ b/services/platform/convex/lib/rls/errors.ts
@@ -1,26 +1,52 @@
 /**
  * RLS Error Types
+ *
+ * Every RLS refusal extends `ConvexError`, so a throw that escapes a public
+ * function is an APPLICATION error to the Convex runtime (detected via the
+ * `Symbol.for('ConvexError')` marker, which subclasses inherit): the client
+ * receives the structured `data` (`{ code, message }`) instead of a redacted
+ * "Server Error", and the UI can dispatch on `data.code` exactly as it does
+ * for `requireOrgMembershipById`'s errors. Server-side `instanceof` catches
+ * (approvals, sandbox, members, tasks, governance, …) keep working unchanged
+ * — the hierarchy is the same, only the base class changed. (#3021)
+ *
+ * `message` stays the human sentence (`ConvexError` would stringify the data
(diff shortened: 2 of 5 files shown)


larryro:
Closing as obsolete. The Convex tree this patched (`services/platform/convex/lib/rls/**`) was retired in #3125; the 0.5 backend already carries the behaviour this PR added — `backend/auth/membership.ts` `requireOrganizationMember` distinguishes `ORG_NOT_FOUND` (org row gone) from `ORG_FORBIDDEN` (exists, not a member / disabled), and refusals reach clients as coded HTTP errors, not redacted server errors.

PR #3025, opened 2026-08-20
fix(platform): retire automation triggers of a deleted organization

### LINKED ISSUE #3022: Bug: deleting an organization leaves its automation schedule triggers enabled — the minutely scan then crashes in orgSlugFromId forever
## Problem

Deleting an organization cascades a lot — audit log, personalization hard-delete, the org's file/config directory, `memberMirror` (`convex/organizations/delete_cleanup.ts`, `convex/auth.ts:531-547`) — but **`automationTriggers` rows are never disabled or deleted**.

If a deleted org leaves behind a schedule trigger with `enabled: true`:

- the minutely cron keeps firing it (`convex/crons.ts:40-46` → `scanScheduledTriggers`, which scans `by_kind_enabled` — `convex/automations/triggers.ts:79-110`);
- each run reaches the **throwing** slug resolver `orgSlugFromId` in `convex/automations/agent_host.ts:638` → one uncaught `OrgSlugUnresolvableError` per scheduled occurrence, **forever**, plus the wasted runs.

Contrast: the crawler already handles this state gracefully — `convex/knowledge/crawl_action.ts:377` resolves with `orgSlugFromIdOrNull` and skips deleted orgs.

## Context

Split from the deleted-org incident #3019. That incident's org happened to have no schedule triggers (its noise cadence was human-shaped, not cron-shaped), so this gap is latent today — but it becomes real the first time anyone deletes an org that has scheduled automations, and unlike #3019 it needs no returning visitor: it self-sustains on every scheduled occurrence.

## Suggested

- In the deletion cascade (`prepareOrganizationDeletion`, `convex/organizations/delete_cleanup.ts:20`, or the `/organization/delete` middleware), disable or delete the org's `automationTriggers`.
- Defense in depth: have the trigger scan / agent host resolve with `orgSlugFromIdOrNull` and auto-disable any trigger whose org no longer resolves, so a missed cascade can't error forever.
- While there, audit the other org-scoped tables (e.g. `websites`) for orphan-row policy after org deletion.

larryro:
Deleting an organization cascades a lot — audit log, personalization hard-delete, the org's config directory, `memberMirror` — but **`automationTriggers` rows were never disabled or deleted**. A surviving `enabled: true` schedule keeps coming due on the minutely scan (`crons.ts` → `scanScheduledTriggers`), and each started run crashes deep in the agent host's throwing slug resolver (`agent_host.ts` `orgSlugFromId`): one uncaught `OrgSlugUnresolvableError` per scheduled occurrence, forever, plus the wasted runs. Unlike the #3019 incident this needs no returning visitor — it self-sustains. The incident's org happened to have no schedule triggers, which is the only reason this gap is still latent.

Fixes #3022.

## What changed

Two layers, so a missed cascade can't error forever:

1. **Delete cascade** (`convex/auth.ts` + `convex/automations/triggers.ts`): the `/organization/delete` after-hook now calls the new `internal.automations.triggers.cascadeDeleteOrgTriggers` right next to `cascadeDeleteOrgMembersMirror` — i.e. only after Better Auth actually deleted the org row, so a failed deletion keeps a live org's schedules and webhook tokens intact. Deleting rather than disabling matches the platform's hard-delete posture for org-owned rows (memberMirror, personalization); the mutation is idempotent and indexed (`by_org`). It has its own catch + warn so a failure is neither mislabeled as a mirror failure nor fatal to the hook.
2. **Scan-side retirement** (`scanScheduledTriggers`): a *due* trigger now resolves its org via `orgSlugFromIdOrNull` (the same graceful pattern the crawler uses in `knowledge/crawl_action.ts`); if the org no longer resolves, the trigger is patched `enabled: false` with a warn and never starts a run. This retires the orphans that pre-date the cascade, and anything a failed hook leaves behind. The lookup runs only for due triggers (not per scan row), so the steady-state cost of the minutely scan is unchanged; a transient lookup failure still throws, and the scan simply retries next tick.

Webhook and event paths need no scan-side guard: their trigger rows die with the cascade, and a deleted row already yields the token-secrecy 404.

## Tests

`convex/automations/triggers.test.ts`:

- new: a due schedule whose org is gone is examined but not fired, its row flips to `enabled: false`, no run row exists, and the next scan examines nothing;
- new: the cascade deletes exactly the deleted org's triggers (schedule + webhook), leaves the other org's intact, and is idempotent;
- new: after the cascade, the dead org's webhook token gets the indistinguishable 404 and starts nothing;
- updated: the two existing scan tests now seed real betterAuth organization rows (`components.betterAuth.adapter.create`, the `arena_action.test.ts` pattern) — required because a schedule that should *fire* now needs an org that resolves; the fixture org names deliberately don't.

Full `convex/automations/` suite: 236 passed (14 files); `convex/auth.test.ts` 7 passed.

## Not in scope (follow-up candidates)

Other org-scoped tables (`automationVersions`, `automationRuns`, `websites`, …) still keep orphan rows after an org deletion — inert without a trigger, but worth an orphan-row policy audit as noted in #3022.

Related: #3019 (the incident this was split from), sibling PRs for #3019/#3020/#3021.

Gate: `tsc --noEmit` clean, `oxlint --type-aware` clean, oxfmt, vitest 243 passed (automations + auth).

### DIFF
--- a/services/platform/convex/auth.ts
+++ b/services/platform/convex/auth.ts
@@ -544,6 +544,22 @@ export const getAuthOptions = (ctx: GenericCtx<DataModel>) => {
                   internal.members.mirror_sync.cascadeDeleteOrgMembersMirror,
                   { organizationId: bodyOrgId },
                 );
+                // Same posture for the org's automation triggers: a surviving
+                // schedule row would keep coming due (and crashing its runs)
+                // forever (#3022). Own catch so a failure here isn't logged
+                // under the mirror's label — the schedule scan retires
+                // orphaned rows as the backstop either way.
+                try {
+                  await runCtx.runMutation(
+                    internal.automations.triggers.cascadeDeleteOrgTriggers,
+                    { organizationId: bodyOrgId },
+                  );
+                } catch (err) {
+                  console.warn(
+                    '[automations] trigger cascade after organization delete failed; the schedule scan will retire the rows',
+                    err instanceof Error ? err.message : err,
+                  );
+                }
               }
             } else {
               const returned = isRecord(mw.context.returned)
--- a/services/platform/convex/automations/triggers.test.ts
+++ b/services/platform/convex/automations/triggers.test.ts
@@ -18,7 +18,7 @@ import { convexTest, type TestConvex } from 'convex-test';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import type { Automation } from '../../lib/engine/core/types';
-import { internal } from '../_generated/api';
+import { components, internal } from '../_generated/api';
 import betterAuthSchema from '../betterAuth/schema';
 import schema from '../schema';
 import { cronMatches, dueOccurrence, parseCron } from './cron';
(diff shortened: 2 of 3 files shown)


larryro:
Closing as obsolete code: it patches `services/platform/convex/**`, retired in #3125. The underlying gap is NOT closed in the 0.5 backend — see the note on #3022, which stays open.

PR #3026, opened 2026-08-20
feat(platform): resume the task-agent conversation on later kicks

larryro:
## What

A task is one piece of work — later kicks of the same task (**Retry**, **Request changes**, an idle-agent @mention) now continue the predecessor's harness conversation via `--resume`, instead of opening a fresh conversation on a rebuilt 10-message brief. A new `projectAgentRuns` row is still minted per kick; the fresh brief on the preserved files remains the fallback whenever the handle does not bind. Steering, the watchdog, cancel, quota-park, and the harvest lanes are untouched; no UI change (resume is invisible — the buttons already say Retry / Request changes).

## How

- **Decision**: pure `resolveTaskKickResume` (`convex/tasks/task_kick_resume.ts`, table-tested) + a db half `resolveTaskKickStartArgs` shared verbatim by the only two start schedulers (`kickTaskAgentRun`, `wakeParkedTaskAgentRuns`). The predecessor is the task's latest terminal run **by this agent** that actually **launched** — never-launched rows (no handle, no op row) and other agents' rows say nothing about this session's box or conversation and are skipped. Never "latest conversation on the standing session": the session is per agent and two tasks can run on it at once.
- **Binds** = same agent + harness + sessionId + same sandbox incarnation: the run row stamps `agentSessionId` + `sessionCreatedAt` at its terminal mark; pre-stamp rows recover the handle from their own session op (point read), where op-purge-on-destroy is the incarnation guarantee (belt: `live.createdAt <= startedAt`). The start re-checks the stamp after the session ensure and downgrades to fresh on mismatch.
- **Dead-handle trap** (found in review by live-probing the CLI): a `--resume` of a missing conversation does not throw — the CLI emits one errored result **echoing the bad id back**, so naive stamping would re-arm the dead handle on every Retry forever. The stamp is withheld from launch-failure-shaped windows (`isResumeLaunchFailure`: terminal + errored + no content), and such a window relaunches **fresh in the same attempt under the same execId** (runnerd refuses duplicate ids only while an exec is live) with the inspect-the-box note.
- **Box semantics**: a settled predecessor sweeps its harvested leftovers even on resume — an unswept box re-harvests stale files every review cycle and starves the 64-file harvest cap. A failed/cancelled predecessor keeps the box (it may hold the only copy of unpublished work).
- **Between-runs comments**: the resume prompt carries the discussion delta since the predecessor's settle (brief entries gained `at`) — those comments were silently undelivered before.
- **Error surfacing** (root cause of the motivating incident): an errored turn's run error is now the harness's own final text. Observed live on dev: a subscription OAuth token revoked mid-run surfaced as the generic "the agent run failed"; the real `401 OAuth access token has been revoked` lived only in the op timeline.
- Side fixes: the stalled-turn reaper's no-op-row grace honors `updatedAt` (a wake of a long-parked run was reapable while its restart was in flight); `managed-settings.json` pins `cleanupPeriodDays: 365` (the CLI's ~30-day transcript janitor would silently break resume on older tasks — takes effect on the next sandbox image build).

Schema: two `v.optional` fields on `projectAgentRuns` — data-safe growth, `migrations:check` passes ("20 data-safe change(s)"), no migration per the gate's own policy. The handle stays off every public payload (all reads are explicit projections; pinned by a new test).

## Verification

- `bun run check` green from the repo root (format, lint incl. SAST, typecheck, all suites incl. i18n); `migrations:check` OK. New/updated suites: `task_kick_resume.test.ts` (decision table incl. never-session-latest, op-fallback, destroyed-incarnation, malformed-handle rows), `task_agent_runs.test.ts` (kick/wake carry `resume` per predecessor, never-launched skip, wake-uses-predecessor), `task_reviewer.test.ts` (request-changes resumes), `agent_run_prompt.test.ts` (resume prompt, launch-failure shape, failure-reason), `task_agent_run_op.test.ts` (handle off public payloads), `external_turn_bridge_exec.test.ts` (`--resume` argv passthrough; the per-harness goldens already freeze the flag for all 9 harnesses).
- **Live e2e on dev** (the real incident task — BGB.pdf→zh translation whose two runs died mid-work on subscription 401/429): clicking **Retry** started a run whose exec announced the **same** conversation id as the predecessor (`c2a38047…`, recovered via the op fallback — pre-feature rows), and the agent opened with "继续任务。上一轮两个试点代理都因账户级 429 限流失败…调整策略：翻译子代理改用 sonnet" — full memory of the prior turn, checked its 81 prepared chunks intact, and re-did none of the environment setup the previous fresh-brief retry had wasted ~10 minutes repeating. Also confirms the pinned CLI keeps the session id across `--resume`.
- Plan was cross-reviewed by four parallel review agents (codebase reality / races & lifecycle / security & tenant isolation / tests & conventions) before implementation; all blockers and majors they raised are addressed above (the dead-handle echo loop, the sweep-vs-resume split, the never-launched predecessor, the discussion delta, the stamp exec-guard, handle format validation).

## Notes

- After a suspected context poisoning, the operator flush for a resumed lineage is the agent's session **Destroy** (fresh incarnation ⇒ fresh conversations); a "Start over" button is deliberately out of scope for v1.
- Out of scope (per plan): compaction/context-length policy, steer/watchdog changes, persisting the handle on `tasks`, softening `tasks.subject.requestChangesDialog` copy.


larryro:
## External review (pr-3026-resume-review canvas) — second-confirmation + disposition

Each finding was independently re-verified by three verification passes against HEAD before acting. Landed in 1128088a8:

| # | Review said | Second-confirmation | Action |
|---|---|---|---|
| F1 | blocker: settled+binds sweeps the box the conversation remembers | **Confirmed, downgraded to major** — no data path loses anything (deliverables on `task.outputs` + re-staged read-only); the gap was the prompt's silence | Resume prompt now names the staged copies, states the box was emptied (only on the swept path — the kept-box path must not lie the other way), and restates same-name-replaces |
| F2 | major: 15-row scan cap → first-start sweep | **Confirmed (medium)** — precondition even weaker than claimed (any 15 non-matching rows) | The pure decision distinguishes `'unknown'` (exhausted) from `null` (drained): exhausted keeps the box + inspect note; drained still sweeps. Both pinned by tests |
| F3 | major: cancel→instant Retry = two execs, one conversation | **Partial (low-medium)** — mechanics real; claude's whole-line JSONL append + the launch-failure fallback bound the damage | A bound resume now reaps the predecessor's exec first (best-effort, the steer lane's own posture); the window shrinks from unbounded-on-lost-cancel to ≤ the kill grace |
| F4 | major: kept-box leftovers starve the 64-file harvest cap | **Partial (minor)** — leftovers after a failed run ARE the unpublished deliverables; over-cap skips are surfaced in the settle comment | Deferred (follow-up in the #3001 cap lineage: mtime-desc ordering before the cap) |
| F5 | major: root sweep can delete a sibling's loose files | **Partial (minor, pre-existing since #2894** — this branch had already narrowed it) | Root leg is now age-gated on the agent-turn deadline: a live turn's files are always younger than its own deadline, so hygiene can't race it |
| F6 | medium: `settledAt` delta bound drops during-run comments | **Confirmed** — non-mention comments and gave-up steers were silently lost | Bound moved to the predecessor's `startedAt`; a mid-turn duplicate is harmless, a dropped comment is not. Also tightens the op-fallback incarnation re-check to the pure fn's own rule |
| F7 | medium: docs oversell, de/fr calques, task-automation stale, manual observable | **(a) partial (b) refuted — the renders are native idiom (c) partial (d) confirmed** — prompts never render in transcripts, so the manual's observable was unexecutable | task-automation.md's two rework sentences carry the continuation fact (en/de/fr); manual F15/B2 rewritten to behavioral observables + the backend fallback-log backstop |
| F8 | medium: dead-handle predicate over/under-triggers, 409 linger | **Partial** — over-trigger real only for zero-content first-response failures; under-trigger speculative (no in-repo harness behaves that way); 409 settles once, no loop | Predicate **narrowed** (not widened): an errored empty window must have announced no id or the attempted id; a fresh-id first-response failure keeps its handle. (b)/(c) deferred |
| F9 | medium: stamps not cleared, no kill switch, English prompts | **Refuted as a defect** — destroy → new `createdAt` + purged ops + the start's re-check make stale stamps inert (three independent guards); prompts are model-facing and the repo keeps prompt text monolingual | No change; org-level resume toggle noted as a product follow-up |
| F10 | low: handle validation only on the kick path | **Confirmed (pre-existing call sites)** | Steer restart lane now validates with the same `isValidResumeHandle` (invalid → the existing fresh-restart branch); the automation ask-resume one-liner is flagged as a follow-up outside this PR's blast radius |
| F11 | low: session read ignores sessionId/org | **Confirmed (zero practical impact)** — kick-read and start-read agree by construction | Derived-sessionId pin added; org check rejected (C2's derivation argument holds) |
| C1/C2 | review's own closures | **Both sound** | `liveTaskAgentRun` now documents why `collect()` IS the double-kick OCC guard (after a failed run the range read is the only conflict surface) |

Not adopted from the review: splitting the deps/e2e commits out (they gate this PR's CI and are atomic fixes), and messages.yml entries for prompts (monolingual-prompt rule).

PR #3027, opened 2026-08-20
fix(platform): group browser Sentry events for repeated Convex errors

### LINKED ISSUE #3020: Improvement: browser Sentry events never group — convex/react's "[Request ID: …]" in console.error makes every failure a new GlitchTip issue

larryro:
Any Convex call that fails repeatedly in the browser floods GlitchTip with **one issue per event**: convex/react logs every failure as `console.error('[CONVEX A(module:fn)] [Request ID: <unique>] …')`, `captureConsoleIntegration` promotes each line to an event, and `Sentry.init` had no `beforeSend` / fingerprint — so the per-call Request ID defeats message grouping every single time. Concrete cost: the deleted-org incident #3019 minted **100+ single-event issues** (roughly a third of the demo project's total at review time), and the incident itself sat unnoticed in that pile for a month. Nothing about this is org-specific — any recurring browser-side failure keeps minting issues at event rate until grouping works.

Fixes #3020.

## What changed

New `app/lib/sentry/convex-console-events.ts`, wired as `beforeSend` in `app/router.tsx`:

- **Group by what failed, not which request failed first.** For `[CONVEX Q|M|A(…)]` lines the Request ID is stripped out of the message and the event is fingerprinted on the stripped text (capped at 200 chars). Rewriting the message too means the events group even where custom fingerprints aren't honoured, and issue titles stop embedding a random id.
- **Nothing is lost.** The id survives as a searchable `convex.request_id` tag, and `captureConsoleIntegration` already preserves the raw console arguments in `extra.arguments` (verified against `@sentry/core` 10.x sources: string-only console args go through `captureMessage` → `event.message`).
- **Console copies of client-handled outcomes are dropped.** `ORG_NOT_FOUND` / `ORG_FORBIDDEN` (the dashboard bounces the tab) and `UNAUTHENTICATED` (the session-rotation window the layout boundary deliberately retries, #2013) return `null` — the recovery is the feature, and the server-side report of the same failure exists anyway. Dropping applies **only** to `logger: 'console'` events: an error that actually reached a boundary is real user impact and is kept, just normalized the same way (its exception value is cleaned and fingerprinted).
- Foreign events pass through untouched.

## Risk

`beforeSend` runs on every browser event, so the helper is deliberately anchored (`^\[CONVEX …`) and allocation-light; non-Convex events take two cheap checks and exit. The drop-list is three exact codes the app demonstrably handles (`dashboard/index.tsx` / `$id.tsx` bounce; `use-action-query.ts` treats structured ConvexErrors as terminal). Server-side visibility of those failures is unchanged — Convex backend reporting is a separate pipeline.

## Tests

`app/lib/sentry/convex-console-events.test.ts`: two occurrences of one failure get the same fingerprint AND message with their ids preserved as tags; each handled code's console copy is dropped; a boundary exception with a handled code is kept but normalized; foreign message/exception events pass through untouched; the raw normalizer strips/returns the id and rejects non-Convex text.

## Verifying after deploy

Trigger the same failing call twice from a browser (e.g. a garbage org id against a public action) and confirm GlitchTip shows **one** issue with 2 events rather than two issues. Then bulk-resolve the single-event backlog per the checklist in #3019.

Related: #3019 (incident), sibling PRs for #3019/#3021/#3022.

Gate: `tsc --noEmit` clean, `oxlint --type-aware` clean, oxfmt, vitest 9 passed.

### DIFF
--- a/services/platform/app/lib/sentry/convex-console-events.test.ts
+++ b/services/platform/app/lib/sentry/convex-console-events.test.ts
@@ -1,0 +1,109 @@
+import type { ErrorEvent } from '@sentry/tanstackstart-react';
+import { describe, expect, it } from 'vitest';
+
+import {
+  convexConsoleBeforeSend,
+  normalizeConvexConsoleMessage,
+} from './convex-console-events';
+
+// #3020 regression: convex/react embeds a unique `[Request ID: …]` in every
+// failure it logs, so two occurrences of the SAME failure never grouped —
+// one GlitchTip issue per event. The beforeSend must make two such events
+// carry one identity, keep the id findable, and drop only the console copies
+// of outcomes the app already handles.
+
+const LIST_AGENTS_A =
+  '[CONVEX A(agents/actions:listAgents)] [Request ID: 5dabf348f1164725] Server Error';
+const LIST_AGENTS_B =
+  '[CONVEX A(agents/actions:listAgents)] [Request ID: 91c0aa10777d4e02] Server Error';
+
+// `ErrorEvent` distinguishes itself from transactions by a required
+// `type: undefined` — spell it out so these literals typecheck.
+function consoleEvent(message: string): ErrorEvent {
+  return { type: undefined, logger: 'console', message };
+}
+
+function exceptionEvent(value: string): ErrorEvent {
+  return {
+    type: undefined,
+    exception: { values: [{ type: 'Error', value }] },
+  };
+}
+
+describe('normalizeConvexConsoleMessage', () => {
+  it('strips the request id and hands it back separately', () => {
+    expect(normalizeConvexConsoleMessage(LIST_AGENTS_A)).toEqual({
+      message: '[CONVEX A(agents/actions:listAgents)] Server Error',
+      requestId: '5dabf348f1164725',
+    });
+  });
+
+  it('leaves non-convex text alone', () => {
+    expect(normalizeConvexConsoleMessage('ResizeObserver loop limit')).toBe(
+      null,
+    );
+  });
+
+  it('accepts a line without a request id', () => {
+    expect(
+      normalizeConvexConsoleMessage('[CONVEX Q(members/queries:list)] boom'),
+    ).toEqual({
+      message: '[CONVEX Q(members/queries:list)] boom',
+      requestId: undefined,
+    });
+  });
+});
+
+describe('convexConsoleBeforeSend', () => {
+  it('gives two occurrences of one failure the same identity', () => {
+    const first = convexConsoleBeforeSend(consoleEvent(LIST_AGENTS_A));
+    const second = convexConsoleBeforeSend(consoleEvent(LIST_AGENTS_B));
+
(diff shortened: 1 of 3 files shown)


larryro:
Closing as obsolete. The convex/react client that logged `[CONVEX …] [Request ID: …]` console errors is gone with #3125, and the grouping half of this change already lives on main: `app/lib/sentry-normalize.ts` strips the request id in `beforeSend` (wired in `app/router.tsx`).

PR #3031, opened 2026-08-21
fix(platform): collapsible uploads tree and sectioned setup gate

larryro:
## What

Three operator-reported UX defects around automation settings, verified live on the Swiss VAT return desk:

1. **Uploads panel listed the whole subtree flat.** It is now a real tree on the shared `file-tree-primitives` (same semantics as the project Files tab): folders collapsed by default, click = expand + pick as upload target, re-click = collapse + clear; `role="tree"` with keyboard navigation. With a declared `subdir` the panel lists **strictly that subtree** — the automation's own settings-folder records (`history-*.json` seeds, roll-ups) no longer surface as misfiled, deletable "uploads"; they stay managed in Knowledge.
2. **First-time setup gate was a wall of text with two disjoint action zones.** Each declared form now renders as a `BorderedSection`; long pack descriptions clamp to three lines behind a shared `SettingsFormDescription` Show-more toggle (reused by the edit dialog); the form is a real `<form id>` submitted from the dialog footer next to Cancel via the button's `form` attribute — one pinned action row.
3. New i18n keys (`settings.showMore/showLess`, `settings.uploads.treeLabel`) shipped in en/de/fr.

## Companion (configs repo)

tale-project/configs@afba342 — the vat-return-desk pack: migrated to the `/agent/*` sandbox lane from #3026 (every calculation had been dying as "the script wrote no result.json" = `PermissionError: '/user'`), removed the park-to-To-do node (failed rounds stay at In progress where the subject panel derives `stalled` → **Start again**), failure comments now carry the real exec error, and the refused-VAT tax-base call moved out of the prompt doctrine into a quantified operator ask. Deployed and E2E-verified as automation v17–v19: the previously always-failing task now converges — onboarding agent → one recommendation-first ask → recalculation PASS → In review with figures.

## Verification

- `test:ui` 3353 pass (incl. new tree/strict-subtree/submit-wiring tests), platform node suite 75 387 pass, typecheck + lint clean; pack suite 179 pass.
- Live-verified on dev: collapsed tree + expand/pick on real filed-returns folders, sectioned setup gate with Show more, footer submit, and the stalled → Start again → converged-to-review loop end to end.


larryro:
Cross-review round on the 40-reviewer report (`fix-automation-settings-ux-review-2026-08-21-report.md`): 7 verification agents adjudicated every finding against the branch before anything was changed.

**Confirmed and fixed in 8227f1c0c** — all 8 must-fix items (FormDialog submit propagation — a nested create-folder dialog could save the whole setup gate through React-tree portal bubbling; missing `noValidate`; silent validation refusals now scroll+focus the first offending field via `SettingsSaveResult.firstInvalid`; `ensureRoot` single-flight + `FOLDER_DUPLICATE_NAME` adoption; `isLoading` gates; file rows as navigable treeitems + roving-tabindex fallback so the tree is Tab-reachable; strengthened tests), plus the confirmed should-fixes: stale-pick reconciliation, upload-in-flight blocks save, Cancel locked while saving, tree frozen for mouse AND keyboard while busy, `Tabs keepMounted` so the edit dialog's panels survive tab switches, per-instance `DropZone` input id, `aria-expanded` on Show more, and the stale per-form-Save sentence in `docs/{en,de,fr}/platform/automations/catalog.md`.

**Verified-refuted, left as is** — edit-dialog description clamp (progressive disclosure, full text one click away), lifting `useSettingsEditor` into the task modal (the `form`-attribute seam is the native pattern and both claimed failure windows are already guarded), `useId` for the setup form id (single mount site by construction), `Id<'folders'>`→`toId` widening (the documented house pattern), tree label wording and the h-6 toggle (both match established house idioms), folder-count chips and a tree height cap (design calls, would need Files-tab parity), reusing `ProjectCreateFolderDialog` (its contract diverges: parent preselect + lazy chain creation).

Live-verified on dev on top of the suites: creating a folder from inside the setup gate no longer saves the gate underneath (this drive also caught and fixed a busy-mirror render loop the jsdom suite couldn't see). Platform suites, @tale/ui (1171), and docs checks green.

PR #3040, opened 2026-08-22
feat(platform): serve agent nodes on subscription credentials

larryro:
## Why

An automation `agent` node resolved its model through the direct-only connector walk (`resolveServingTarget`), so an org whose only credential is a vendor subscription — e.g. an Anthropic **subscription broker** — failed every agent turn with:

> agent: no configured provider serves model "claude-fable-5" — an llm node's model must be listed in a connected provider's catalog and permitted by its credential

The same model worked on a pinned project agent, whose task lane already has a subscription lane. `docs/en/platform/agents/harnesses.md` already promises subscription serving for "a project agent or an automation agent node" — the code was behind the docs.

## What

- **New `convex/lib/providers/agent_serving.ts`** — the shared serving core:
  - `walkDirectServing`: the direct-credential connector walk, extracted from `llm_call.ts` and delegated back to, so the llm-node walk and the agent walks can never drift (error messages byte-identical).
  - `sanctionSubscriptionHarnessTurn` + `subscriptionApiBaseUrl`: the subscription sanction the task lane used inline (provider still declares the auth method → `resolveExecution` forces the exact harness → harness yml declares a `subscription` delivery). `tasks/task_serving.ts` now calls the shared helpers — behavior and error strings unchanged, its suite untouched and green.
  - `resolveWorkflowAgentServing`: the agent node's unpinned two-lane walk. **Direct pass first** — an org that serves the model directly today keeps that serving and its billing lane (no silent swap onto a broker) — then the subscription pass over the same fetched credential rows (no double queries). Nothing serving fails the node with every subscription refusal named, e.g. a forced-harness mismatch.
- **`automations/agent_host.ts`** — kick, start, and answered-ask resume resolve two-lane; new `mintWorkflowTurnAuth` (shared by start + resume) keeps the session virtual key on the gateway lane and, on the subscription lane, redeems the provider's default credential **inside the start action** (secrets never ride scheduler args), hands the vendor token to the harness CLI, and keeps the capability bridge on a random `tale-sub-` session token — no VK, no gateway budget, nothing to revoke at settle (`releaseTurnKey` already handles the no-key case). Vision polyfill stays gateway-only, mirroring the task lane.
- **Deploy-safe**: `startWorkflowAgentTurn`'s new `lane`/`apiBaseUrl` args are optional — in-flight scheduled starts from before the deploy keep their exact gateway behavior.
- **Out of scope, by design**: `llm` nodes stay direct-only — a subscription credential is bound to its vendor harness and never serves a raw API call. A broker-only org still needs an api-key/env credential for `llm` nodes.
- No schema/table change (cursor shape untouched; the optional action args are transient), no migration, no UI strings, no docs change needed.

## Verification

- New `agent_serving.test.ts` (11 cases): broker-only org serves on the subscription lane with the right `apiBaseUrl`; a direct serving is never stolen by an earlier broker connector; equivalent-id catalog spelling; dedicated coding endpoint (subscription-key); forced-harness mismatch, missing subscription delivery, allowlist, catalog miss, unreachable catalog; single credential query per connector across both passes.
- Full `convex/automations/` (242 tests), `convex/tasks/` (446), `convex/sandbox/` + `external_turn_shared` (155) green; lint (type-aware) and typecheck clean; pre-commit SAST 0 findings.
- **Still open**: live run on a broker-only org (deploy to the demo instance, or a local org pointed at the same broker) — the resolution + mint reuse the primitives the task lane's subscription turns already run on in production.


larryro:
Two additions since the opening description:

**`305729401` fix: `agent_serving.ts` was missing `'use node'`.** vitest/tsc/lint all pass without it, but a real `convex dev` push fails ("Could not resolve node:path" — the module imports the node-only catalog/provider loaders), so the branch as originally pushed would not deploy. Found while live-verifying against the local backend; no CI gate covers this class.

**`03748fe00` feat: automation delete entry.** The detail page gains a delete action (admin/developer-gated) with a confirm dialog, backed by a transactional cascade in `store.ts`: versions + deployment + triggers + project bindings. Run history is kept (audit; the workflowLog retention sweep owns it), and deletion is refused while a run is queued/running/waiting. A new `automationTombstones` table stops the default-pack seeder from resurrecting a deleted builtin on the next deploy — saving the name again clears the tombstone. Locales en/de/fr + editor docs ×3; new convex-test suite covers the cascade, the active-run refusals, tenant isolation, and the reseed/revive loop; two new UI tests cover the confirm flow and the blocked-delete refusal.

Live-verified on the local stack end to end: created a throwaway automation with an armed schedule trigger, deleted it, confirmed via direct table reads that versions and trigger rows are gone and the tombstone row carries the deleting user.

PR #3103, opened 2026-08-29
fix(platform): restore pdf image detection on scanned files

### LINKED ISSUE #3018: Bug: pdf.js operator-list read fails on every page it indexes
PDF indexing logs `Failed to read PDF operator list: this._intentStates.getOrInsertComputed is not a function`, repeatedly, during `uploadFileToRag`.

## What it is

`getOrInsertComputed` is a recent `Map` method. pdf.js is calling it and the Convex isolate does not provide it, so the operator-list read fails per page.

## Impact

Degraded rather than broken. On a live deployment the affected PDFs still produced chunks with sensible character counts and embeddings, so text extraction is falling back to a path that works. But the failure is silent to the operator, fires once per page, and the fallback's quality is unmeasured — a PDF whose text lives in the operator list rather than in the text layer would extract worse, or empty, with only a warning to say so.

The same log also carries `Warning: Please use the 'legacy' build in Node.js environments`, which suggests the non-legacy pdf.js build is being loaded where the legacy one is intended.

## Shape

Establish which pdf.js build is bundled and whether the legacy build resolves both messages. Then decide whether the operator-list path is needed at all for text extraction, or whether the fallback is the intended route and the warning should be silenced rather than fixed.

Worth a measurement before a change: compare extracted character counts for a PDF that currently warns against the same file extracted with the legacy build.

Israeltheminer:
Scanned PDFs index thin or empty because image detection never runs. This restores it.

Closes #3018.

## Why

pdfjs 5.x calls `Map.prototype.getOrInsertComputed`, which Node 22 does not ship, so the operator-list read throws once per page:

```
Failed to read PDF operator list: this._intentStates.getOrInsertComputed is not a function
```

The damage is narrower than the log suggests, and worse where it lands. Text-layer extraction (`getTextContent`) still runs, so an ordinary digital PDF indexes fine and only produces noise. But the operator list is what finds embedded images and decides "scanned page → OCR / describe". Without it a scanned or image-heavy PDF never reaches OCR, and indexes as thin or empty text with nothing but a warning to say why.

## What changed

One shim beside the existing ES2025 fills in `pdfjs_dom_polyfill.ts`, guarded like the rest so a real implementation wins once the runtime moves to Node ≥24.

Scope is `Map` only, from the library rather than from the proposal: every call site in both the main and worker builds — `_intentStates`, `methodPromises`, `objs`, `_cachedBitmapsMap`, `popupToElements` — is a plain `Map`. No `WeakMap` call exists, so none is shimmed.

Deliberately NOT the legacy pdfjs build. That would resolve the same call, and the loader already avoids it for Convex bundle size.

## Risk

**Silencing the warning would have been the wrong fix.** Suppressing it without restoring the operator list leaves scanned PDFs indexing badly with nothing to explain it — a quality loss that no longer announces itself.

**Presence, not truthiness.** A callback returning `undefined` still inserts, so a second call does not recompute. Getting this wrong would be invisible until a cache entry legitimately held `undefined`.

**The callable check runs before the lookup**, as the proposal specifies. Without that ordering a present key returns happily and a bad callback goes unnoticed until some later call happens to miss.

## Tests

Seven cases: the method is installed, compute-and-insert, an existing value returned without calling the callback, `undefined` stored as a real value, the computed value winning when the callback inserted the key itself, a non-callable callback rejected for both a present and an absent key, and a real implementation left alone.

Four deliberate breakages, all caught. One survived the first round: dropping the callable check still threw, because calling a string throws `TypeError` anyway — the case that distinguishes them is a present key, which is now tested.

The last case installs a stand-in for a native implementation and removes it in a `finally`, because the guard correctly refuses to replace it and every later suite in that worker would otherwise inherit it.

## Scope

Does not change extraction, OCR or chunking. Scanned PDFs indexed before this reach the corpus with whatever thin text they produced; re-indexing them is a separate step.

Gate: repo-wide typecheck, `oxlint --type-aware`, oxfmt, knip, SAST 0 findings, platform suite 75,819 passing.

### DIFF
--- a/services/platform/convex/lib/knowledge/extraction/pdfjs_dom_polyfill.test.ts
+++ b/services/platform/convex/lib/knowledge/extraction/pdfjs_dom_polyfill.test.ts
@@ -61,3 +61,112 @@ describe('installPdfjsDomGlobals base64 codecs', () => {
     expect(new TextDecoder().decode(plaintext)).toBe('integration-secret');
   });
 });
+
+describe('installPdfjsDomGlobals Map.getOrInsertComputed', () => {
+  it('installs the method pdfjs calls on its Maps', () => {
+    // Its absence is what broke the operator-list read per page (#3018).
+    installPdfjsDomGlobals();
+    expect(typeof Map.prototype.getOrInsertComputed).toBe('function');
+  });
+
+  it('computes and inserts when the key is absent', () => {
+    installPdfjsDomGlobals();
+    const map = new Map<string, number>();
+    const calls: string[] = [];
+    const value = map.getOrInsertComputed('a', (key) => {
+      calls.push(key);
+      return 1;
+    });
+    expect(value).toBe(1);
+    expect(map.get('a')).toBe(1);
+    expect(calls).toEqual(['a']);
+  });
+
+  it('returns the existing value without calling the callback', () => {
+    // pdfjs relies on this: the callback builds an intent state, and calling
+    // it twice would discard the one already in flight.
+    installPdfjsDomGlobals();
+    const map = new Map<string, number>([['a', 1]]);
+    let called = false;
+    const value = map.getOrInsertComputed('a', () => {
+      called = true;
+      return 2;
+    });
+    expect(value).toBe(1);
+    expect(called).toBe(false);
+  });
+
+  it('stores a value the callback returns even when it is undefined', () => {
+    // Presence, not truthiness: a second call must not recompute.
+    installPdfjsDomGlobals();
+    const map = new Map<string, undefined>();
+    map.getOrInsertComputed('a', () => undefined);
+    expect(map.has('a')).toBe(true);
+    let recomputed = false;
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Superseded by #3170, which carries the same fix against the current tree.

Both files this branch touches were deleted by #3125. The polyfill module itself was re-homed to `backend/core/lib/knowledge/extraction/pdfjs_dom_polyfill.ts`, so the fix is now a small addition to `ensureEs2025Shims` rather than a new block — it sits beside the `Promise.try` and base64 fills that already exist for the same reason.

#3018 stays open until #3170 lands.

One thing worth recording from redoing it: Bun ships `Map.prototype.getOrInsertComputed` natively and Node does not. The suite runs under Node, so the assertions exercise the real gap — but a test run under Bun would have passed against the native implementation and proven nothing.

Israeltheminer:
Closing: #3170 carried this fix and is merged. #3018 is closed by it.

PR #3105, opened 2026-08-29
fix(platform): stop a search returning one passage twice

### LINKED ISSUE #3061: Bug: a search can return the same passage twice
A knowledge search can spend two of its result slots on the same passage, because two corpus rows can hold byte-identical content.

## Observed

On one deployment, a corpus of nine rows contained two pairs of duplicates. Each pair shared its `content_hash`, its chunk count, and its total character count, while carrying a different `file_id`:

- one pair at 64 chunks / ~105k characters
- one pair at 9 chunks / ~13.5k characters

Both members of each pair were `status = 'completed'`, unscoped, and at the folder root. So the same text is retrievable twice, under two refs.

## Why it happens

Not in the corpus writer, which handles this deliberately and well. `indexDocument` computes a `content_hash`, calls `findDuplicate` for a completed same-org row with identical content, and clones that row's chunks instead of re-embedding. So a duplicate costs storage, not an embedding run.

Cloning is also right in general: two documents may legitimately hold the same bytes at different scopes — the same PDF filed in two projects — and each needs its own row because retrieval decides scope per row.

The duplication starts upstream. Two separate blobs were created for identical content, so two `fileMetadata` rows and two corpus rows follow. Nothing at upload compares content before storing.

## The cost that is a defect

Retrieval does not deduplicate by content. `searchKnowledge` filters hits by retrievable ref and returns what is left, so two rows holding the same text both match and both consume a slot in a bounded result set. A caller asking for eight passages can receive the same passage twice and lose a distinct one that would have ranked ninth.

That is a relevance defect rather than a disclosure one: nothing is exposed that the caller could not already read.

## Shape

Two independent directions, either of which closes the retrieval symptom:

- Deduplicate results by `content_hash` before returning them, keeping the highest-scoring ref. Narrow, and does not touch ingest.
- Deduplicate at upload, so identical content does not become a second blob. Wider, and also reclaims storage.

The first is the smaller change and fixes what a user sees. The second addresses the origin, and overlaps the existing duplicate-`fileMetadata` cleanup question.

## Scope

Does not cover the separate, documented behaviour where replacing a document leaves its previous corpus row in place. That is deliberate — retrieval validation hides the stale row — and is not what this issue is about.

Israeltheminer:
A search no longer spends two of its result slots on the same passage.

Closes #3061.

## Why

The same text can sit in the corpus twice — a file uploaded as two documents, or a paragraph two documents share. Both copies match, and a bounded result set then says one thing twice while a different answer falls off the end. A caller asking for eight passages can receive seven distinct ones.

Observed as duplication on a deployment: two pairs of corpus rows with matching content hashes, chunk counts and character counts under different refs. The retrieval symptom is reasoned from the code rather than seen there, because both copies of each pair were orphaned and already hidden.

## What changed

`searchKnowledge` drops a passage the caller has already been given, keeping the best-scoring one. Fusion has already sorted by score, so the first occurrence is the best.

**Keyed on the passage, not the document.** The issue proposed deduping by `content_hash`, which the corpus query does not select — plumbing it through the SQL, the row type and fusion would be a lot of moving parts for the same outcome. A result slot holds a passage, so that is what is compared. It catches the whole-document case too, since identical documents produce identical chunks, and it also catches a shared paragraph between otherwise different files.

Whitespace is normalized, so two copies that differ only in how their source wrapped lines still count as one.

Corpus is part of the key: a web page and a document saying the same thing are two findings, and only one is citable by URL.

## Risk

**It runs after the retrievability filter, deliberately.** Deduping first could keep a copy the caller cannot read and drop the readable one — the gate would then remove what was kept, and the passage would vanish entirely. That is worse than showing it twice, and it is the case a reader of this code is most likely to get wrong. A test pins it.

**A second chunk of the same document survives**, because it is a different answer. Deduping per document would have lost it.

## Tests

Five cases: a duplicate pair collapsed to the higher-scoring ref, two distinct passages from one document both kept, copies differing only in whitespace collapsed, the readable copy kept when the better-scoring one is denied, and identical text across different corpora kept separate.

Four deliberate breakages, all caught — including moving the dedupe ahead of the gate, which is the ordering the test exists for.

## Scope

Does not stop duplicates entering the corpus. Identical content still becomes two rows, still costs storage, and the corpus writer still clones chunks for it rather than re-embedding. This only stops the duplication reaching a reader.

Gate: repo-wide typecheck, `oxlint --type-aware`, oxfmt, knip, SAST 0 findings, platform suite 75,817 passing.

### DIFF
--- a/services/platform/convex/knowledge/search.test.ts
+++ b/services/platform/convex/knowledge/search.test.ts
@@ -97,3 +97,115 @@ describe('searchKnowledge live document validation', () => {
     });
   });
 });
+
+describe('searchKnowledge — the same passage twice', () => {
+  beforeEach(() => {
+    retrieveMock.mockReset();
+  });
+
+  /** Two refs holding identical text — the same file indexed twice. */
+  function duplicatePair(text: string) {
+    return [
+      { ...hit('documents', 'copy_a'), text, fusedScore: 0.9 },
+      { ...hit('documents', 'copy_b'), text, fusedScore: 0.4 },
+    ];
+  }
+
+  const ACCESS = { teamIds: [], projectIds: [], includeHub: true };
+
+  it('returns one copy, keeping the higher-scoring ref', async () => {
+    // A bounded result set spending two slots on one passage pushes a
+    // different answer off the end.
+    retrieveMock.mockResolvedValueOnce({
+      hits: [...duplicatePair('Refunds within 30 days.')],
+      diagnostics: {},
+    });
+    const runQuery = vi.fn(async () => ['copy_a', 'copy_b']);
+    const result = await searchKnowledge({ runQuery } as never, {
+      organizationId: 'org_1',
+      orgSlug: 'acme',
+      query: 'refunds',
+      access: ACCESS,
+    });
+    expect(result.hits.map((h) => h.source.ref)).toEqual(['copy_a']);
+  });
+
+  it('keeps a distinct passage from the same document', async () => {
+    // Deduping is per passage, not per document — a second chunk of the same
+    // file is a different answer.
+    retrieveMock.mockResolvedValueOnce({
+      hits: [
+        { ...hit('documents', 'doc'), text: 'First passage.' },
+        { ...hit('documents', 'doc'), text: 'Second passage.' },
+      ],
+      diagnostics: {},
+    });
+    const runQuery = vi.fn(async () => ['doc']);
+    const result = await searchKnowledge({ runQuery } as never, {
+      organizationId: 'org_1',
+      orgSlug: 'acme',
+      query: 'passages',
+      access: ACCESS,
+    });
+    expect(result.hits).toHaveLength(2);
+  });
+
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Superseded by #3172, which carries the same fix against the current tree.

Both files this branch touches were deleted by #3125; `convex/knowledge/search.ts` is now `backend/core/knowledge/search.ts`. The fix transplants unchanged — same helper, same key, same placement after the retrievability filter.

Three assertions were added beyond this branch's two, because the mutation pass showed the original pair did not cover all four properties: the corpus is part of the key (so the same wording in a document and on a crawled page stays two answers), and the after-the-gate ordering needs its own case, since deduping first only bites when one copy is unreadable.

#3061 stays open until #3172 lands.

Israeltheminer:
Closing so the queue reflects reality — #3176 replaces this and is open. Both of this branch's files were deleted by #3125, so it cannot be rebased. #3061 stays open and is closed by #3176.

PR #3122, opened 2026-08-31
fix(platform): erase the subject's cloud grants, sync configs and blobs

### LINKED ISSUE #3120: Bug: GDPR erasure leaves the subject's cloud tokens, sync configs and upload blobs behind
A GDPR erasure leaves six of the subject's tables untouched, including their stored OAuth access and refresh tokens for Google Drive and OneDrive. The receipt reports `done`.

## Why

The 0.4 cascade ran 19 `eraseSubject*` arms. The 0.5 cascade in `backend/domains/erasure/service.ts` runs 10 passes. These tables hold subject rows in 0.5 and no pass reaches them:

| Table | What the subject leaves behind |
| --- | --- |
| `app.user_cloud_authorizations` | sealed OAuth2 access + refresh token per (org, user, provider) |
| `app.cloud_import_oauth_states` | an in-flight authorization keyed to the user |
| `app.onedrive_sync_configs` | a sync that keeps running under their grant |
| `app.google_drive_sync_configs` | the same, for Drive |
| `app.video_link_jobs` | pasted URL, video title, transcript blob ref |
| `app.file_metadata` blobs | the `uploads` pass deletes the row, never the object |

`grep -rn "DELETE FROM app.user_cloud_authorizations" services/platform/backend` returns nothing outside the integration harness. Same for the other four tables.

The token one is the sharpest. `0042_cloud_import.sql` describes the row as an intentional Documents-import grant, sealed with the same envelope connector credentials use. It outlives the subject's membership, and erasure is the only thing that was supposed to take it.

The blob one is quieter but the same class. `purgeDocument` deletes the corpus entry and then the blob. The `uploads` pass deletes only the row, so a bare upload's bytes stay in object storage and its chunks stay in the corpus.

## Where it came from

The cascade arms for cloud grants and Drive sync were added on 24 Aug in #3046 and #3048, six days before #3107 replaced the Convex backend. Both PRs are otherwise fully ported — `cloud_import` and `google_drive` are marked done in `backend/MIGRATION.md`. Only the erasure arms were left behind, and `MIGRATION.md` has no note deferring them.

## Scope

This is erasure completeness, not access control. The rows are correctly scoped while the subject is a member; they simply survive a request to erase them.

Legal holds already gate the whole cascade at `processErasure`, so a per-pass hold check is not needed.

Not covered here: `eraseSubjectPolicyAcknowledgements`, `eraseSubjectTwoFactorAttempts`, `eraseSubjectAutomationRuns` and `eraseSubjectWfExecutions` from 0.4. Their 0.5 tables need checking before anyone claims cascade parity.

Israeltheminer:
A GDPR erasure now removes the subject's cloud-import OAuth grant, their Drive and OneDrive sync configs, their video-link jobs, and the blobs behind their bare uploads. All of those survived a request that reported `done`.

Closes #3120.

## Why

The 0.4 cascade ran 19 `eraseSubject*` arms. The 0.5 cascade runs 10 passes. Six categories had no pass at all:

| Table | What survived erasure |
| --- | --- |
| `app.user_cloud_authorizations` | sealed OAuth2 access + refresh token |
| `app.cloud_import_oauth_states` | an in-flight authorization |
| `app.onedrive_sync_configs` | a sync still running under that grant |
| `app.google_drive_sync_configs` | the same, for Drive |
| `app.video_link_jobs` | pasted URL, video title, transcript ref |
| object storage | every bare upload's bytes |

The token row is the sharpest. `0042_cloud_import.sql` describes it as an intentional Documents-import grant, sealed with the same envelope connector credentials use. It outlives the subject's membership.

The blob is the same class of miss. `purgeDocument` deletes the corpus entry and then the blob; the `uploads` pass deleted only the row.

## What changed

- `purgeBlobArtefacts` in `retention/service.ts` — the corpus-then-blob pair lifted out of `purgeDocument`, which now calls it. One implementation, two callers.
- `uploads` purges each file's artefacts before deleting its row.
- Three new passes: `videoLinks`, `cloudGrants`, `syncConfigs`.

`video_link_jobs` gets its own pass rather than folding into `uploads`: the job can own a blob when the linked `file_metadata` row never landed, and a welcome-page paste carries no thread, so neither the uploads pass nor the thread cascade reaches it.

Sync configs go because they name the member whose grant the sync runs under. Documents those syncs imported stay; that is org content, reached by the `documents` pass when the subject created it.

## Risk

Each new pass loops with a knowledge-database round trip per row, where `uploads` was a single statement. The `documents` pass directly above it already has that shape, and the cascade runs as a background job.

## Tests

`integration-check.ts` seeds all six categories for the erased subject and counts them into `leftovers`, which the existing assertion requires to be `0`.

Observed against a throwaway Postgres, driving `processErasure` directly:

| Cascade | Result |
| --- | --- |
| this branch | 0 rows survive; receipt `done`; counts `uploads:1 videoLinks:1 cloudGrants:2 syncConfigs:2` |
| `origin/main` | 5 rows survive; receipt still `done` |

The blob purge itself is not asserted. Observing it needs MinIO, which would not pull in this environment. It rides `purgeBlobArtefacts`, the same path `purgeDocument` already uses under CI's S3 lanes.

## Scope

Four 0.4 arms still have no 0.5 pass: `eraseSubjectPolicyAcknowledgements`, `eraseSubjectTwoFactorAttempts`, `eraseSubjectAutomationRuns` and `eraseSubjectWfExecutions`. Their 0.5 tables need checking before anyone claims cascade parity. #3120 records that.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check` and `lint:sast` green.

### DIFF
--- a/services/platform/backend/domains/erasure/service.ts
+++ b/services/platform/backend/domains/erasure/service.ts
@@ -1,10 +1,12 @@
 import type { Sql, TransactionSql } from 'postgres';
 
+import { isRecord } from '../../../lib/utils/type-utils.ts';
 import { isAdminRole } from '../../auth/membership.ts';
 import {
   ERASURE_REASON_CODES,
   ERASURE_WATCHDOG_TIMEOUT_MESSAGE,
 } from '../../core/governance/erasure_constants.ts';
+import { normalizeAuthEmail } from '../../core/lib/auth/normalize_auth_email.ts';
 import { parseBlobRef } from '../../core/lib/storage/blob_ref.ts';
 import { toJson } from '../../db/sql.ts';
 import { addJobInTx } from '../../jobs/enqueue.ts';
@@ -31,6 +33,12 @@ import { writeNotificationForOrgs } from '../notifications/service.ts';
  * divergence (the 0.4 signed-checkpoint window collapses to this flag +
  * the receipt + the gdpr audit rows — rule 5).
  */
+
+/** Stands in for the subject on a record that is kept but de-identified. A
+ *  review decision is the audit trail of a governance gate, so the row stays
+ *  and the identity goes. Same value 0.4 used, so old and new rows read
+ *  alike. */
+const ERASED_SUBJECT = 'erased-user';
 
 const HOUR_MS = 60 * 60 * 1000;
 const DAY_MS = 24 * HOUR_MS;
@@ -551,6 +559,37 @@ async function scrubSubjectAuditLogs(
   return scrubbed.length;
 }
 
+/**
+ * Does the subject still belong to another organization that has not
+ * disabled them?
+ *
+ * `login_attempts`, `login_block_counters` and the two-factor tables are
+ * keyed globally, by email or by user id, while a GDPR request is scoped to
+ * one organization. Wiping them for one org's request would reset the
+ * lockout and 2FA backoff counters protecting every OTHER org the subject
+ * belongs to, which hands a multi-org user a cross-tenant bypass. 0.4
+ * refused the wipe in that case; so does this.
+ *
+ * 0.4 paged Better Auth's adapter at 256 memberships and failed CLOSED when
+ * the page came back full, because it could not see past the cap. SQL
+ * answers exactly, so neither the cap nor its guard is ported.
+ */
+async function subjectBelongsToOtherActiveOrg(
+  sql: Sql,
+  userId: string,
+  excludeOrgId: string,
+): Promise<boolean> {
+  const rows = await sql<{ elsewhere: boolean }[]>`
+    SELECT EXISTS (
+      SELECT 1 FROM "member"
+      WHERE "userId" = ${userId}
+        AND "organizationId" <> ${excludeOrgId}
+        AND "role" <> 'disabled'
+    ) AS "elsewhere"
+  `;
+  return rows[0]?.elsewhere ?? false;
+}
+
 /** The cascade — each pass bounded and idempotent; per-pass counts land on
  * the receipt. Reuses the retention purge primitives for threads and the
  * document trash for owned documents. */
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Extended, and rescoped around #3135 to stop us colliding.

**Dropped from this PR:** the `uploads` pass. #3135 rewrites the same pass and does the blob half better than I had it — strict `s3DeleteObject` so a failed delete fails the pass and lands the receipt `partial`, where I was using the retention sweeps' best-effort helper that swallows the failure. That argument is right: a receipt saying `done` is a claim about the bytes. I also reverted my `purgeBlobArtefacts` refactor of `retention/service.ts`, so that file is untouched here now.

**Added,** after an audit of all 19 of 0.4's `eraseSubject*` arms against the 10 passes on main. Five more categories survived a `done` receipt:

| Table | What survived |
| --- | --- |
| `app.notifications` | the org-level bells ABOUT the subject — `subject_user_id` was added for this pass and never read |
| `app.automation_runs` | runs they started; `input`, `output`, `trace`, `effects` hold every node's resolved values |
| `app.approvals` + `app.tasks.reviewer_user_id` | reviewer identity, and a live routing pointer at an erased user |
| `app.login_attempts` + `app.login_block_counters` | the lockout trail, keyed by email |
| `app.two_factor_attempts` + `app.two_factor_grace` | 2FA backoff, keyed by user id |

The last two are global rather than org-scoped, so they are refused while the subject is still an active member of another organization — those counters protect those organizations too, and wiping them on one org's request would hand a multi-org user a lockout bypass. 0.4 had that gate; it paged Better Auth at a 256-membership cap and failed closed on truncation. SQL answers exactly, so neither the cap nor the guard is ported.

Review decisions are de-identified rather than deleted, to the same `erased-user` sentinel 0.4 used. The decision is the audit record of a governance gate, so the row stays; `reviewer_user_id` is live routing, so it is cleared.

Video-link blobs now delete strictly, matching #3135's argument rather than the retention idiom.

## Tests

`integration-check.ts` seeds all five and counts four of them into `leftovers`, which the existing assertion requires to be `0`. The review decision gets its own assertion, because de-identified-and-kept is not the same claim as deleted.

Observed against a throwaway Postgres, driving `processErasure` directly:

| Cascade | Result |
| --- | --- |
| this branch | 8/8 assertions pass; counts `orgNotifications:1 automationRuns:2 reviewDecisions:1 authState:3` |
| `origin/main` | 7 of 8 red |

The eighth is an invariant, not a regression probe: it asserts the approvals ROW is still present, which holds on both sides.

## Scope

Four findings from the same audit are deliberately not here, because they are different concerns and two of them are #3135's:

- The legal hold is checked once before the cascade and never again, so a hold placed mid-cascade is ignored for every remaining pass. 0.4 re-read holds inside all 19 arms and its comment names the reason as FRCP 37(e) spoliation. Separate PR.
- The receipt cannot distinguish "found nothing" from "never looked", and the drawer's Full breakdown renders blank on 0.5 because `counts` is now `Record<string, number>` while the renderer skips non-objects. Separate PR.
- `app.conversations` retention never sweeps, so an admin-set `externalConversations` window is inert. Separate PR.
- `app.two_factor_attempts` also lost its 30-day age sweep, and `login_block_counters` quietly went from 30 to 90 days. Separate PR.

`policyAcknowledgements` and `wfExecutions` are the two 0.4 arms with no 0.5 table, so their absence is not a regression. The docs still promise both, which is #3135's cross-class note.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check` and `lint:sast` green.

Israeltheminer:
Re-applied onto current `main` rather than rebased, because two erasure PRs landed in the meantime and one of them changed the primitive this was built on.

**What #3135 and #3140 already fixed, so it is no longer here:** the `uploads` pass and its blob leak. #3140 rewrote it with a strict `s3DeleteObject` — a failed delete throws, the receipt lands `partial`, the rows stay for Retry — plus a refcounted release seam and `PurgeIncompleteError`. That is better than the best-effort helper I had, so my version of that pass and my `retention/service.ts` refactor are both gone. That file is untouched now.

**What is still needed, verified pass-by-pass against `main`:** all seven. None of `orgNotifications`, `automationRuns`, `reviewDecisions`, `authState`, `cloudGrants`, `syncConfigs` or `videoLinks` exists on `main` today.

The `videoLinks` blob purge now follows #3140's idiom in the same file rather than my own — lazy store resolution inside the s3 branch, a ref-dedupe set, and a throw that fails the pass instead of swallowing the failure.

## Tests

Extended `integration-check.ts`'s existing erasure check: seeds all seven, counts six into `leftovers` (which the existing assertion requires to be `0`), and gives the review decision its own assertion, because de-identified-and-kept is a different claim from deleted.

Run against a real Postgres and MinIO: **369/375**, with the erasure check passing —

```
receipt=done, thread=0, leftovers=0, reviewDeidentified=1 (want 1), scrubbed=1
```

The six failures are the warm-MinIO bucket collision, two `yt-dlp`-dependent probes and three agent-lane probes — the same set a baseline run of unmodified `main` produces on this box.

**The test caught two of my own fixture bugs before it caught anything else,** which is worth recording. First run: `leftovers=6, reviewDeidentified=0`. I had seeded the seven categories *after* the cascade had already run, so the rows were never there to erase — it read as a code defect. Second: with the seed moved ahead of the filing, it passes. The lesson is the same one as #3171's fixture: a failing assertion is not evidence of a failing fix.

## Still not here

Four findings from the same audit, all separate concerns and tracked in #3142: the legal hold is checked once rather than per pass, the receipt cannot distinguish "found nothing" from "never looked", the `externalConversations` retention window is inert, and `two_factor_attempts` lost its age sweep. The first two are #3144 (stacked on this branch, so it needs a rebase after this lands); the second two are #3143.

PR #3141, opened 2026-09-02
fix(platform): make the RAG retrieval gate fail closed

### LINKED ISSUE #3121: Bug: emailed attachments are not indexed or retrievable on the Postgres backend
An email attachment is never indexed on 0.5, so chat cannot find anything inside one. The four PRs that built this (#3011, #3012, #3014, #3062) are merged and unreachable — they live in `convex/`, which the 0.5 backend only reaches through the shim.

## Why

Three separate cuts, each deliberate and each recorded in the code:

The bind stops short of indexing. `backend/domains/conversations/shim.ts:546` says it: "the 0.4 conversation-scope RAG upgrade rides the conversation-corpus retrieval leg (still honest-empty in 0.5) — bound, never queued here." So #3010's binding works and #3012's indexing does not.

The retrieval gate has no conversation branch. `backend/domains/knowledge/service.ts` decides a ref on `documentId` or `threadId` only. Its SELECT does not read `fm.conversation_id`, even though `0037_conversation_attachments.sql` adds the column. A conversation-scoped row would fall through to the same-org fallback, which is the posture #3011 was written to prevent.

Message refs are refused outright. `service.ts:107`: `if (ref.startsWith('msg:')) { continue; }` — the whole of #3062.

`backend/MIGRATION.md` lists "conversation/email retrievable branches" as PENDING on the knowledge row, so this is deferred work rather than a porting slip.

## What still works

`bindFileToConversation` binds, so #3010 is live. The Documents-side listings from #3033 and #3035 read `file_metadata` directly and are unaffected. Conversation search (#3006, #3084) searches subjects, contacts and message bodies in SQL, not the corpus, and is unaffected.

The gap is retrieval of attachment CONTENT: a CV, a signed contract, a spreadsheet someone emailed in.

## Scope

Three pieces, in the order #3011 and #3012 shipped them:

1. Carry `conversation_id` through the corpus write and the SQL scope clause, and add the conversation branch to `filterRetrievableRagFileIds`, decided by `conversationAssignmentAllows` — the predicate already ported to `backend/domains/conversations`.
2. Queue on bind, in the bind's own transaction. #2995 is the record of what splitting that promise across a boundary costs.
3. Restore the `msg:` branch once something writes a message ref.

#3015 step 3 (index email bodies) depends on all three and should not start before them.

Not a security exposure. Nothing indexes these files, so no corpus rows exist to leak — the fallback is unreachable today. It becomes reachable the moment step 2 lands, so step 1 ships first.

Israeltheminer:
The re-check that decides which corpus refs a chat turn may read was serving refs it should refuse. A file row bound to nothing was allowed to any member of the organization.

## Why

Retrieval has two gates. The SQL half admits rows and fails open on purpose — `backend/core/knowledge/corpus.ts` says so and names `filterRetrievableRagFileIds` as the decider. That function had drifted to allow by default, under a comment claiming it was 0.4's posture:

```ts
// Legacy/unbound: same-org fallback (the 0.4 posture).
retrievable.push(ref);
```

0.4's line for the same case is `continue` — deny.

The shape is reachable. `indexUploadedFile` stamps corpus scope only from a bound document, so a row with no document carries no project and no team, and `corpus.ts` reads that as an org-wide hub row. Both gates then passed it.

Four guards 0.4 had were missing. Measured against a real Postgres, seven refs where only the first should pass:

| Ref | `origin/main` | This branch |
| --- | --- | --- |
| the document's current file | allowed | allowed |
| a ref the document has since replaced | **allowed** | denied |
| `document_id` pointing at a deleted row | **allowed** | denied |
| bound to nothing | **allowed** | denied |
| bound to a conversation | **allowed** | denied |
| `rag_status = 'running'` | **allowed** | denied |
| the file row trashed | **allowed** | denied |

Six of the seven denials were being served.

## What changed

The SELECT reads four more columns: `fm.rag_status`, `fm.lifecycle_status`, `fm.conversation_id`, and both `d.id IS NOT NULL` and `d.file_ref`.

- Indexing must have finished. A row mid-reindex still has its previous chunks in the corpus, and answering from them serves content the caller's current scope was never checked against.
- A trashed file row is denied, not only a trashed document.
- A `document_id` whose row is gone no longer reads as a hub document. The LEFT JOIN's NULLs were falling into the hub branch.
- A ref the document has since replaced is denied. A replacement upload moves `file_ref` on and leaves the previous corpus row behind, so this also closes the retrieval half of the superseded-document defect.
- Bound to nothing, or bound to a conversation, is denied.

The conversation case gets its own explicit branch rather than relying on the new default, so the next reader sees where the allow branch goes. That branch is #3121's, decided by `conversationAssignmentAllows` against the conversation's current assignment.

## Risk

The `rag_status = 'completed'` requirement makes retrieval briefly deny a knowledge entry while a new version reindexes — `knowledge_entries/service.ts` sets `rag_status = NULL` and re-queues. That is the conservative direction and it is what 0.4 did. Every other `rag_status` writer converges on the one completion write in `knowledge/service.ts`.

## Tests

A new `checkRetrievalGate` in `integration-check.ts` pins the table above. It needs no object storage, so it runs on every invocation — unlike the RAG loop below it, which skips without `ITEST_S3_ENDPOINT`. Placing it inside that function would have meant it never ran where the gate matters most.

Mutation: against `origin/main`'s copy of the function, six of the seven assertions go red and the legitimate hub document still passes.

## Scope

Does not close #3121. Emailed attachments still are not indexed; this makes the gate safe for that work to land against, which is why it goes first.

The indexer-side half of the superseded defect is untouched — a replacement still leaves an orphaned corpus row, and a stale job still spends extract-and-embed cost on a blob nobody will read. This denies the read; purging the row is separate.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check` and `lint:sast` green.

### DIFF
--- a/services/platform/backend/domains/knowledge/service.ts
+++ b/services/platform/backend/domains/knowledge/service.ts
@@ -63,9 +63,21 @@ interface AccessScopeArg {
 }
 
 /**
- * Tier-A retrievable filter over app tables: a file ref passes when its
- * metadata row belongs to the org AND (bound document is un-trashed and in
- * scope | thread-bound and the thread is in scope | legacy unbound same-org).
+ * The mandatory re-check over app tables: which corpus refs may this turn
+ * actually read.
+ *
+ * This is the gate `core/knowledge/corpus.ts` delegates to by name. The SQL
+ * half only ADMITS rows — its own comment says so — so this function is the
+ * one that decides, and it must FAIL CLOSED. A ref passes only when its
+ * metadata row belongs to the org, finished indexing, is not trashed, and
+ * lands in exactly one of the two allow branches: a bound document that is
+ * live, un-trashed, still the document's current file, and in scope; or a
+ * thread-bound upload whose thread is in the turn's lineage.
+ *
+ * Everything else is denied, including a row bound to nothing. Denying the
+ * unbound case is what 0.4 did, and it matters because the corpus stamps no
+ * project and no team for such a row, which the SQL half then reads as an
+ * org-wide hub row.
  */
 async function filterRetrievableRagFileIds(
   sql: Sql,
@@ -82,7 +94,12 @@ async function filterRetrievableRagFileIds(
     {
       storageRef: string;
       threadId: string | null;
+      conversationId: string | null;
+      ragStatus: string | null;
+      fileTrashed: boolean;
       documentId: string | null;
+      docExists: boolean;
+      docFileRef: string | null;
       docTrashed: boolean | null;
       docProjectId: string | null;
       docTeamId: string | null;
@@ -90,7 +107,12 @@ async function filterRetrievableRagFileIds(
     }[]
   >`
     SELECT fm.storage_ref AS "storageRef", fm.thread_id AS "threadId",
+           fm.conversation_id AS "conversationId",
+           fm.rag_status AS "ragStatus",
+           (fm.lifecycle_status = 'trashed') AS "fileTrashed",
            fm.document_id AS "documentId",
+           (d.id IS NOT NULL) AS "docExists",
+           d.file_ref AS "docFileRef",
            (d.lifecycle_status = 'trashed') AS "docTrashed",
            d.project_id AS "docProjectId", d.team_id AS "docTeamId",
            d.team_tags AS "docTeamTags"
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Closing in favour of #3177.

#3140 rewrote `filterRetrievableRagFileIds` into a pure `decideRetrievable` in a new `retrievable.ts`, which incidentally fixed three of this branch's four findings: superseded refs (it now queries documents BY `file_ref`), trashed file rows, and the deleted-document-behind-a-live-id case. So this 130-line SQL change is obsolete.

The one that survived is the important one — an unbound row is still admitted to the whole organization — and against the new structure it is a two-line change in a pure function that already has a test harness. #3177 does that.

The `rag_status` guard is deliberately dropped there, with the reasoning stated: the superseded case that motivated it is already handled, and the remaining window is stale own-content rather than a leak.

PR #3144, opened 2026-09-02
fix(platform): re-check legal holds per erasure pass and fix the receipt

Israeltheminer:
Two ways an erasure receipt could claim more than happened, and the panel meant to show the detail rendered blank.

Stacked on #3122 — review that first. Refs #3142.

## Why

**The legal hold was read once and never again.** `processErasure` gates on it before the cascade, then runs ten passes that are not one transaction. Two of them fan out per-thread lineage purges and per-document blob and corpus deletes, so the window is real. A hold placed during that window was ignored for every remaining pass.

0.4 re-read holds inside all 19 of its arms, and the helper's docstring names why: "a hold placed AFTER scheduling and BEFORE the per-table mutation runs must still win… Pre-fix, only `orgHeld` was checked, leaving a window where a custodian-hold race could let GDPR erasure delete data the hold was meant to preserve — FRCP 37(e) spoliation."

**A hold that stopped the cascade halfway still reported `done`.** Only a thrown pass set `partial`. Under Art 19 that receipt is the subject's confirmation, so `done` has to mean every category was reached.

**The Full breakdown panel rendered blank for every recent request.** A pass records a plain count; the renderer did `if (typeof value !== 'object' || value === null) continue;`, so every entry was dropped *before* the empty-category branch. The result was an empty list with not even the "no data in N other categories" line. The headline counts still rendered, so it looked populated.

## What changed

The hold is re-read before each pass. An unreadable hold table skips the pass rather than running it — a table you cannot read is not evidence that nothing is held.

A held-off pass is tracked separately from a failed one. Both make the receipt `partial`, and the error line names which, so an operator can tell a hold from a fault.

The breakdown fold reads both shapes. Receipts written before the backend change still carry the object form, so dropping it would have broken the panel in the other direction.

## What is not covered

The per-pass re-check itself has no automated test. Proving it needs a hold to appear *between* two passes, and every way I could arrange that was timing-dependent — a flaky test on a spoliation guard is worse than none. What is pinned instead is the rule the re-check feeds: that a held-off pass makes the receipt `partial` and says so. If someone reverts the re-check, that test stays green. Naming the gap rather than implying coverage.

## Tests

`erasureReceiptStatus` and `erasureReceiptError` are extracted from the cascade so the "done means every category was reached" rule is testable at all, with 6 assertions. `foldBreakdownEntries` is extracted from the drawer for the same reason, with 6 more.

| Mutation | Went red |
| --- | --- |
| status ignores held-off passes | the hold case |
| the hold reason dropped from the error line | the combined-reason case |
| the fold's plain-count branch removed | the count-shape case |
| the fold treats a held-off category as empty | two cases |

The fold mutation is the original bug, so seeing that assertion red is seeing the defect reproduced.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check`, `lint:sast` green; 12 unit assertions passing.

### DIFF
--- a/services/platform/app/features/settings/governance/data-subject-requests/breakdown-entries.test.ts
+++ b/services/platform/app/features/settings/governance/data-subject-requests/breakdown-entries.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, test } from 'vitest';
+
+import { foldBreakdownEntries } from './breakdown-entries.ts';
+
+describe('foldBreakdownEntries', () => {
+  test('reads the plain-count shape a pass records', () => {
+    // The whole panel rendered blank on this shape before: every entry was
+    // skipped, so neither a row nor the empty-category line appeared.
+    expect(
+      foldBreakdownEntries({ uploads: 3, cloudGrants: 2, memories: 0 }),
+    ).toEqual({
+      visible: [
+        { key: 'uploads', rows: 3, skippedByHold: 0 },
+        { key: 'cloudGrants', rows: 2, skippedByHold: 0 },
+      ],
+      zeroCount: 1,
+    });
+  });
+
+  test('still reads the older object shape on existing receipts', () => {
+    expect(
+      foldBreakdownEntries({
+        documents: { rows: 4, skippedByHold: 0 },
+        threads: { rows: 0, skippedByHold: 2 },
+        feedback: { rows: 0, skippedByHold: 0 },
+      }),
+    ).toEqual({
+      visible: [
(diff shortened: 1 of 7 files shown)


Israeltheminer:
Closing in favour of the replacement above: this branch was stacked on #3122's pre-merge branch, and #3122 was re-applied and then merged, so the stack's base no longer exists. The replacement is the same change applied directly onto `main`, with both mutations re-verified on the merged base.

PR #3145, opened 2026-09-02
fix(platform): space the daily governance crons

Israeltheminer:
The three daily governance jobs were running in one 30-minute window. They are spaced again, and a test now pins the rule rather than the times.

Refs #3142.

## Why

| Job | 0.4 | Before this | Now |
| --- | --- | --- | --- |
| legal-hold releases | 01:00 | 04:15 | 01:00 |
| audit-chain verify | 02:00 | 04:30 | 02:00 |
| retention sweep | 04:00 | 04:00 | 04:00 |

Both 0.4 comments state the reasoning. The releases cron: "Picks an off-peak hour so it doesn't compete with the main 04:00 sweep." The verify cron: "02:00 avoids the 01:00 legal-hold release sweep and the 04:00 retention sweep."

Two consequences of the packing:

A maker-checker legal-hold release that clears its cooldown used to free data three hours before the sweep. At 04:15 it lands after the sweep has started, so the data it freed waits roughly a day for the next one.

The audit verify ran 30 minutes into the retention window, so it could walk the chain while retention deletes audit prefixes. That is the overlap the 0.4 comment named, and it compounds with the anchor change `backend/MIGRATION.md` documents: the verifier trusts the oldest surviving row, and it may now read that row mid-deletion.

## What changed

Three cron strings, reordered in the file so the sequence reads in the order it runs, with the reasoning written down rather than left to be rediscovered.

## Tests

A new `schedules.test.ts` pins the invariant, not the clock: the three do not share an hour, releases precede the sweep, and the verify is not in the sweep's hour. Plus two roster hygiene checks — unique names, five cron fields each. `SCHEDULES` is exported for this; it was module-private.

Mutation: setting the two times back to `15 4` and `30 4`, which is exactly what main ships, turns three of the five assertions red.

## Scope

Only the three daily governance jobs move. Every other entry is untouched, including the deliberate staggers already in the file — the two Drive scans on offset 15-minute cadences, and the three 5-minute file-pipeline sweeps on offset minutes.

Does not address the anchor change itself, which is signed off in the ledger as an intentional divergence. This removes the overlap that made it sharper.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check` green; 5 unit assertions passing.

### DIFF
--- a/services/platform/backend/jobs/schedules.test.ts
+++ b/services/platform/backend/jobs/schedules.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, test } from 'vitest';
+
+import { SCHEDULES } from './schedules.ts';
+
+/** `m h dom mon dow` — these three are all daily, so hour and minute suffice. */
+function timeOf(name: string): { hour: number; minute: number } {
+  const entry = SCHEDULES.find((row) => row.name === name);
+  if (entry === undefined) throw new Error(`no schedule named ${name}`);
+  const [minute, hour] = entry.cron.split(' ');
+  return { hour: Number(hour), minute: Number(minute) };
+}
+
+describe('the daily governance schedules stay spaced', () => {
+  const releases = timeOf('governance.effect_hold_releases');
+  const verify = timeOf('audit.integrity_check');
+  const sweep = timeOf('governance.retention_cleanup');
+
+  test('releases run before the sweep that frees their data', () => {
+    // A cleared maker-checker release frees data for the same night's sweep
+    // instead of waiting another day.
+    expect(releases.hour).toBeLessThan(sweep.hour);
+  });
+
+  test('the audit verify does not run inside the retention window', () => {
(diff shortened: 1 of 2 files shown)


Israeltheminer:
Rebased onto current `main`, which had added a fourth daily job in the same window: `knowledge.reconcile_corpus` at `45 4`.

That one is kept where it is, because its own comment gives the reason — it runs after retention so rows the sweep just purged reconcile the same night. So it deliberately DOES share the sweep's hour, unlike the three this PR separates.

The four now read 01:00 releases → 02:00 verify → 04:00 sweep → 04:45 reconcile.

The test gained a case pinning that, so the spacing rule is not read as "spread everything out" and the reconcile moved off the sweep's hour later. Two mutations, both red:

| Mutation | Went red |
| --- | --- |
| the three packed back into 04:00–04:30 (what `main` ships) | 3 of 6 |
| reconcile moved to 03:45, before the sweep | 1 of 6 |

Issue #1505, opened 2026-04-13
Feature: Harden audit log (append-only guarantee, scheduled integrity check, retention)


larryro:
## Context

Compliance requires that logs are protected against tampering and unauthorized access, that admins cannot delete or falsify their own activity logs, and usually a defined retention.

We already have a strong base:
- Hash-chained audit log in [services/platform/convex/lib/helpers/audit_hash.ts](services/platform/convex/lib/helpers/audit_hash.ts)
- Admin-only chain verification in [services/platform/convex/audit_logs/verify_integrity.ts](services/platform/convex/audit_logs/verify_integrity.ts)
- Export in [services/platform/convex/audit_logs/export_audit_logs.ts](services/platform/convex/audit_logs/export_audit_logs.ts)

Gaps:
1. No explicit guarantee that there is no mutation/delete endpoint on the `auditLogs` table. Hash chain **detects** tampering but doesn't **prevent** row deletion.
2. No scheduled chain verification — breaks would only be seen if an admin manually runs the verify query.
3. No retention policy.

## Proposed change

1. Audit existing codebase — confirm no `patch`/`delete`/`replace` on `auditLogs`. Remove any that exist. Add a unit test asserting the table is append-only.
2. Add a scheduled Convex cron that runs `verify_integrity` nightly and emits a `security` audit entry + email to admins on failure.
3. Add configurable retention (default: keep forever, optional archive-then-delete after N days). Archived exports must also be signed/hashed.
4. Review read-side access: only org admins should be able to read audit logs. Confirm RLS in [services/platform/convex/lib/rls/](services/platform/convex/lib/rls/) enforces this.

## Acceptance criteria

- [ ] No code path allows deletion or mutation of `auditLogs` rows.
- [ ] Nightly integrity check runs automatically.
- [ ] Retention policy configurable per org.
- [ ] Docs updated.

Part of the security compliance baseline.

yannickmonney:
Triaged — keeping open but re-scoping, since the base is stronger than the issue assumes. Already implemented: the audit-log **hash chain** (`audit_logs/helpers.ts`), an admin **integrity-verify** action (`verify_integrity.ts`), **export**, and **retention** (daily cron hard-deletes beyond the window, with signed checkpoints + re-anchoring). 

Remaining, genuinely-open work:
[...]

PR #3149, opened 2026-09-02
fix(platform): restore the audit-log, conversation and 2FA gates

Israeltheminer:
Three authorization and audit behaviours the port dropped. Any member could read the audit log; a read-only member could send outbound mail.

Refs #3142.

## Audit-log reads were open to every role

0.4 restricted them to admin and owner (#1505). The log enumerates who did what to whom across the organization, and it carries the GDPR erasure trail — data subject names and the lawful grounds for erasing them.

Fixed in the ability matrix rather than route by route, so all six doors are covered together: list, detail, summary, export, integrity verify and status. Gating them individually invites the next door to ship ungated.

Developers keep `write`, so a developer's own actions still append their audit rows. Members hold neither.

## A read-only member could write to conversations

Reply, send outbound mail, close, bulk-act, delete. 0.4 ran every conversation mutation through `mutationWithRLS`, whose `conversations` and `conversationMessages` rules put each write through `authorizeRls` at editor-or-above.

`viewerCanWrite` reuses that same matrix rather than declaring a second role list. It composes with the half that was correctly ported: a write door checks the role first, then loads the row through `loadVisibleConversation`, which applies `conversationAssignmentAllows` per row. Role decides *whether you may change a conversation*; assignment decides *which ones you can see at all*; assignment itself is stricter still.

Sending outbound mail is the sharpest of the five — it leaves the building under the organization's name.

## 2FA and passkey events audited only on failure

An attacker who added a passkey and disabled 2FA left no successful-action trail. All five events now audit, inside the same serializable transaction as the state change.

The 0.4 action names are reused verbatim — `2fa_enrolled`, `2fa_disabled`, `passkey_added`, `passkey_removed`, `passkey_sign_in`. A renamed action silently breaks any existing query or export that groups on it. Checked against 0.4's full vocabulary: `passkey_removed` and `passkey_revoked_by_admin` were distinct events there and stay distinct here.

## Tests

`access.test.ts` pins the matrix, and `integration-check.ts` covers the doors end to end.

| Mutation | Went red |
| --- | --- |
| `auditLogs: ['read', 'write']` restored to the developer role | two assertions |

Each finding asserts both directions — the lower role is refused **and** the higher role still allowed. Testing only the refusal is half a test: a gate that refuses everyone passes it.

## The caller audit

Tightening a gate breaks callers, so every door was traced before commit. The audit-log doors are reached only from the admin-facing settings surface. The conversation write doors are reached from the inbox UI, the REST surface (which carries its own editor-role gate), and the connector lanes, which run under system auth. The 2FA changes add audit writes and gate nothing, so they have no caller risk.

Role comes from the session-resolved value where the seam offers it, not a raw `member` table read — trusted-headers deployments insert a placeholder member row, so a proxy-declared admin can hold a non-admin role in that table.

## Scope

Retention destruction still emits no audit rows — sixteen 0.4 action names exist nowhere in the tree. Same class, different file, and #3143 is already editing it.

The audit chain's `pii_scrubbed` recompute skip and its oldest-row trust anchor are untouched. Both are signed off in `backend/MIGRATION.md` as intentional divergences, so they are a design question rather than a porting defect.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check`, `lint:sast` green; the new unit suite 7/7.

### DIFF
--- a/services/platform/backend/auth/access.test.ts
+++ b/services/platform/backend/auth/access.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from 'vitest';
+
+import { authorizeRls } from './access.ts';
+import { isAdminRole } from './membership.ts';
+
+/**
+ * The Better Auth role matrix must agree with the ported RLS matrix in
+ * `core/lib/rls/helpers/access_control.ts` (which has its own suite) on the
+ * rows where the two overlap. `auditLogs` is the row that drifted: the port
+ * derived every role from `uniformGrants`, which handed `member`, `editor`
+ * and `developer` a read they must not have.
+ */
+describe('authorizeRls — auditLogs (#1505 read-side hardening)', () => {
+  it('grants read only to admin and owner', () => {
+    expect(authorizeRls('owner', 'auditLogs', 'read')).toBe(true);
+    expect(authorizeRls('admin', 'auditLogs', 'read')).toBe(true);
+    expect(authorizeRls('developer', 'auditLogs', 'read')).toBe(false);
+    expect(authorizeRls('editor', 'auditLogs', 'read')).toBe(false);
+    expect(authorizeRls('member', 'auditLogs', 'read')).toBe(false);
+    expect(authorizeRls('disabled', 'auditLogs', 'read')).toBe(false);
+  });
+
+  it('keeps write for every role whose own actions are audited', () => {
+    expect(authorizeRls('owner', 'auditLogs', 'write')).toBe(true);
+    expect(authorizeRls('admin', 'auditLogs', 'write')).toBe(true);
+    expect(authorizeRls('developer', 'auditLogs', 'write')).toBe(true);
+    expect(authorizeRls('editor', 'auditLogs', 'write')).toBe(true);
+    expect(authorizeRls('member', 'auditLogs', 'write')).toBe(false);
+    expect(authorizeRls('disabled', 'auditLogs', 'write')).toBe(false);
+  });
+
+  it('agrees with the isAdminRole gate the audit-log doors use', () => {
+    for (const role of ['owner', 'admin', 'developer', 'editor', 'member']) {
+      expect(authorizeRls(role, 'auditLogs', 'read')).toBe(isAdminRole(role));
+    }
+  });
+});
+
+/**
+ * The conversation WRITE gate reads the same matrix. 0.4 gave
+ * `conversations` / `conversationMessages` ALL to editor-and-above and
+ * READ_ONLY to `member`, so a read-only member can see a thread assigned to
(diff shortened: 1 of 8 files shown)


Israeltheminer:
Split. The verified half is #3181; this branch's remaining finding needs real work, so I am not shipping it behind a passing fix.

**Dropped: the audit-log third.** `main` fixed it independently and more thoroughly than this branch did — the ability matrix pins `auditLogs` per role, `routes.ts` consults it via `authorizeRls`, and `checkAuditSurface` asserts the refusal end-to-end. My route-level `isAdminRole` check was a divergent second expression of the same rule.

Worth recording: my first attempt at removing that third **deleted `main`'s own `checkAuditSurface` function** — `main` has one, and my commit had appended a block inside it, so cutting "my" block took `main`'s with it. Caught by diffing function names against `main`; restored byte-for-byte and re-applied my checks with a 3-way patch instead of a hand splice. That is the second time today a splice ate `main`'s content, the other being `MIGRATION.md` on #3146.

**Shipped: the 2FA lifecycle audit** → #3181. Verified 374/380 with the full five-name trail.

**Held: the conversation write-role gate.** It is faithful — 0.4's rule is literally `authorizeRls(role, 'conversations', 'write')` and this repo's matrix gives `member` read only — but it does not survive the harness:

| | result |
| --- | --- |
| both halves | aborts at 178 checks, 9 failures |
| `main` | 373/379, 6 known failures |
| 2FA half alone | 373/379, identical 6 |

So the gate owns the damage, and the first observable failure is a genuine interaction rather than a fixture accident:

```
FAIL  blob-ref authority: outbound mail attaches only the sender's uploads
      compose → 403 FORBIDDEN (want 403 attachment_not_owned)
```

The gate fires before the attachment-ownership check, so that assertion receives the wrong 403, and several chat checks cascade behind it. Two possible resolutions, and picking between them is a decision rather than a fix: raise the roles of the harness actors that compose mail (if the fixtures were written assuming members may write), or move the gate relative to the ownership check so the more specific refusal still wins.

Leaving this open as the record of that decision. The finding itself stands — a read-only member can currently reply, send outbound mail, close, bulk-act and delete conversations — and it is in #3142.

Israeltheminer:
Superseded by #3187, which rebases this onto current main.

Two changes beyond a rebase:

- The audit-log third is dropped. #3159 landed the same role matrix, and main's doors already call `authorizeRls(role, 'auditLogs', 'read')` — which that matrix now denies below admin. Swapping them to `isAdminRole` would restate the rule in a second place.
- The attachment-fetch door is gated too, so all twelve write-shaped doors carry the check, plus a static test that fails when a new door ships without it.

PR #3150, opened 2026-09-02
docs: describe the Postgres backend instead of Convex

Israeltheminer:
The documentation still told operators to run a `tale-convex` container, store blobs in Convex storage, and look for code under `services/platform/convex/`. None of that exists. 63 pages across all three locales.

Refs #3142.

## Why

#3107 replaced the Convex backend with Node, Hono and Postgres; #3125 deleted the tree. The docs were not swept.

The self-hosted pages matter most because they are procedures someone follows during an incident.

**Backups and restore was the dangerous one.** It listed `rag-data` and `crawler-data` as snapshot volumes — neither exists — and never said that uploaded file blobs sit **outside** a snapshot. An operator following it would restore an instance with no files and no warning that anything was missing.

Verified against `SNAPSHOT_VOLUMES` in `tools/cli/src/lib/backup/constants.ts`: `db-data`, `convex-data`, `caddy-data`, `caddy-config`. `object-store-data` is deliberately absent. The page now says so, before the procedure rather than after it.

**The container overview was wrong twice.** It said eight containers; a first pass said ten; `compose.yml` declares eleven. The one nobody had documented is `bgutil-provider` — and it is the only image in the stack Tale does not build, which is exactly the kind of thing an operator reviewing their supply chain wants named. All three locales now say eleven and describe it.

`convex-data` keeps its name, and the page now explains why: the volume predates the retirement, and renaming it would have forced every operator to migrate a volume for cosmetics. The code comment says the same thing.

Also corrected: the status page's liveness probe (it described probing `${CONVEX_URL}/version`, a service that no longer exists), the WebDAV page's blob storage and lock table, the API reference's content endpoint, and the data-residency request path.

## What is not here

`services/docs/app/content/frontmatter.json` is a generated manifest whose key order is not deterministic — regenerating it churns 2,196 lines with zero net change, and a structural test requires it to match the pages. So its diff is hand-limited to the **6 entries that actually changed**, keeping the committed key order, rather than 4,392 lines of noise that would conflict with every other docs branch.

That non-determinism is a separate repo defect and is not fixed here.

## Method

Every replacement claim was read out of the code, the compose file, or `backend/MIGRATION.md` before being written. No env var, container name, path or command is invented. English follows the repo's docs contract; German and French are written natively per the translation contract rather than rendered word-for-word from the English.

Gate: `oxfmt --check` clean, docs lint exit 0, docs structural suite **194/194**.


Israeltheminer:
Closing — #3154 has since aligned the self-hosted docs with the Postgres backend, and this branch is now behind main on two facts:

- It describes eleven containers with a separate `tale-knowledge-db`. That is the repo's local-dev `compose.yml`; the `tale deploy` stack is CLI-generated, has no separate knowledge service, and reaches the corpus through a `knowledge-db` network alias on `db` — which is what main already says.
- It describes the platform as React + TanStack Start. Main correctly says Vite + TanStack Router.

Applying it would regress both.

#3188 carries the part nobody had done: eleven comments in the shipped `compose.yml` that still describe the Convex runtime, plus two references to deleted paths.

Issue #1508, opened 2026-04-13
Feature: WebAuthn / Passkeys as second factor


larryro:
## Context

Follow-up to the TOTP 2FA work. WebAuthn / passkeys give us phishing-resistant MFA and satisfy the compliance wording that explicitly mentions client certificates, smart cards, and ID tokens.

Depends on the 2FA plugin being enabled first.

## Proposed change

1. Enable WebAuthn via Better Auth's plugin (or `@simplewebauthn/server`).
2. Allow users to register multiple authenticators (platform and roaming).
3. Users can prefer passkey over TOTP at sign-in time.
4. Admin UI to list/revoke a user's registered authenticators.

## Acceptance criteria

- [ ] Users can register a passkey on a supported browser/device.
- [ ] Passkey satisfies the 2FA check during sign-in.
- [ ] Revocation flow works.

Part of the security compliance baseline.

yannickmonney:
Triaged — valid feature, keeping open. The TOTP 2FA prerequisite is already in place (Better Auth two-factor hook with per-user lockout, `2fa-enroll` route), so this would add WebAuthn/passkeys as an additional phishing-resistant factor on top — a natural fit for the Better Auth passkey plugin. Assigned to me.

yannickmonney:
Implementation outline (builds on the existing Better Auth + 2FA setup from #1507). The one upfront decision is the **dependency policy**, so flagging it first.

**Decide first — dependencies vs. Renovate.** Passkeys need the Better Auth `passkey` plugin + `@simplewebauthn/server` and `@simplewebauthn/browser`, none of which are installed. They'll trip Renovate's `minimumReleaseAge` (90d). Need a `packageRules` exception/pin before adding them.

**Then:**
1. **Schema.** Run the `generate-auth-schema` skill to add the `passkey` credential table to `convex/betterAuth/schema.ts`, and regenerate `_generated` (needs a running Convex backend).
2. **Server.** Add the passkey plugin to `convex/auth.ts` `plugins:[]`, deriving RP ID + origin from `siteUrl` (reuse the existing HTTPS guard).
3. **Client.** Add the passkey client plugin to `lib/auth-client.ts`.
4. **Enrollment UI.** Model on `app/routes/2fa-enroll.tsx` (register / name / list / remove credentials).
5. **2FA policy.** Make `convex/two_factor/auth_hooks.ts` count an enrolled passkey as satisfying `two_factor_policy` (so passkey is a valid second factor / passwordless step-up).
[...]

PR #3181, opened 2026-09-03
fix(platform): audit successful 2FA and passkey lifecycle events

Israeltheminer:
An attacker who registered their own passkey and turned 2FA off left no successful-action trail. 0.4 audited every successful second-factor lifecycle event (#1508); the port kept only the failure half.

Split out of #3149. Refs #3142.

## Why

The failure half audits, so a wrong code is recorded — but enabling 2FA, disabling it, and adding, removing or signing in with a passkey were silent. Those are the events that matter most in an account takeover.

All five 0.4 action names are reused verbatim: `2fa_enrolled`, `2fa_disabled`, `passkey_added`, `passkey_removed`, `passkey_sign_in`. A renamed action silently breaks any query or export that groups on it. Checked against 0.4's full vocabulary — `passkey_removed` and `passkey_revoked_by_admin` were distinct events there and stay distinct here.

Rows are written in the same transaction as the state change, carrying the 0.4 `security`/`success` shape.

## What is NOT here, and why

#3149 also restored the conversation write-role gate, so a read-only member could no longer reply, send outbound mail, close, bulk-act or delete. **That half is held back.** It is faithful — 0.4's rule is literally `authorizeRls(role, 'conversations', 'write')`, and this repo's own matrix gives `member` read only — but it breaks the harness:

```
mine (both halves):  aborts at 178 checks, 9 failures
main:                373/379, 6 known failures
2FA half alone:      373/379, identical 6
```

The first observable failure is real, not incidental:

```
FAIL  blob-ref authority: outbound mail attaches only the sender's uploads
      compose → 403 FORBIDDEN (want 403 attachment_not_owned)
```

The gate fires before the attachment-ownership check, so that assertion gets the wrong 403 — and several later chat checks then cascade. Either the harness fixtures compose as under-privileged actors and need their roles raised, or the gate's placement relative to the ownership check needs rethinking. That is real work, not a rebase, so it is not being smuggled in behind a verified fix.

## Tests

`integration-check.ts` asserts the whole trail and its shape, on a **dedicated throwaway user** rather than the harness's own:

```
trail=2fa_enrolled,2fa_disabled,passkey_added,passkey_sign_in,passkey_removed
shape=security/success   enabledLeftOn=false
```

That isolation is load-bearing, and finding out cost three runs. Running the lifecycle on the shared session left that user's second factor changed and their session invalid, which failed five unrelated checks — the 2FA grace check plus four chat checks. The throwaway user also needs an org membership, because audit rows are org-scoped and without one the trail reads empty.

WebAuthn cannot be driven from a script, so the passkey trio is asserted at the writer: the action names land as rows with the security shape.

**374/380**, and the six failures are byte-identical to a baseline run of unmodified `main` on the same instance — the warm-MinIO bucket collision, two `yt-dlp` probes and three agent-lane probes.

Gate: `typecheck`, `oxlint --type-aware`, `oxfmt --check`, `lint:sast` green; branched from `origin/main` at `899fcc08a`.

### DIFF
--- a/services/platform/backend/auth/auth.ts
+++ b/services/platform/backend/auth/auth.ts
@@ -29,8 +29,10 @@ import {
   evaluateTwoFactorEnforcement,
   getTwoFactorLockState,
   recordTwoFactorFailure,
+  recordTwoFactorLifecycleEvent,
   recordTwoFactorSuccess,
   setGraceUntilIfAbsent,
+  type TwoFactorLifecycleAction,
 } from '../domains/two_factor/service.ts';
 import { addJobInTx } from '../jobs/enqueue.ts';
 import { readGovernancePolicy } from '../lib/org-config.ts';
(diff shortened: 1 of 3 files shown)


Israeltheminer:
Superseded by #3196, which carries what is left of this.

The five audit-event calls in `auth.ts` and `two_factor/service.ts` are now byte-identical to main — #3187 landed them when I folded #3149's 2FA third in without checking my own open PRs first, so that half shipped twice.

#3187 also carried the lifecycle check, but the shared-session version. The isolation fix here — a dedicated user, plus the `member` row its org-scoped audit rows need — never made it across, and main currently has the version that invalidates the shared session and fails five later checks. #3196 is that fix alone.

PR #3223, opened 2026-09-05
fix(platform): harden the REST v1 door against malformed input and drift

larryro:
## Summary

Hardens the `/api/v1` REST door against malformed client input and against drift between the routes, the OpenAPI spec and the API reference. Client faults that reached the app-level 500 handler (and error reporting) now answer the documented 4xx envelope; six list routes are bounded through the shared `pageLimit`; the project-file bind runs the org upload policy the contracts promise; three routes complete paths the published contract describes; the spec is corrected to what the handlers do and gains a status/shape parity test layer; ~330 lines of retired 0.4 REST wrapper helpers are removed.

## Findings fixed

- rest-v1-1 — knowledge-entry writes charge `knowledge:mutate` through `chargeOrgRateLimit` before the try and answer the standard 429 + `Retry-After` instead of a 500.
- rest-v1-2 — every unguarded `c.req.json()` (~20 write routes across all families) reads through `readJsonBody` (shared.ts); a malformed/empty body answers the route's own 400 envelope; one `'{'` → 400 regression per family (core, tasks, automations, threads, projects).
- rest-v1-3 — GET /documents, GET /knowledge-entries, GET /threads/{id}/messages, GET /websites, GET /websites/{id}/pages (offset floored at 0), POST /websites/{id}/search all run through `pageLimit` (truncate + clamp); spec names the crawl-read caps.
- rest-v1-4 — POST /projects/{id}/files runs the org upload policy inside the bind transaction in the session lane's order — HEAD (`statOrgBlob`) → `validateDocumentUploadForOrg` with the attested size → `registerUpload` with the type the gate resolved — so the per-user volume sum never sees the row it is admitting (file:upload budget, size caps, MIME/extension allowlist, per-user quota); a refusal rolls the intent consume back; a blob that never landed answers 404 `BLOB_NOT_FOUND`.
- rest-v1-5 — spec.ts corrected to the routes (30-minute TTL, PUT-only handoff + `s3Ref`, intent refusal 409, real 400 + codes for policy refusals, real `executionId`, DELETE /agents 204/404, generation enum idle/queued/streaming, no `agentSlug` on …/messages, `projectId` semantics); POST …/folders now passes `folder:mutate` and POST …/comments `task:comment` as promised; rate-limits docs (en/de/fr) lose the contradicting sentence; `public/openapi.json` regenerated; spec.test.ts gains a status/shape parity layer; the bind description names "the organization's `file:upload` budget" (the in-app door additionally charges per user, so the two doors do not share one bucket). The bind description names "the organization's `file:upload` budget" rather than one the in-app upload "shares" (that door additionally charges per user).
- rest-v1-6 — `core/lib/rest/helpers.ts` trimmed to what the SCIM and MCP doors import; dead describes removed; stale `withRestAuth` comments fixed.
- rest-v1-7 — POST /automations/{name}/projects uses the atomic `bindProject` (INSERT … ON CONFLICT DO NOTHING) instead of a set rewrite from a stale read; repeat bind answers 200 `added:false`.
- rest-v1-8 — POST /threads declares and forwards `agentSlug` like the session door.
- rest-v1-9 — GET …/files/{documentId}/content presigns with the document title so Content-Disposition carries the filename.
- rest-v1-10 — the door and `assertExplicitOrg` map org-resolution failures through `domainErrorResponse`: 400/403/404 per the domain's code, rethrow (→ app handler + error reporting) for a driver failure.
- rest-v1-11 — POST/PATCH /websites parse `domain` through `parseWebsiteDomain` (400 `Invalid domain` for `https://`, `a b`, `::`); title ≤200 / description ≤2000.

## Skipped

- none

## Tests & gates observed

Review follow-up (commits 1f3d5edbb, 17b64eb06, 90c5332bc), observed on 90c5332bc:

- `bunx tsc --noEmit`: exit 0. `bunx oxlint --type-aware`: exit 0. Pre-commit ran on each of the three commits: oxfmt + opengrep `0 findings`.
- `bunx vitest --run --project server`: `Test Files 498 passed (498) / Tests 5952 passed (5952)`. The ordering regression `counts the file it is admitting once against the per-user volume quota` is red with the previous route file swapped in (`expected 400 to be 201`, 5 of the 9 bind tests fail) and green on the fix.
- `bunx turbo run test --filter=@tale/platform --force` (uncached): `@tale/platform:test: Test Files 504 passed (504) / Tests 73166 passed (73166)`.
- `bun run check` (worktree root): EXIT=0 — `Tasks: 40 successful, 40 total` (33 turbo cache hits on identical inputs; the platform test task was additionally run uncached above).
- `bun run knip:check`: exit 0 (only the pre-existing `cron-parser` configuration hint).
- `bun run generate:openapi` → `jq -S` identical to the committed `public/openapi.json`.
- Integration proof: NOT re-run to completion on the follow-up commits — the run was cut off by the session budget while in flight and aborted cleanly (cherry-pick dropped, containers removed, tree clean). The original-submission run below (0dc3fff1a) exercised the REST bind end-to-end; the follow-up changes the bind's in-transaction order only (HEAD → policy → row) and is covered by the unit regression above. A reviewer wanting the real-Postgres proof on the tip: `run-itest.sh` with PR #3222's `aa3d1d7a9` cherry-picked temporarily (it is still open).

Original submission:

- `bun run check` (worktree root): EXIT=0 — `Tasks: 40 successful, 40 total`; `@tale/platform:test: Test Files 504 passed (504) / Tests 73160 passed (73160)`; `@tale/platform:test:ui: Test Files 455 passed (455) / Tests 3503 passed (3503)`; every `*:lint: Found 0 warnings and 0 errors.`; `format:check: All matched files use the correct format.`
- `bun run knip:check`: exit 0 (only the pre-existing `cron-parser` configuration hint).
- `bun run generate:openapi` → `jq -S` identical to the committed `public/openapi.json` (committed form is the oxfmt-formatted one the format gate accepts).
- Integration proof (`run-itest.sh`, real Postgres + MinIO, with PR #3222's jsonb fix applied temporarily and dropped afterwards): `[itest] 464/466 checks passed across 133/133 lanes`, no `RUN TRUNCATED`. The two failed checks are not this branch's: (1) `webdav re-home (protocol + tree + locks + visibility on pg)` — red on main before this campaign (known); (2) `knowledge: corpus scope drift is corrected and reported — drifted=null, first=corrected 0/scanned 12` — the lane #3222 un-truncates; its `LIMIT 1` (no `ORDER BY`) over `app.documents` picked a file-backed document with no `private_knowledge.documents` row, so the drift `UPDATE` matched nothing (`drifted=null`). This branch touches nothing under `domains/knowledge`, `domains/documents` or `core/knowledge`, and no REST lane runs before it. Every REST lane passed: `REST machine door`, `REST machine journey (projects → files → tasks → start)`, `REST resources (bulk, documents, entries, skills, chat, search)`, `REST resources: agent refusals map to 400/403/422/404, never 500`, `REST door rate-limit attribution`, `REST pagination walks to the last page`, `websites REST family`, `platform MCP endpoint (/api/v1/mcp)`, `documents write matrix: REST v1 door refuses read-only member`, `corpus scope: REST team change re-stamps retrieval scope`.

## Review follow-up (three commits on top)

- `1f3d5edbb` fix(platform): gate the REST file bind before its metadata row exists — the blocking rest-v1-4 ordering bug. `registerUpload` INSERTed the `app.file_metadata` row (`size = HEAD`) on the bind transaction BEFORE `validateDocumentUploadForOrg` summed that table for the uploader, so the file was counted twice under `maxTotalVolumeBytesPerUser` (headroom `S <= H < 2S` falsely refused with `volume_exceeded`). The bind now goes HEAD → policy → row like `createDocumentFromBlobUpload`; the retype `UPDATE` is gone (the row is born with the resolved type). Regression `counts the file it is admitting once against the per-user volume quota` (1234-byte file, 2000-byte quota, the fake sum includes the landed bytes only once `registerUpload` ran) is red on the previous order — `expected 400 to be 201` — and green now; refusals are pinned to happen before `registerUpload` is called; the resolved type is asserted on the `registerUpload` call; a missing blob answers 404 `BLOB_NOT_FOUND` and registers nothing.
- `17b64eb06` test(platform): pin the malformed-body 400 in every REST family — one `'{'` → 400 case in v1-tasks / v1-automations / v1-threads (v1-projects got its case in the commit above), each asserting nothing is written.
- `90c5332bc` docs(platform): name the org file:upload budget at the REST file bind — spec.ts + regenerated `public/openapi.json` (one line).

## Notes for the reviewer

- Follow-up: the missing-blob refusal at the bind is a `FileError('BLOB_NOT_FOUND', …, 404)` (the same envelope `registerUpload` answered for the same case before) rather than a bare `RestRefusal`, so the `code` stays on the wire. `registerUpload` still performs its own HEAD after the policy's `statOrgBlob` — one extra HEAD per bind, the price of reusing the existing register core unchanged (the session lane's `registerUploadedBytes` takes the size as an argument, but it answers an already-registered blob idempotently instead of refusing with 409 `BLOB_ALREADY_REGISTERED`, and it has no 512 MiB ceiling — the REST lane relies on both).

- `domainErrorResponse` now also unwraps a domain-wrapped rate-limit cause (`rateLimitExceededCause`) so a `DocumentError` coded `RATE_LIMITED` answers the one 429 shape with `Retry-After`; this is what makes the bind's `file:upload` refusal uniform with the other lanes.
- `pageLimit` was widened to accept an already-numeric body field (`string | number | undefined`) for the website search body limit.
- Commit 0dc3fff1a is a pure `bun run format` pass: the worktree carried no `.husky/_`, so the pre-commit formatter never ran on the earlier commits; SAST was run manually (0 findings). `public/openapi.json` in that commit is layout-only (`jq -S` identical).
- POST /threads keeps its `c.req.json().catch(() => ({}))` on purpose — an empty body is a valid "create with defaults" there.
- The `webdav re-home` itest lane is red on main before this campaign (known, not this theme).
- The REST bind now HEADs the blob twice: `statOrgBlob` for the gate, then `registerUpload` (kind `external`) re-HEADs as its own attestation. Folding that into one call would change `registerUpload`'s contract for its other callers; left as is (one extra S3 HEAD per bind).


larryro:
Integration proof re-run on the repaired tip (90c5332bc, with #3222's aa3d1d7a9 cherry-picked temporarily for the lane-19 fix, then dropped):

```
[itest] 464/466 checks passed across 133/133 lanes
```

The two remaining fails are the known main-side lanes (`knowledge: corpus scope drift is corrected and reported` — harness picks a document without a corpus row; `webdav re-home` — red on main before this branch). Every REST lane passes, including the project-file bind after the HEAD → validate → register reorder.

PR #3278, opened 2026-09-07
fix(ui): close remaining interaction taste gaps

AdeolaAdekoya:
## Summary
Follow-up to #3276 (stack on that branch). Closes the leftovers from the Emil taste audit:

- Remove nested `TooltipProvider`s from ChartCard, ProgressBar, and CopyableField so AppShell skip-delay works across analytics + credentials UI.
- Select / searchable-select / multi-select popovers get origin-aware transform, `--duration-short`, and `motion-reduce`.
- Drop `transition-colors` on multi-select, label-editor, and dropdown menu keyboard/focus rows.
- Wire `Button isLoading` on log-in, 2FA, credential create, skill save/upload, blank automation, and task create/setup submits.
- Dialog/Sheet overlays + ResponsiveDialog enter/exit get fade + tokenized duration + reduced-motion.

## Test plan
- [ ] Hover chart info tip then a rail tip — skip-delay still works (no extra wait).
- [ ] Open Select / MultiSelect / SearchableSelect — fade/slide, origin feels attached to trigger.
- [ ] Arrow through multi-select / dropdown / label editor — highlight jumps with no color fade lag.
- [ ] Log in / 2FA / create credential / save skill / create automation / create task — spinner on primary while pending.
- [ ] Open task modal (ResponsiveDialog) and settings sheet — overlay fades; reduce-motion disables motion.
- [ ] Merge/land only after #3276 (this PR targets that branch).

PR #3276, opened 2026-09-07
fix(ui): restore overlay motion and interaction taste

AdeolaAdekoya:
## Summary
- Mount a single AppShell `TooltipProvider` (`delay` 200ms / `skipDelay` 300ms) and stop wrapping every tip in its own Provider so toolbar tips open instantly after the first.
- Restore `tw-animate-css` so `animate-in` / slide / fade classes actually run again on dialogs, sheets, menus, and tooltips.
- Cap sheet open at `--duration-standard` (200ms) with ease-out; wire button press to `--scale-pressed` / duration tokens; origin-aware menus/popovers via Radix transform-origin vars.
- Form/Confirm primary actions use `Button` `isLoading` (spinner); drop `transition-colors` on keyboard-highlighted search/select rows.

## Follow-ups (not in this PR)
- Collapse dual platform vs `@tale/ui` dialog/tooltip skins into one vocabulary.
- Marketing web tip Providers (isolated tiles) can stay intentional for now.

## Test plan
- [ ] Hover adjacent toolbar tips (rail user/org/bell, composer icons) — first tip delays, subsequent tips are instant.
- [ ] Open settings sheet / dialog / dropdown — enter/exit animate; sheet feels snappy (&lt;300ms).
- [ ] Submit a form dialog / confirm dialog — spinner on primary while pending.
- [ ] Arrow through command palette / searchable select — highlight has no color transition lag.
- [ ] `prefers-reduced-motion: reduce` — overlays do not animate.
- [ ] `bun run --filter @tale/ui test` / platform `test:ui` for touched files.

PR #3279, opened 2026-09-07
fix(ui): unify tooltip skin and tighten craft/copy

AdeolaAdekoya:
## Summary
- Unify platform and `@tale/ui` tooltips on one inverted surface (`tooltipContentClassName`) so toolbar tips no longer mix accent vs fg/bg skins.
- Wire remaining submits to `isLoading` (forced password, data-residency test, website Load more) and use `CircleHelp` for interactive field tips; drop selected-row `transition-colors` on table rows.
- Tighten EN copy (softeners, `Copied`, `Couldn't` instead of `Failed to`) with matching de/fr tweaks; leave Agent vs Assistant alone (chat role vs workforce).

## Test plan
- [ ] Hover a platform tip and a Button `title` tip — same inverted chrome
- [ ] Forced password change: submit shows spinner via `isLoading`, label stays put
- [ ] Data residency Test connection / websites Load more: spinner, no label swap
- [ ] Label / assignee / usage tip icons are `CircleHelp`; error glyphs stay `Info`
- [ ] Copy a chat reply — tip says **Copied** (no bang); reconnect copy has no “hold tight”
- [ ] Spot-check a few former “Failed to …” toasts read as “Couldn't …”

Stacked on #3278; land after #3276 → #3278.


AdeolaAdekoya:
Superseded by #3280 — same stack collapsed into one PR against main.

PR #3289, opened 2026-09-08
fix(cli): fetch the images tale dev needs under a step that says so

larryro:
Two things found while cold-starting **v0.5.12** on a host wiped to zero images, containers and
volumes.

## 1. `tale dev` spent three minutes downloading under a label about volumes

The config-mountpoint helper runs on the app image, so a first bring-up waited for a 2.94 GB pull
inside a step that says nothing about a download. Measured against the released 0.5.12:

```text
[ + ] Volumes & networks ready (3m03s)
[ + ] Sandbox runtime image ready (7m40s)
```

The second wait is more than twice as long and nobody minds it — the label says what is happening.
The first reads as a hang, on the very first run of the quickstart.

Both images the CLI needs in its own hands before compose runs now come from one step that names
them, and the volume work goes back to being instant. The runtime image is the same fetch as before,
moved up; the app image is a plain pull when absent, best-effort, because compose pulls it moments
later and reports its own failure. `imagePresent` is now shared rather than asked in two different
sets of words.

## 2. A provider-only org has a knowledge base that cannot index, and nothing says so

An embedding model has no working default — `EmbeddingNotConfigured` is a refusal, never a guess. An
organization that adds a provider and stops there has knowledge that cannot be indexed or searched,
and until now the only places that said so were the data-residency page itself and the error on a
document that had **already** failed to index, which is after the reader spent the upload.

A dashboard banner closes the gap, in the shell beside the two-factor nudges so it is visible from
chat — where a reader lands after the setup wizard.

Two gates keep it from being a tax on everyone else:

- **Only once a provider credential exists.** Before that there is nothing to choose a model with,
  and the setup wizard is already asking for a provider; a second nudge there is noise on an empty
  org.
- **Only for a reader with `read orgSettings`** — the permission the page it links to requires. A
  member who cannot open it never sees it.

It clears itself the moment a model is configured, so it carries no dismissal. An unknown state stays
silent: a loading or failed read is not evidence of a missing model, and "knowledge search is off" on
an unknown state is worse than saying nothing.

## Verification

- `tools/cli`: `bun test` 436 pass / 0 fail, `tsc --noEmit` clean, `oxlint` clean. Four new cases
  cover fetch-only-when-absent, the pull, and the warn-not-throw path.
- `services/platform`: the touched UI projects run 9 files / 73 tests green; `tsc --noEmit` clean;
  `oxlint --type-aware` clean; i18n parity 50 tests green. Six new cases pin the banner's gates and
  its polite `status` role, asserting the shipped English strings so the catalog wiring is exercised.
- EN/DE/FR updated together.

**Not yet verified end to end:** the "after" run for change 1 needs the app image absent, and it is
currently pinned by the running 0.5.12 stack on this machine. The "before" number above is measured
from that same released binary; I will run the cold comparison once the stack is free.

### DIFF
--- a/services/platform/app/features/settings/data-residency/components/embedding-setup-banner.test.tsx
+++ b/services/platform/app/features/settings/data-residency/components/embedding-setup-banner.test.tsx
@@ -1,0 +1,136 @@
+// @vitest-environment jsdom
+import React from 'react';
+import { describe, expect, it, vi } from 'vitest';
+
+import { render, screen } from '@/tests/utils/render';
+
+// The banner reports a broken knowledge base, so it must be a polite `status`
+// live region like its dashboard siblings — an assertive `alert` would
+// interrupt screen-reader users on every page of the shell.
+
+const { mockState } = vi.hoisted(() => ({
+  mockState: {
+    canRead: true,
+    abilityLoading: false,
+    embedding: undefined as unknown,
+    embeddingError: false,
+    credentials: undefined as unknown,
+  },
+}));
+
+vi.mock('@/app/hooks/use-ability', () => ({
+  useAbility: () => ({
+    can: () => mockState.canRead,
+    cannot: () => !mockState.canRead,
+  }),
+  useAbilityLoading: () => mockState.abilityLoading,
+}));
+
+vi.mock('../hooks/queries', () => ({
+  useOrgKnowledgeEmbedding: () => ({
+    data: mockState.embedding,
+    isError: mockState.embeddingError,
+  }),
+}));
+
+vi.mock('@/app/features/settings/providers/hooks/queries', () => ({
+  useProviderCredentials: () => ({ data: mockState.credentials }),
+}));
+
+vi.mock('@tanstack/react-router', () => ({
+  Link: ({
+    to,
+    children,
+    className,
+  }: {
+    to: string;
+    children: React.ReactNode;
+    className?: string;
+  }) => (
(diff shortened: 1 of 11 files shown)


larryro:
Verified the step attribution without a 10 GB cold start, by pointing the CLI at a registry that does
not exist so both fetches fail immediately:

```text
[ - ] Fetching the sandbox runtime and app images...
[ x ] Failed to pull image: ghcr.io/…-does-not-exist/tale-platform:latest
[ ! ] Could not fetch …/tale-platform:latest; compose will try again.
[ + ] Images ready (2.6s)
[ - ] Preparing volumes & networks...
[ ! ] Could not prepare the org config mountpoints in probe-…-dev_config-data: Unable to find image …
[ + ] Volumes & networks ready (2.5s)
[ - ] Starting Tale...
[ x ] docker compose up failed
```

That covers the three claims:

- **Attribution** — the app-image fetch is attempted and reported in the step that names it, not
  inside "Preparing volumes & networks". That step no longer owns a download.
- **Best-effort** — both failures warn and the run continues; compose reports the real error once,
  rather than the CLI throwing first and hiding it.
- **Graceful degradation** — the mountpoint helper says which volume it could not prepare instead of
  failing the bring-up.

What this does not measure is the wall-clock improvement on a real cold host; the 3m03s "before" in
the description is measured from the released 0.5.12 binary on exactly that host.

PR #3295, opened 2026-09-08
feat(platform): support external Postgres and external S3

larryro:
## Why

The three stores could already be pointed at infrastructure an operator brings, but the **service** still behaved as if they were containers next to it. Five gaps, all of which fail quietly:

| Gap | What it cost |
| --- | --- |
| `OBJECT_STORE_*` read on the **first boot only** | rotating an external bucket's key meant hand-editing a SOPS-encrypted file inside a container volume |
| `ensureBucket` issued an unconditional `CreateBucket` | a least-privilege key answered 403 → connection never landed → every upload refused, with only a boot warning |
| endpoint mandatory, path-style hardcoded | AWS S3 proper could not be the deployment default at all |
| no probe saw a store go away | `/ready` reads a drain flag, `/ping` is static — a replica with dead credentials kept taking traffic |
| two drivers, one URL, two TLS meanings | `verify-ca` checked the hostname, and node-postgres let the connection string override the `ssl` option beside it |

## What changed

**The object store follows the environment.** Variables are read on every start and `default/object-storage/connection.json` is reconciled from them, so repointing the bucket or rotating its credentials is an env change plus a restart. A file marked `"managedBy": "operator"` is never touched — the guarantee the old "never overwrite an existing file" rule gave, now explicit rather than incidental. A pre-marker file is adopted only if it still names the same bucket at the same endpoint; if you had repointed it by hand, your edit wins. Boot logs `seeded` / `reconciled` / `adopted` / `ignored` / `skipped`.

**Least-privilege keys work.** `HeadBucket` first; create only what is absent; 403 on HEAD means "exists, you may not list it", which is still perfectly usable for objects.

**AWS S3 proper is a first-class target.** The endpoint is optional now (absent ⇒ virtual-host addressing), joined by `OBJECT_STORE_REGION`, `OBJECT_STORE_FORCE_PATH_STYLE` and `OBJECT_STORE_PREFIX` — all three already existed in the connection schema and simply had no environment spelling.

**One owner for Postgres TLS.** `backend/db/ssl.ts` parses the URL once, strips the TLS parameters so neither driver can reinterpret them, and hands identical explicit options to all five seams (app pool, boot migrator, pg-boss, Better Auth, knowledge pools). `POSTGRES_CA_FILE` supplies a PEM bundle for the roots managed providers sign with and Node does not ship. `verify-ca` now checks the chain without the hostname, which is what it is defined to do.

**`tale_backend_store_up{store=…}`** probes all three stores, time-bounded and TTL-cached, and logs transitions. Deliberately *not* wired into `/ready`: a flapping bucket must not cut a colour out of DNS.

**`DATABASE_POOL_MAX`**, because an external database has a connection budget the code should not assume (it costs this twice per replica — app pool plus job queue).

## Slightly outside the brief

Larry scoped me to the service, not compose. Two lines fall outside that and I think they should land anyway: the shipped compose lanes pinned `DATABASE_URL` and `OBJECT_STORE_ENDPOINT` in `environment:`, which outranks `env_file`, so an operator setting either in `.env` was silently ignored — while the docs promised it worked. They now pass the variable through. Easy to drop if you disagree.

Relatedly, the docs advertised `POSTGRES_URL` as the way to reach an external Postgres and **no code has ever read it**; it is replaced with the real `DATABASE_URL`, which was not documented at all. The two constraints that actually bite are now written down in all three locales: the corpus database needs `pgvector` pre-installed (Tale creates schemas and tables, never extensions), and neither database may sit behind a transaction-mode connection pooler (the job queue holds `LISTEN`, the migrator holds a session advisory lock, queries use prepared statements).

## Verification

Hand-written compose standing up **only** a customer's Postgres and MinIO — no bundled `db`, `knowledge-db` or `object-store` — with the real backend pointed at them:

- 83 app migrations land in the external Postgres; the corpus schemas resolve; the bucket is created in the external MinIO.
- Rotating `OBJECT_STORE_SECRET_KEY` to a second MinIO user → `reconciled`, sidecar rewritten, uploads keep working.
- A MinIO user whose policy denies `s3:CreateBucket` (confirmed denied out of band) → boots and seeds against a pre-created bucket. This is the case that used to fail closed.
- `"managedBy": "operator"` + a different `OBJECT_STORE_BUCKET` → `ignored`, file byte-identical, actionable log line.
- Stopping the external MinIO → `tale_backend_store_up{store="object_store"} 0` within the TTL, a log line naming the store, and `/ready` still `200`.
- Postgres behind a self-signed private CA: `verify-full` + `POSTGRES_CA_FILE` connects; the same URL without the CA fails at boot with `unable to verify the first certificate`; `sslmode=require` connects without one. `pg_stat_ssl` shows `postgres.js`, `tale-backend` (pg-boss) and Better Auth's pool all on TLSv1.3 — the disagreement is gone.

Gates: `bun run check` (43/43 tasks, ~77k tests), `lint:sast` (0 findings), `lint:manual`, `knip:check`. No migration — nothing here changes the schema.

New manual boxes `DATA-F8`, `DATA-F9`, `DATA-B6` cover the deployment-default lane, which is environment-driven and has no UI surface.

## Not in scope

STS / `sessionToken` / instance-role credentials (IRSA, EC2 instance profile) and SSE-KMS. Real support needs a refreshing credential provider chain; a static session token in an env var would be half a feature. Worth its own PR if AWS-native deployments need it.


larryro:
Follow-up commit `adc9e7bed` — the docs sweep this PR owed.

The first commit updated `environment-reference.md` and `data-residency.md` and stopped there, which left the rest of the self-hosted journey describing the old behaviour. Worse, it missed **Run Compose yourself** — the page whose entire audience is people running Tale against infrastructure they already have.

| Page | What it now says |
| --- | --- |
| `install/own-compose.md` | New **Stores you already have**: which service each variable lets you delete (plus the `depends_on` entries that go with it), SQL to prepare the databases, the bucket + CORS policy, a worked `.env`, and how to confirm from the boot log and `tale_backend_store_up`. Its env table listed `DATABASE_URL` and `OBJECT_STORE_ENDPOINT` as fixed wiring to the bundled services — they are the knobs. |
| `operate/backups-and-restore.md` | The thing nothing in the product says: blobs in an external bucket are announced and skipped, an external **database** is not — the snapshot looks complete and holds none of it, and a restore rolls local volumes back while that database stays put. |
| `operate/observability/troubleshooting.md` | New **Every upload is refused**, keyed by the boot-log line, ending on the CORS policy the server-side connection test structurally cannot catch. |
| `operate/observability/prometheus-grafana.md` | `tale_backend_store_up` panel + `TaleStoreUnreachable` rule. The page already named knowledge-database reachability as worth paging on, with no series to page on. |
| `overview.md` | Said the corpus was the first thing you could move off the box; all three stores move now, each on its own. |

All three locales, authored natively. One thing worth flagging: the docs contract says the link checker does not verify anchors, so I computed the three new anchors through `slugifyHeading` rather than assuming — the French heading decomposes (`déjà` → `deja`) and my first attempt was wrong.

Gates: `bun run check` 43/43, docs suite 202/202, `oxfmt --check` clean.

PR #3326, opened 2026-09-11
fix(platform): show archived work in search, labelled as archived

### LINKED ISSUE #2999: Decision: should an archived project's documents still answer RAG searches?
## The current behaviour

Archiving a project does **not** affect knowledge retrieval. Its documents remain fully searchable and citable, and nothing indicates that.

`convex/documents/access.ts` builds the caller's readable project set by walking every project in the org and gating only on access:

```ts
const projects = ctx.db
  .query('projects')
  .withIndex('by_organization', (q) => q.eq('organizationId', args.organizationId));
for await (const project of projects) {
  if (hasProjectAccess(project, context.teamIds, context.role)) {
    projectIds.push(project._id);
  }
}
```

`archivedAt` is never consulted — not here, not in `projects/access.ts`, not in the SQL pre-filter.

## Why it's a question rather than a bug

The product already takes a position that removed content should stop answering: `lifecycleStatus` (`active | trashed | expired | deleted`) **is** honoured, via `isActiveDocument` in `documents/filter_retrievable_rag_file_ids.ts`. A trashed document is not retrievable.

So the asymmetry is deliberate-looking: trash means gone, archive means put away. Whether "put away" should also mean "stops answering questions" is a product call, and both readings are defensible:

**Archive should exclude.** Someone archives a project to retire it. If its documents keep informing chat answers, superseded material silently competes with current material, and the person who archived it has no way to know. That is the same failure shape as a stale label: the UI says "not in play", retrieval disagrees.

**Archive should not exclude.** Archive is explicitly not deletion. Retired projects are often exactly what you want to search — past contracts, finished engagements, previous decisions. Trash already exists for "stop using this", and overloading archive removes the only non-destructive way to keep material searchable while clearing it from the working set.

**A middle option:** keep archived material retrievable but demote it in ranking, and mark citations from archived projects so an answer visibly says the source is retired. More work, and needs the corpus to carry the archive flag.

## If the answer is "exclude" — two gates, not one

Retrieval enforces scope twice, and both need to agree or the result is a silent half-fix:

- **SQL pre-filter** — `convex/knowledge/corpus.ts`, `d.project_id = ANY($n)` against the readable set. Fixing only the Convex side leaves archived material occupying result slots before being filtered out, which degrades recall for the material that *should* be there.
- **Convex-truth re-check** — `documents/filter_retrievable_rag_file_ids.ts` → `knowledgeScopeAllows`, reading the live document row. This one fails closed, so fixing only here is safe but wasteful.

Excluding at the `projectIds` source (the loop above) covers both, since both derive from that set. That is probably the smallest correct change — but note it also affects every other consumer of `knowledgeAccess()`, so the blast radius wants checking before assuming it is a one-liner.

## Related

- `hasProjectAccess` returns true for owners/admins before consulting teams, so an admin's searches already span every project — archived ones included. Whatever is decided, admins are the case most likely to notice.
- A project with no teams is org-wide by definition (`getProjectTeamIds(...).length === 0`), so an archived teamless project's documents are archived *and* org-wide readable.
- Whatever is decided should be written down where the scope rules already are (`lib/knowledge/types.ts`, beside `documentScopeKind`), since that is now the single place scope questions are answered.

No action proposed until the product call is made.

### LINKED ISSUE #3325: Bug: search hides archived work, contradicting the #2999 decision
The ⌘K palette returns no archived tasks or projects, and no rows that belong to an archived project. #2999 decided the opposite in August, and #3007 implemented that decision for chat retrieval only.

## Why

#2999 decided that archived material stays searchable and citable, and that results say they are archived. #3007 shipped two fields for it: `archived` (the row itself) and `projectArchived` (its container). Both survived the Postgres port and live in `backend/core/chat/assistant_tools.ts:126`.

#3007 scoped the UI out on purpose: "Not included: any UI change." Nothing tracked the rest, so the palette still runs the pre-decision predicate.

The two paths now disagree about the same rows:

| Path | Archived tasks |
| --- | --- |
| Agent, `chat/shim.ts:195` | returned, `archivedAt` stamped on the row |
| Palette, `tasks/service.ts:2764` and `:2803` | excluded, with no opt-in |

`projects/service.ts:1884` excludes archived projects the same way.

A third effect is wider than either. `searchTasks` builds its project set from `listProjects`, which defaults `includeArchived` to false (`projects/service.ts:1711`). So a live task in an archived project is unfindable. That is the case #3007 named: "A live task in a retired project may still be work nobody closed." `searchDocumentsView:1289` resolves projects the same way, so those projects' files go too.

No test pins any of this. Deleting either predicate turns nothing red.

## What to change

- `SearchResult` has no field for a label, so a hit cannot say it is archived (`packages/ui/src/components/search/types.ts:13`). Add one.
- `searchTasks` and `searchProjects` return archived rows and label them.
- Those two, plus `searchDocumentsView`, resolve projects with `includeArchived: true` and label `projectArchived`.
- Live rows sort ahead of archived ones. Search caps at 25 rows, so without an explicit order archived work can crowd live work out of the page.

## Scope

Chats and contacts stay as they are. `searchChats` excludes archived threads under a different vocabulary and #2999 never covered it. Contact trash is a hard fence that #3007 deliberately left alone.

Israeltheminer:
The ⌘K palette now returns archived tasks and projects, and rows belonging to an archived project, each labelled **Archived** or **Archived project**. Live rows still list first.

## Why

#2999 decided that archived material stays searchable and that results say they are archived. #3007 shipped the two labels for it, `archived` and `projectArchived`, and scoped the UI out: "Not included: any UI change." Nothing tracked the rest, so the palette kept the pre-decision predicate while the agent path implemented the decision. The same query answered differently depending on who asked.

The wider effect was the project set. `searchTasks` built its readable projects from `listProjects`, which defaults `includeArchived` to false, so a live task in an archived project was unfindable. That is the case #3007 named: "A live task in a retired project may still be work nobody closed." `searchDocumentsView` resolved projects the same way, so those files went too.

## What changed

- `SearchResult` gains `badge`, rendered as a chip beside the title. The contract had no field a hit could say this in.
- `searchTasks` returns archived tasks, reads archived projects, and stamps both labels in one place so its two legs cannot disagree.
- `searchProjects` returns archived projects. `PROJECT_COLUMNS` already carried `archived_at_ms`.
- `searchDocumentsView` reads archived projects and stamps `projectArchived`. Trash is untouched: `lifecycle_status` stays a hard fence.
- Live rows sort ahead of archived ones, in the SQL page order and again in the merge. Archiving bumps `updated_at_ms`, so under a recency-only sort an archived row takes the freshest slot on a page capped at 25.
- Badge copy in en, de and fr.

Archived work also says so where it lives, not only in search:

- A board card and a list row carry the **Archived** badge. Colour was the sole cue (`opacity-70` at `task-card.tsx:121` and `tasks-list.tsx:326`), which WCAG 2.1 AA 1.4.1 does not allow. `TASK-F5` already claimed the rows wore this badge; now they do.
- An archived project carries the badge beside its breadcrumb name on every tab. It previously announced itself only in the Projects list, so a retired project looked live from inside.

The tasks page is affected by the same defect through a different route. It renders `board ∩ search` (`tasks-workspace.tsx:166`), so the board read is a ceiling the search read can only narrow. With **Show archived** on, the board returned archived tasks and `searchTasks` did not, so typing in **Search tasks** silently removed the rows the toggle had just revealed. Widening `searchTasks` makes the two agree; the toggle still governs.

## Risk

A search page is capped at 25 rows and archived rows now compete for it. The two sort keys keep that from costing live results. The lane `a capped page keeps live tasks and drops the archived one` pins it at exactly the cap.

Chats and contacts are unchanged. `searchChats` excludes archived threads under its own vocabulary and #2999 never covered it; contact trash is a fence #3007 left alone.

## Tests

Real Postgres, 630/631 checks across 147/147 lanes. The one failure is `tasks: a burst of serializable commenters`, which fails on clean `origin/main` too and varies run to run (`attempts` 26 and 28 against a `≤24` bound; once `landed=10/12`).

| Assertion | Mutation that turned it red |
| --- | --- |
| An archived task is returned and labelled | Restore `AND t.archived_at_ms IS NULL` in the field leg |
| A live task in an archived project is returned | Restore `listProjects(sql, auth)` without `includeArchived` |
| Live rows sort ahead of archived ones | Drop the band from the merge sort |
| A capped page keeps live rows | Drop the band from the SQL page order |
| An archived project is returned by project search | — |
| A row renders its badge | Render the badge never |
| A row without one renders none | Render the badge always |
| Show archived plus a query keeps the task | Make the board ignore `includeArchived` |
| A board card badges when archived | Force the card condition false, then true |
| A list row badges when archived | Force the row condition false, then true |
| A project breadcrumb badges when archived | Force the breadcrumb condition false, then true |

The third mutation showed the merge sort masks the SQL order in a two-row fixture, so the fourth assertion builds 25 live rows plus one archived to pin the SQL key on its own.

That lane asserts both reads parsed before reading them. Its first draft required an `_id` the board does not return, so both parses failed and the `want false` half passed while proving nothing.

The card's negative case needed a second pass. Forcing the badge to always render did not fail it: the fixture had no project key, so `formatTaskIdentifier` returned null, the identifier row was skipped, and the badge could not appear either way. Both fixtures now render that row, and the permissive mutation fails.

Manual boxes `NAV-F15`, `TASK-F18` and `PROJ-F22` cover what a headless run cannot judge: the chip's weight beside the title, dark mode, and 400px.

## Scope

Closes #3325. Does not close #2999, which is already closed.

Gate: `bun run check` green (43/43) apart from two pre-existing failures reproduced on clean `origin/main` — `@tale/docs` `deploy-sim.test.ts` under concurrency (passes in isolation, 206/206) and two `@tale/sandbox-runtime-daemon` host-dependent tests, in a workspace this branch does not touch. `knip:check` green; `lint:manual` 777 boxes; `lint:sast` 0 findings in changed files.

### DIFF
--- a/packages/ui/src/components/search/search-command.test.tsx
+++ b/packages/ui/src/components/search/search-command.test.tsx
@@ -332,4 +332,30 @@ describe('SearchCommand', () => {
       expect(screen.getByRole('dialog')).toHaveAttribute('aria-modal', 'true');
     });
   });
+
+  // A result's status label. The palette shows archived work (#2999), so the
+  // row has to be able to say a hit is archived.
+  it('renders a result badge beside the title', async () => {
+    respond = () => ({
+      kind: 'ready',
+      results: [result({ id: '1', badge: 'Archived' })],
+    });
+    const { user } = renderCommand();
+    await user.type(screen.getByPlaceholderText(L.placeholder), 'config');
+    await waitFor(() =>
+      expect(screen.getByText('Archived')).toBeInTheDocument(),
+    );
+  });
+
+  it('renders no badge when a result carries none', async () => {
+    respond = () => ({ kind: 'ready', results: [result({ id: '1' })] });
+    const { user } = renderCommand();
+    await user.type(screen.getByPlaceholderText(L.placeholder), 'config');
+    await waitFor(() =>
+      expect(
+        screen.getByRole('option', { name: /Configuration/ }),
+      ).toBeInTheDocument(),
+    );
+    expect(screen.queryByText('Archived')).not.toBeInTheDocument();
+  });
 });
--- a/packages/ui/src/components/search/search-result-row.tsx
+++ b/packages/ui/src/components/search/search-result-row.tsx
@@ -108,6 +108,11 @@ export function SearchResultRow({
             <span className="truncate">
               <Highlight text={result.title} terms={highlightTerms} />
             </span>
+            {result.badge ? (
+              <span className="border-border-base/70 text-fg-subtle shrink-0 rounded border px-1.5 py-px text-[10px] font-medium">
+                {result.badge}
+              </span>
+            ) : null}
           </span>
           {secondary ? (
             <span className="text-fg-muted mt-0.5 line-clamp-2 text-xs leading-relaxed">
--- a/packages/ui/src/components/search/types.ts
+++ b/packages/ui/src/components/search/types.ts
@@ -18,6 +18,9 @@ export interface SearchResult<TData = object> {
   /** Secondary line under the title (a chat preview, an entity subtitle).
    *  Ignored when `body` is present — `body` drives a highlighted snippet. */
   subtitle?: string;
+  /** Short status label rendered beside the title (e.g. "Archived"). The
+   *  source supplies it already localized. */
+  badge?: string;
   /** Group key — section for docs, date-bucket for threads, entity type for
    *  mixed lists. Falls back to a single "Results" group when omitted. */
   group?: string;
--- a/services/platform/app/features/documents/data/documents-search-source.ts
+++ b/services/platform/app/features/documents/data/documents-search-source.ts
@@ -4,6 +4,7 @@ import type { SearchResult, SearchSource } from '@tale/ui/search';
 import { useMemo } from 'react';
 
 import { useBackendQuery } from '@/app/hooks/use-backend-query';
+import { useT } from '@/lib/i18n/client';
 
 const NO_RESULTS: SearchResult<DocumentSearchHitData>[] = [];
 
@@ -19,6 +20,7 @@ export function createDocumentsSearchSource(options: {
 }): SearchSource<DocumentSearchHitData> {
   const { organizationId, enabled = true } = options;
   return (query, { active }) => {
+    const { t } = useT('dialogs');
     const trimmed = query.trim();
     const hits = useBackendQuery(
       'documents/search:searchDocuments',
@@ -34,6 +36,11 @@ export function createDocumentsSearchSource(options: {
         id: hit.documentId,
         title: hit.title,
         subtitle: hit.snippet,
+        // A document has no archive state of its own; its project's is the
+        // only label it can carry (#3007).
+        ...(hit.projectArchived
+          ? { badge: t('search.badgeProjectArchived') }
+          : {}),
         group: 'documents',
         data: {
           kind: 'document' as const,
@@ -41,7 +48,7 @@ export function createDocumentsSearchSource(options: {
           projectId: hit.projectId,
         },
       }));
-    }, [hits.data]);
+    }, [hits.data, t]);
 
     if (!enabled || !active || trimmed.length === 0) {
       return { results: NO_RESULTS, status: 'ready' };
--- a/services/platform/app/features/projects/components/project-archived-badge.tsx
+++ b/services/platform/app/features/projects/components/project-archived-badge.tsx
@@ -1,0 +1,20 @@
+'use client';
+
+import { Badge } from '@tale/ui/badge';
+import { Archive } from 'lucide-react';
+
+import { useT } from '@/lib/i18n/client';
+
+/**
+ * Inline badge when a project is archived — the project page's breadcrumb
+ * leaf, so a retired project says so on every one of its tabs rather than
+ * only in the Projects list.
+ */
+export function ProjectArchivedBadge({ className }: { className?: string }) {
+  const { t } = useT('projects');
+  return (
+    <Badge variant="yellow" icon={Archive} className={className}>
+      {t('archived.badge')}
+    </Badge>
+  );
+}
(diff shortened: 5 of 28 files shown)


Israeltheminer:
**Smoke test is red for a reason outside this branch.** Docker Hub no longer serves `minio/minio` anonymously:

```
$ curl -s -o /dev/null -w "%{http_code}" -H "Authorization: Bearer $ANON_TOKEN" \
    https://registry-1.docker.io/v2/minio/minio/manifests/RELEASE.2025-04-22T22-12-26Z
401
$ curl -s https://registry-1.docker.io/v2/minio/minio/tags/list -H "Authorization: Bearer $ANON_TOKEN"
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required", ... "Name":"minio/minio","Action":"pull"}]}
```

In the job that surfaces as `pull access denied for minio/minio`, then `object-store` never becomes healthy, then `db` / `knowledge-db` / `object-store` each time out at 300s and the job is cancelled. Two full reruns of the Build workflow reproduce it identically.

The tag is pinned at `tools/cli/src/lib/compose/types.ts:191` and `tools/cli/src/lib/deployment/runtime-test-helper.ts:63`. This branch touches neither, and no compose, image or Dockerfile. `main`'s Build passed at 11:05 today and this started failing at 21:40.

It is not specific to CI: a self-hosted install that pulls the image fresh hits the same 401. Worth its own issue and fix; deliberately not folded in here.

Every other check is green: 44 success, 2 skipped (fork-PR variants), 1 neutral (Trivy).

Israeltheminer:
**Resolved, and my earlier comment's call to action is stale.** #3328 landed on main at 00:31 today and switches every reference to `quay.io/minio/minio`. This PR's checks run against a merge with main, so Smoke test is green here now: 45 success, 2 skipped (fork-PR variants), 1 neutral (Trivy), 0 failures.

The cause I described holds. Docker Hub still refuses `minio/minio` anonymously, and it is not a rate limit — a control image returns 200 from the same address with `x-ratelimit-remaining: 100;w=3600`, while `minio/minio` returns 401 `authentication required`. #3328 reached the same conclusion independently. No separate issue is needed.

PR #3349, opened 2026-09-14
feat(platform): let the chat Projects section collapse

AdeolaAdekoya:
## Summary
- The chats panel's **Projects** header is now a disclosure: hover (or keyboard focus) reveals a chevron; click folds the whole folder list so **Chats** stays on screen when an org has many projects.
- The choice persists per device. **New project** stays on the header and still works while collapsed; creating a project re-expands the section.
- On a narrow viewport the chevron stays visible without hover. Individual project folders keep their own expand/collapse.

## Test plan
- [ ] Open the chats panel with several projects so the folder list would otherwise push **Chats** down.
- [ ] Confirm **Projects** shows no chevron at rest on desktop; hover the header and a chevron appears next to the label.
- [ ] Click **Projects** — folders hide, **Chats** is in view, **New project** still works.
- [ ] Click again — folders return. Reload the page — the collapsed/expanded choice is kept.
- [ ] Keyboard: Tab to **Projects**, Enter/Space toggles; focus ring is visible.
- [ ] Create a project while the section is collapsed — the section expands so the new folder is visible.


yannickmonney:
Closing: superseded on `main` by a different, deliberate resolution of the same problem. Instead of making the PROJECTS block a disclosure, the chat sub-panel now caps PROJECTS at a strict half of the available height with its own scroller and gives CHATS the rest (`thread-list.tsx`, `max-h-1/2 min-h-0` + per-section `overflow-y-auto`), so a long folder list can never bury Chats and nothing has to be collapsed to get them back. Per-folder collapse is unchanged. Branch deleted as part of a stale-branch sweep.

PR #3352, opened 2026-09-14
fix(platform): align dialog header actions with close

AdeolaAdekoya:
## Summary
- Contact details looked lopsided because **New email** lived in the title row (a 36px icon button) while **Close** was absolutely parked in the corner at a different size.
- Header actions and Close now share one `items-center` cluster. Contact details uses the `sm` icon-button size so the envelope and X match.

## Test plan
- [ ] Open a contact row → **Contact details**. Envelope and Close sit on one line, same size, same inset from the top-right.
- [ ] If the contact is editable, **Edit** sits between them in the same cluster.
- [ ] A dialog with no header actions still has Close in the top-right corner.
- [ ] A drill-in dialog with Back still keeps Close on the back row.


yannickmonney:
Closing: this fix has already landed on `main`. The dialog component moved into `@tale/ui` (`packages/ui/src/components/dialog/dialog.tsx`) and carries exactly this change — the close control is rendered inside the `headerActions` cluster (`{headerActions}{!hideClose && !onBack && <DialogCloseButton />}`), the `absolute top-3 right-4` float is guarded with `&& !headerActions`, and the header row centres when there is no description. Main adds a `headerActionsPlacement` variant on top. Branch deleted as part of a stale-branch sweep.

PR #3353, opened 2026-09-14
fix(platform): close the 2026-09-14 API evaluation's seventh-pass findings

larryro:
## Summary

The seventh external black-box evaluation (2026-09-14, v0.5.25, contract 1.8.0) reported 3 S2, 24 S3, 34 S4 and 9 lead-lane findings. Every finding was second-confirmed against the source (six investigator reports); all three S2 are real defects and are fixed at the root with regression tests and integration lanes. The S3/S4 split three ways: code defects fixed, contract-precision items closed in the OpenAPI document and the en/de/fr docs, and thirteen product gaps recorded in `.agents/repo.md`'s contract debt ledger with their pay-down. The contract moves 1.8.0 → 1.9.0 (additive), with two rolling-safe migrations (0102, 0103).

## The S2s, root cause → fix

- **A mirrored conversation re-homed onto a recreated contact.** `DELETE /api/v1/contacts/{id}` frees the contact's `externalId`, and the next `POST /api/v1/conversations/sync` re-resolved the conversation's `externalContactId` onto whatever contact now carried it — an unrelated contact inherited the history. A content snapshot for a conversation whose contact is in the trash now answers 409 `CONVERSATION_CONTACT_TRASHED`, nothing applied; a `deleted: true` teardown still closes the mirror; the receipt reports `contactStatus` and the bound `externalContactId`, and an existing conversation's contact is never re-resolved. Integration record walks the whole choreography.
- **Documents released from a deleted project stayed unfindable.** The project delete in `detach` mode (and the app's delete door) released documents without re-stamping the corpus rows that retrieval pre-filters on (`project_id`, `team_ids`, `folder_path`), and a `retry-indexing` reported success while changing nothing. Both delete doors now batch re-stamp the released documents after the commit, and the indexer's skip-unchanged branch re-stamps off the current scope, so a retry heals a drifted stamp.
- **Six concurrent task starts began five runs.** The start guard was a check-then-act inside a SERIALIZABLE transaction whose snapshot froze before the advisory lock, so every racer read "no live run". The rule now lives in the schema — migration 0102 collapses existing live duplicates onto the earliest and adds the partial unique index `automation_runs_one_live_per_task` — the REST door starts under READ COMMITTED with the task reload still inside the transaction, and a loser's unique violation is reconciled by re-reading the winner behind a savepoint (`already_running`, never a second billable run). Integration lane: six concurrent starts → one run.

## Confirmed and fixed (S3/S4)

- **Runs.** `Run.failureCode` / `RunSummary.failureCode` — one enum over the engine's, the provider's and the agent turn's causes (migration 0103, written by `finishRun`); a cancel clears the stale park `detail` and answers the run's `status` beside `cancelled`; `GET /api/v1/me` answers `capabilities.developer`; `startedAt`'s meaning and where authoring lives are documented.
- **Indexing.** `indexing.errorCode` is the closed RAG vocabulary: `unsupported_type`, `image_no_vision`, `empty`, `not_text`, `malformed` are terminal (`unsupported`, a retry reproduces the answer and is refused as terminal), the rest `failed` with the job retrying `embedding_upstream`/`indexer_error`/`index_rebuilding`. `ExtractionError` carries the terminal causes out of the text and PDF extractors; NUL is stripped before the PII policy and the chunker (a `.txt` that was a binary used to fail the chunk INSERT after the embedding was paid for); the raw parser/provider sentence stays in the platform log. The stored `mimeType` is resolved from the extension first (`json`, `yaml`, `yml` mapped). `GET /api/v1/projects/{id}/files/{documentId}` reads one file row with its `indexing` state (closes the ledger row that said there was none).
- **Crawler.** A page the crawler looked at and stored nothing for is a failure of its own kind (`unsupported_content`) instead of a row that reads unfetched; `X-Robots-Tag: noindex` is honoured (`robots_noindex`); a TLS handshake failure is `tls_error`; a same-host link naming a port is never dialed (the ~200 `network_error` rows the tester saw); a scan that stored nothing ends on `error` with the reason instead of `active`; `Website.status` is an enum and `?status=` refuses a value outside it; website search is documented as BM25 with `score` semantics.
- **CRM and contract doors.** `Idempotency-Key` blank or over 255 characters → 400 `INVALID_HEADER` (a blank header no longer reads as "no key"); `POST /contacts/bulk` judges rows one by one (`errors[].issues`, 201 for the batch); `contacts?source=` is the closed set; `products.imageUrl` is held to the crawl-target host rule as a string; a contact, product or project `PATCH` that changes nothing writes nothing; `If-Match` on the document update → 412 `PRECONDITION_FAILED` with `data.etag`; the REST 429 carries a sentence and `requestId` like every other refusal; a body string with a NUL or a lone surrogate is refused with its path; `Contact.email` and `Product.imageUrl` carry their formats; `Automation.document.version` is the integer it always was.
- **Chat and MCP.** A reply that settles with no text carries no empty text part (a reasoning-only cut keeps its `reasoning` part); a call the reply cap withheld says whether its arguments were cut or complete; every MCP string argument refuses a blank at the transport (-32602).
- **Edge and app.** A malformed HTTP/1.1 chunked request body answers 400 `BODY_CHUNK_MALFORMED` at the edge instead of the 502 outage envelope with retry advice; `//api/v1/…` answers the JSON 404; `robots.txt` no longer pairs `Allow: /` with `Disallow: /` and carves out the developer pages; the `/docs` page links the developer guides and `/openapi.json`, the user menu's Documentation item uses the docs-site constant, and the app shell's `<noscript>` names both; `bluetooth` dropped from the Permissions-Policy; the compression floor sits in the OpenAPI description; the tutorial's `tale_...` placeholder is `<api-key>`.

## Found by the second confirmation (fixed)

- Contacts had the same no-op `PATCH` defect as products (every retry moved `updatedAt`, wrote an audit row and raised an event).
- The extension-first type rule reaches the OneDrive import too — its test pinned the declared type winning over `a.docx`.
- `tests/manual`: `KNOW-F14` (skipped-page reasons and the scan-end verdict), the automation coverage map, error-code rows; `app/routes/**` component tests are owned by the `ui` vitest project (the server project collected the new docs-page test).

## By design, no change

- Webhook delivery dedupe is by the delivery id, never the body (a 409 would loop a vendor's redelivery forever) — the parameter and the 202 now say so, with the header precedence list.
- Automation authoring stays on MCP / the app / `tale deploy`; the Automations tag and the reference row point there.
- The accepted-send queue runs in deployment-wide batches of `WORKER_CONCURRENCY`; documented as waves, no `queuedAhead` yet (ledger).

## Recorded as contract debt (not built this round)

Run usage/cost, REST twins for approvals and asks, task archive/delete over REST, `inputsAcceptDelivery` on a webhook bind, `repeatsExhausted`, `scanStartedAt` + `<meta name="robots">` + crawl knobs, website-search diagnostics / a `websiteId` filter on knowledge search, `Idempotency-Key` on the task start, `queuedAhead`, Office parse errors as `malformed`, `/.well-known/security.txt`, a changelog feed, an SDK / per-code table — each row in `.agents/repo.md` carries the design.

## Verification

- `bun run check`: platform 685 test files / 75,420 tests green, docs 206/206, UI and shared packages green. `@tale/cli#test` was red locally only because the gitignored embedded build was regenerated from the dirty tree (`EMBEDDED_CLI_BUILD.clean: false`); with the flag flipped the file passes 25/25, and CI builds it from a clean checkout.
- `backend:integration` against a throwaway Postgres + MinIO: 654/654 checks across 148/148 lanes, including the four new round-g lanes (trashed-contact mirror; JSON + noindex listed pages; an all-503 site → `error`; six concurrent starts → one run). Migrations 0102/0103 applied at boot; both are rolling-safe (the previous image never violates the index and neither reads the new column).
- `knip:check`, `lint:manual`, the OpenAPI/error-code guards and the docs structural suite are green. German and French docs were drafted natively and reviewed sentence by sentence.


larryro:
Three follow-up commits on top of the round-g set:

- **`fix(e2e): close preview connections so Bun's http shim stays alive`** — the docs Playwright job died three tests in: `vite preview` under Bun 1.4.2 threw `ERR_STREAM_WRITE_AFTER_END` out of its node:http response pipeline after a large response and exited, so every later test met `ERR_CONNECTION_REFUSED`. Running the preview under Node is not an option (`@tale/ui`'s sources use extensionless imports Node's resolver refuses), so the docs and web preview servers answer `Connection: close`; nothing is ever queued behind a closing socket. Both suites pass locally against a fresh build.
- **`fix(e2e): point the user-menu documentation assertion at the docs origin`** — the navigation spec pinned the retired `https://tale.dev/docs` href; the item links `TALE_DOCS_URL` now.
- **`fix(platform): file documents a folder sync adopts under the sync folder`** — reported on the live instance today: ticking a OneDrive folder for a sync import left every file at the hub root with no folder row, and therefore no "Stop syncing" entry (it lives on the folder row). The importer created the folder chain on the new/changed path only; a file the hub already held was adopted by the folder's config (bound, updated, pruned) but never moved, and a folder whose every file was adopted never existed. Both cloud-drive importers now create the sync root as soon as the config exists and re-file an adopted-unchanged document into the folder its selection names (folder id, folder path, corpus scope stamp); one-time re-imports and directly picked files keep their placement. Documented in en/de/fr; unit tests for both importers; the OneDrive integration lane imports a file at the root first and asserts the folder sync adopts it into `ODReports` — `backend:integration` 654/654 again.

PR #3358, opened 2026-09-14
feat(platform): carry conversations on non-mail channels

Israeltheminer:
The Inbox could only be filled by syncing a mailbox and could only answer over email. A product that owns its own customer-facing surface — an in-app support widget, a portal, a kiosk — had no way to use the shared Inbox as the place its staff answer from. This opens both halves, without changing anything about how mail conversations behave.

## Why

The ingest lane was already channel-agnostic: `addMessageToConversation` inherits the conversation's channel and takes `isCustomer`. The assumption lived entirely in the outbound lane and in the UI gate.

- `replyToConversation` required a contact email, built RFC threading headers, and dispatched to gmail/outlook/imap-smtp.
- `sendMessageViaConnectorInTx` stamped every outbound row `channel: 'email'` whatever the conversation said.
- The Inbox nav and routes rendered only when a deployed automation declared the `inbox` builtin view — the three mail sync packs.

## What changed

**`hostPolicy`** — one optional axis on a connector: `exact`, `suffix`, or `credential-origin`. Absent, it derives from `endpointMode`, so every existing connector file parses and behaves unchanged. `credential-origin` takes its single permitted host from the credential, for a connector that calls back to a deployment the operator runs and therefore has no vendor host to ship in `allowedHosts`. The https requirement and the private/link-local/cloud-metadata guard still apply, and no action input ever names a host.

**`webhook-channel` connector** with `send_message` and `send_status`. Native rather than a `yaml-js` body for one reason: each delivery is signed (Standard Webhooks — `webhook-id` / `webhook-timestamp` / `webhook-signature`), and the sandboxed context exposes no crypto. `webhook-id` is `ctx.idempotencyKey`, stable across retries, so a receiver that stores it can recognise a redelivery.

**Channel-aware send.** Threading headers, the `Re:` prefix and the contact address are the EMAIL path; another channel answers the contact's external id. Email behaviour is unchanged and pinned by its own regression tests.

**`conversations.credential_id`** (migration 0086). A conversation recorded its connector but not which of that connector's credentials it belonged to, and the send lane named none — which resolves to the org's default. On mail that is survivable, because the mailbox travels in the thread. Off mail the destination exists only on the credential, so an org running two products through one Inbox would deliver one product's customer reply to the other product's server. Nullable; null keeps today's behaviour.

**`/api/v1/conversations`** — open a thread, post a customer message, read, patch. Idempotent on `externalMessageId`, org-strict, and under the same assignment privacy the app applies.

**Status reaches the product.** `conversation.closed` was in the event vocabulary and nothing raised it; now it is raised, and a status change on a channel conversation is also delivered to the product holding the customer's view. That event alone only fans out to automation triggers, which a product learns nothing from.

**The Inbox opens for an org that has conversations**, not only one with a mail pack deployed — otherwise an org whose Inbox is in active use over the API cannot reach it.

## Risk

Two behaviours changed for existing rows, both covered by tests.

An agent reply now always moves a conversation to `pending` rather than preserving a non-open status. This is deliberate: a reply to a thread the customer had filed away must resurface it, or the answer is buried where they will never look.

The PATCH and bulk status doors read three more columns. The email send path has regression tests asserting subject, threading headers and connector dispatch are byte-identical.

`credential_id` is nullable and unset on every existing row, so mail conversations resolve their connector's default credential exactly as before.

## Tests

6586 server tests, 456 UI. New coverage: channel defaults, the credential-origin host policy (including that it admits the origin exactly and refuses its subdomains), the signed deliveries and their verification, the channel reply path, the status lane, the REST family, and the Inbox gate's second signal.

Three failures in this checkout are pre-existing and untouched by this branch: `boss-align` (`pg-boss` not installed), the date-picker suites (`tw-animate-css` not installed), and `config-watcher` (flaky under load; passes in isolation).

## Scope

Does not close the mail side of the same gap. Gmail and Outlook still reply through the org's default mailbox credential, because `buildSendInput` drops `from` for them and ingest does not stamp `credential_id`. Migration 0086 gives that its fix; wiring it is separate work and is deliberately not bundled here.

Six connectors still express "makes no HTTP requests" as an empty `allowedHosts`, which is indistinguishable from one that forgot its hosts. Worth an explicit policy value, separately.

Bulk status changes still emit no realtime hint, unlike every single-conversation write — pre-existing, and left alone.

Gate: `oxfmt --check`, `oxlint`, `typecheck`, both vitest projects, `lint:manual` (768 boxes), and the opengrep SAST hook all green.


Israeltheminer:
Closing: this duplicates work that landed on main while it was in flight (#3353 and the API-evaluation series). The conversations API there — snapshot sync plus a claimable delivery queue — is the better-integrated design, and a pull model spares the consumer a public endpoint and signature verification entirely. No reason to carry a second mechanism for the same capability.

One thing in here is still an open gap on main rather than a duplicate: `conversation.closed` is declared in `lib/shared/event-types.ts` and nothing raises it, so an org with an event trigger on it gets silence. `updateConversation` and `bulkSetConversationStatus` both emit only a realtime hint. Raising that separately as a small focused PR.

PR #3387, opened 2026-09-16
fix(platform): stop Claude Code re-sending a turn on a slow prefill

yannickmonney:
## What

The `claude-code` harness set no `CLAUDE_STREAM_IDLE_TIMEOUT_MS`, so the pinned CLI (2.1.173) gave up on a stream after its 300 s default without a chunk and **re-sent the same request** while the LLM gateway — which waits `SANDBOX_LLM_GATEWAY_STREAM_IDLE_TIMEOUT_SECONDS` (default 600 s) for a silent upstream — was still serving the first one. Any model whose prefill outlasts 300 s runs the turn twice.

## How

- `gatewayStreamIdleTimeoutSeconds()` in `llm_gateway_admin.ts` is now the single reader of the budget, used by both the provider config push and the managed harness credential (`gateway.streamIdleTimeoutMs`).
- The exec builder exposes it as the `${gateway.streamIdleTimeoutMs}` placeholder; `claude-code`'s `env.managed` sets `CLAUDE_STREAM_IDLE_TIMEOUT_MS` from it, so the two ends share one number.
- BYO-key runs keep the CLI default; the CLI keeps its own 300 s floor and derives its byte watchdog from the same value.

## Why it matters here

Found while running VAT plus against the on-prem Mac Studios: a local model's prefill regularly passes 300 s, and the duplicate turn was visible as two identical desk runs.

## Tests

`exec-builder.test.ts` covers the placeholder and the managed/BYO split; the gateway admin keeps its own suite. Docs: `.env.example` and the environment reference in all three locales.


yannickmonney:
Extended to Codex in f8e3ac1b4: the same watchdog, the same resend, but its knob is a provider config key rather than an env var — the managed block now pins `model_providers.tale.stream_idle_timeout_ms` to the gateway's budget as an unquoted TOML integer. Verified against the pinned CLI (0.142.5) that `stream_idle_timeout_ms` is a `ModelProviderInfo` field, so it sits in the same block as `base_url`/`wire_api` and cannot be a startup error. BYO keeps the CLI default. Fourteen managed goldens re-frozen deliberately, with the pin asserted in exec-builder.test.ts instead of relying on the frozen bytes alone.

yannickmonney:
Live data point from the VAT plus test environment (runtime b26ddbc4, without this PR), 2026-09-16: during an invoice-reader turn on a two-invoice PDF, the sandbox LLM gateway logged `POST /anthropic/v1/messages` → **502 after 321 321 ms** at 11:54:19Z, after a run of normal 16–58 s calls. That is consistent with the pinned Claude Code's 300 s stream-idle abort firing during a long vision prefill (≈21 s to headers + 300 s of silence) and the CLI resending the turn. The resend then streamed and the turn continued. A second case this morning, a setup-assistant turn, went silent after steady 7–14 s calls and had to be cancelled. With `CLAUDE_STREAM_IDLE_TIMEOUT_MS` tied to the gateway budget (1800 s on that deployment), neither turn would have been cut at 300 s.

PR #3388, opened 2026-09-16
feat(platform): delegate the notification export as a capability

yannickmonney:
## Why

A mirror worker that only needs to read notifications is refused `GET /api/v1/notifications/sync` with `403 ROLE_FORBIDDEN` on a Developer seat. The only way to unblock it today is to promote it to **admin** — which also hands it member management, SSO/SCIM administration and password resets. (Live case: VAT plus mirrors desk notifications into the client platform; its worker identity logs `ROLE_FORBIDDEN` on every poll.)

## What

An admin can delegate that one right through the existing competence register — no new table, no migration; grants and revocations use the current `/api/app/governance/competences` routes.

- **`competence.ts`**: a closed `PLATFORM_CAPABILITIES` set under the reserved `tale:` namespace. `grantCompetence` refuses any other `tale:` slug, in any case, with `400 COMPETENCE_CAPABILITY_UNKNOWN`, audited as `competence_grant_denied`. `holdsCapability` reads the member's one unrevoked, unexpired grant in the organization.
- **`removeMembershipCascade`** stamps the member's live `tale:` grants revoked, so a re-added member starts without the right.
- **`GET /api/v1/notifications/sync`** passes an owner or admin, or a member holding a live `tale:notifications.export` grant (never a disabled seat) — checked before the query is parsed or any member row is read. The 403 keeps `ROLE_FORBIDDEN` and names the capability.
- **`GET /api/v1/me`** answers `capabilities.notificationExport` from the same gate, so a client can tell before it polls.
- API contract **1.14.0**, with `openapi.json` and the fingerprint regenerated.

## Tests

Unit coverage for the capability set (unknown slug, casing, revoked/expired grant, disabled seat, cascade) and the two endpoints' gates; docs and the manual notifications suite carry the delegated-export box.


yannickmonney:
Closed a hole in the same promise, in 651549228: the register says a grant ends with the membership, and the members door and SCIM both keep that — but Better Auth's own `POST /api/auth/organization/remove-member` is mounted (it answers 400 on an empty body, unlike `/organization/delete`, which this config disables), and it deletes only the member row and its team rows. A member removed through that door kept a live `tale:` grant, so a re-add handed back a right no admin re-granted. The organization plugin's `afterRemoveMember` hook now runs the same cascade; it is idempotent, so the door that already ran it loses nothing. Test pins the wiring and fails without it.

yannickmonney:
And the door next to it, in 1596af5a2: `/organization/leave` deletes a membership through the plugin's adapter with no hook at all — so unlike remove-member it cannot be given the cascade — and with none of the guards the app's own removal carries (legal hold, audit row). A member who left that way kept their team rows, their SSO team-sync provenance and their live `tale:` grants. Nothing in the product calls it; leaving is an administered act here, the same reasoning that already disables `/organization/delete`, so the path joins `disabledPaths`. Happy to swap this for a guarded app-side leave endpoint if you'd rather keep the capability.

PR #3395, opened 2026-09-16
fix(cli): say why a deploy failed, and check packs before images

yannickmonney:
## Summary

A managed Tale deploy ran silently for 38 minutes, then printed only "Deployment failed. Review the source pins, paths, permissions and retained recovery receipts." I reproduced the cause with that CLI's own code:

- The CLI (3cde850b) predates `review.approve` (#3385).
- Its embedded manifest schema therefore strips `subjects.task.review.approve` from a pack that declares it.
- `assertNativeManifest` correctly refused the pack with `ConfigError("native manifest normalization changes release semantics")`.
- `managedResult` replaced that authored message with its fixed line.
- Prepare pulled and verified every runtime image before validating the configuration, and that pull was the 38 minutes.

This PR fixes all three parts:

1. **One failure rule for every configuration and deployment command** (`commands/operational-failure.ts`). Deliberate, bounded errors keep their words: `ConfigError` as a precondition, `NativeRequestError` and `ExternalToolError` as external dependencies. Anything else still gets the caller's fixed summary, so no unexpected message (credentials, registry or Docker output) reaches a log. `deploy` had its own stricter copy of this rule, and that copy is what hid the cause. `config releases` and `config platform` now use the same function; `config platform` keeps its operation summary for schema errors, which can come from the native target.
2. **The refusal names the fields.** `native manifest normalization changes release semantics at subjects.task.review.approve; this Tale CLI does not read those fields as written, so use a Tale CLI at least as new as the Tale the pack targets`. It lists paths only, never values, capped at five paths; a key that is not a plain identifier is quoted and bounded.
3. **Prepare checks configurations before the runtime** and logs each phase and image (`Preparing configuration <client>/<automation> at <sha>`, `Preparing the Tale runtime at <sha> for <platform>`, `Verifying runtime image <repository>`). A pack this CLI cannot read now fails within seconds. `--json` stays silent as before.

I did not add a commit-graph rule ("refuse a runtime newer than this CLI"). The prepared checkouts don't reliably contain the CLI's own commit, and the real incompatibility is the manifest vocabulary, which step 2 now names.

## Checklist

- [x] `bun run check` passes.
- [x] `bun run lint:sast` passes.
- [x] Translations: N/A. The CLI's messages are English-only; the docs are updated in all three locales.
- [x] Docs updated: `docs/{en,de,fr}/self-hosted/install/cli-install.md` (preparation order and failure reporting) and `tools/cli/README.md`.
- [x] README (root): N/A.

## Test plan

- `operational-failure.test.ts` covers each class mapping, schema drift, and pass-through of CLI and prompt errors. It also checks that an unexpected error with a credential in its message renders only the fixed summary.
- `deploy/managed.test.ts`: through `managedResult`, a configuration refusal from prepare reaches the operator in its own words, and the runtime fake is never called. An unexpected failure keeps the fixed line and hides its message.
- `compiler.test.ts` and `native.test.ts`: the refusal names `subjects.task.review.approve` / `subjects.task.review.unexpected` (the latter through the real `buildRelease` path). Values never appear; long or odd keys are quoted and bounded; more than five paths read "and N more".
- Red first: against the previous sources these tests fail (3 failures); with the change, all pass.
- Reproduction: 3cde850b's schema accepts pack `ec7a024` and refuses `5c4efbe`, which is the silent failure above.


yannickmonney:
Live check of the compiled CLI from this branch (`de60d587`, `bun-linux-arm64`), run in an Ubuntu container with local exact checkouts (`--sources-file`) and no Docker daemon:

- **A pack that declares a field this CLI does not know** (a scratch commit adding `subjects.task.review.futureReviewField`) exits 3 after 2 s, before any image work:
  ```
  Preparing configuration vatplus/vat-return-desk at 19ee2846…
  [ x ] native manifest normalization changes release semantics at subjects.task.review.futureReviewField; this Tale CLI does not read those fields as written, so use a Tale CLI at least as new as the Tale the pack targets
  ```
  The field's value does not appear.
- **The real pack** (the same client at `5c4efbe`) passes configuration preparation (the capsule is written) and moves on to the runtime, which fails as expected without a daemon (exit 5):
  ```
  Preparing configuration vatplus/vat-return-desk at 5c4efbeb…
  Preparing the Tale runtime at 5929b27f… for linux/arm64
  Verifying runtime image ghcr.io/tale-project/tale/tale-db
  [ x ] Docker could not complete the managed runtime operation.
  ```

PR #3396, opened 2026-09-16
fix(platform): stop Claude Code resending a request before its first byte

yannickmonney:
## Summary

#3387 raised Claude Code's stream idle watchdog to the gateway's budget. That covers the wait between chunks, but not the wait before the first byte, and a slow local prefill sends nothing until its first token. Two more bounds apply to that wait in Claude Code 2.1.173:

- **Bun's fetch timeout, about 300 s.** The CLI is a Bun binary and passes `timeout: false` to fetch only for `api.anthropic.com`, where its own byte watchdog takes over, or when `API_FORCE_IDLE_TIMEOUT` is explicitly false (`xM` / `uC_` in the CLI).
- **The SDK's request timeout, `API_TIMEOUT_MS`,** which defaults to 600 s.

Past either bound the CLI sends the request again, while the gateway keeps serving the abandoned copy (its fasthttp server does not cancel the upstream request when the client goes away). Against a model that serves one request at a time, the copies queue behind each other, and each resend waits behind them: a livelock, not a slow turn.

**Seen live on two deployments with a local model:**
- A turn's request was resent every ~300 s: four identical upstream requests at 16:20:48, 16:25:49, 16:30:51 and 16:35:56.
- The original answered 200 after 1,298 s, to a client that had long gone.
- A production desk showed the same pattern: a 200 after 1,251 s that nobody received.
- In both cases the turn could only advance when a call answered within 300 s.

**Fix.** The managed Claude Code env sets `API_FORCE_IDLE_TIMEOUT=0` and `API_TIMEOUT_MS=${gateway.streamIdleTimeoutMs}`, so every client-side bound follows the gateway's budget (`SANDBOX_LLM_GATEWAY_STREAM_IDLE_TIMEOUT_SECONDS`). The gateway stays the single bound on a silent upstream. BYO execs keep the CLI's defaults.

## Checklist

- [x] `bun run check` passes.
- [x] `bun run lint:sast` passes.
- [x] Translations: N/A (no UI strings).
- [x] Docs: the environment reference already says managed Claude Code waits the gateway's budget "including while a slow local model is still processing a long prompt"; this change makes that true. `services/sandbox-llm-gateway/README.md` now names all three client-side bounds.
- [x] README: N/A.

## Test plan

- `exec-builder.test.ts`: a managed Claude Code exec carries `API_FORCE_IDLE_TIMEOUT=0` and `API_TIMEOUT_MS` equal to the gateway budget, and follows a raised budget. A BYO exec sets neither, and no other harness changes with the budget.
- `golden-exec.test.ts`: the claude-code goldens are re-frozen with the two variables. Red without the harness change (2 failures), green with it.
- Controlled run of the real CLI (2.1.173, Linux) in a network-less container against a local fake Messages endpoint that stays silent before answering:
  - **Default env, 420 s silence:** request at t=3.1 s, a resend at **t=303.7 s** (`api_retry` attempt 1, error `unknown`). At t=423 s the first answer found its client gone. The CLI never finished.
  - **`API_FORCE_IDLE_TIMEOUT=0`, 420 s silence:** one request, answered at t=423 s, `result: success` with `ttft_ms` 420350. No resend.
  - The 660 s pair (past the SDK's 600 s default, with and without `API_TIMEOUT_MS=1800000`) will be added in a comment when it finishes.


yannickmonney:
The 660 s pair has finished. Same setup: the real CLI 2.1.173, a network-less container, a local Messages endpoint that stays silent before answering.

| Env | Requests seen by the endpoint | CLI outcome |
| --- | --- | --- |
| default, 420 s silence | t=3.1 s, **t=303.7 s** (resend) | never finished; the first answer at t=423 s found its client gone |
| `API_FORCE_IDLE_TIMEOUT=0`, 420 s silence | t=3.3 s only | `success`, `ttft_ms` 420350 |
| `API_FORCE_IDLE_TIMEOUT=0`, 660 s silence | t=4.0 s, **t=604.7 s** (resend) | never finished; the first answer at t=664 s found its client gone |
| `API_FORCE_IDLE_TIMEOUT=0` + `API_TIMEOUT_MS=1800000`, 660 s silence | t=4.1 s only | `success`, `ttft_ms` 660406 |

So both variables are needed. The first lifts Bun's 300 s fetch timeout, and the second moves the SDK's 600 s request timeout to the gateway's budget. The resends were logged by the CLI as `api_retry` with `error: "unknown"`.

